PR #13, opened 2018-10-07
feat: add query API

### LINKED ISSUE #6: Implement query* API
See https://github.com/kentcdodds/react-testing-library#query-apis.
Basically: instead of throwing, return `null`.

thymikee:
### Summary

`query` APIs corresponding to `get` APIs. For simplicity and code reuse I've decided to go with calling `get` APIs directly – as a downside we throw/catch errors twice. If this ever has a bigger perf impact we can switch the implementation.

Fixes #6 

### Test plan

Added unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,6 +282,30 @@ test('fetch data', async () => {
 });
 ```
 
+## `query` APIs
+
+Each of the get APIs listed in the render section above have a complimentary query API. The get APIs will throw errors if a proper node cannot be found. This is normally the desired effect. However, if you want to make an assertion that an element is not present in the DOM, then you can use the query API instead:
+
+```jsx
+import { render } from 'react-native-testing-library';
+
+const { queryByText } = render(<Form />);
+const submitButton = queryByText('submit');
+expect(submitButton).toBeNull(); // it doesn't exist
+```
+
+## `queryAll` API
+
+Each of the query APIs have a corresponding queryAll version that always returns an Array of matching nodes. getAll is the same but throws when the array has a length of 0.
+
+```jsx
+import { render } from 'react-native-testing-library';
+
+const { queryAllByText } = render(<Forms />);
+const submitButtons = queryAllByText('submit');
+expect(submitButtons).toHaveLength(3); // expect 3 elements
+```
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -282,6 +282,30 @@ test('fetch data', async () => {
 });
 ```
 
+## `query` APIs
+
+Each of the get APIs listed in the render section above have a complimentary query API. The get APIs will throw errors if a proper node cannot be found. This is normally the desired effect. However, if you want to make an assertion that an element is not present in the DOM, then you can use the query API instead:
```
Esemesek: I think we are not using DOM here C:

src/render.js
```diff
@@ -30,6 +39,15 @@ export default function render(
     getAllByName: getAllByName(instance),
     getAllByText: getAllByText(instance),
     getAllByProps: getAllByProps(instance),
+
+    queryByTestId: queryByTestId(instance),
```
Esemesek: I think we should replace exports from `getBy API` and `queryBy API` by some factory function to avoid listing all the re-exports here. Eg.
```jsx
// getBy.js
...
export const getByAPI (instance) => ({
  queryByTestId: queryByTestId(instance),
  ...
});

// render.js
return {
  ...getByAPI(instance),
  ...queryByAPI(instance),
  ...
}
```

src/render.js
```diff
@@ -30,6 +39,15 @@ export default function render(
...
+    queryByName: queryByName(instance),
+    queryByText: queryByText(instance),
+    queryByProps: queryByProps(instance),
+    queryAllByName: queryAllByName(instance),
+    queryAllByText: queryAllByText(instance),
+    queryAllByProps: queryAllByProps(instance),
+
```
Esemesek: Can we additionally expose `root`?
```jsx
root: () => renderer.root;
```

thymikee: Yea, I think that makes sense in this form.

thymikee: After thinking it through, I currently see no real value of exposing this, as we have `get/query` fns to search component tree, and we know the props passed to the root.

PR #18, opened 2018-10-07
Feat/typescript types

### LINKED ISSUE #15: Add TypeScript typings
We use Flow internally, because that's what React Native comes with by default. Would be cool to add TS typings though :)

Esemesek:
Add Typescript types.
Fixes #15.

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -1,0 +1,59 @@
+import * as React from 'react';
+import { ReactTestInstance } from 'react-test-renderer';
+
+export interface GetByAPI {
+  getByName: (name: React.ReactType) => ReactTestInstance;
+  getByText: (text: string | RegExp) => ReactTestInstance;
+  getByProps: (props: Record<string, any>) => ReactTestInstance;
+  getByTestId: (testID: string) => ReactTestInstance;
+  getAllByName: (name: React.ReactType) => Array<ReactTestInstance>;
+  getAllByText: (text: string | RegExp) => Array<ReactTestInstance>;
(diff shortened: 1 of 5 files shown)


typings/index.d.ts
```diff
@@ -0,0 +1,59 @@
+import * as React from 'react';
+import { ReactTestInstance } from 'react-test-renderer';
+
+export interface GetByAPI {
+  getByName: (name: React.ReactType) => ReactTestInstance;
```
thymikee: `name: React.ReactType | string`, same for `getAllByName`

typings/index.d.ts
```diff
@@ -0,0 +1,59 @@
...
+  getByText: (text: string | RegExp) => ReactTestInstance;
+  getByProps: (props: Record<string, any>) => ReactTestInstance;
+  getByTestId: (testID: string) => ReactTestInstance;
+  getAllByName: (name: React.ReactType) => Array<ReactTestInstance>;
+  getAllByText: (text: string | RegExp) => Array<ReactTestInstance>;
+  getAllByProps: (props: Record<string, any>) => Array<ReactTestInstance>;
+}
+
+export interface QueryByAPI {
+  queryByName: (name: React.ReactType) => ReactTestInstance | null;
```
thymikee: same as above

typings/index.d.ts
Esemesek: `React.ReactType` already includes string C: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L52

thymikee: Oh, that's cool! Didn't know that :D

PR #20, opened 2018-10-10
feat: add waitForElement

thymikee:
### Summary

Adds `waitForElement` API.
The idea is to await throwable callback until it stops throwing or exceeds timeout. This way we can test async invoked actions with ease.

### Test plan

Added tests

### DIFF
--- a/src/__tests__/waitForElement.test.js
+++ b/src/__tests__/waitForElement.test.js
@@ -1,0 +1,92 @@
+// @flow
+/* eslint-disable react/no-multi-comp */
+import React from 'react';
+import { View, Text, TouchableOpacity } from '../__mocks__/reactNativeMock';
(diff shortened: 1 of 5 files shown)


src/__tests__/waitForElement.test.js
```diff
@@ -0,0 +1,92 @@
+// @flow
+/* eslint-disable react/no-multi-comp */
```
Esemesek: Missing test case, where you check for rejected Promise when element is not found in timeout.

thymikee: um, is this what you mean? https://github.com/callstack/react-native-testing-library/pull/20/files#diff-1c9876d94966fa35d3d6b0ccf7b9a8fbR51

Esemesek: Yup C:

PR #31, opened 2018-10-15
Fix Remove unnecessary flow expression

Esemesek:
Remove unnecessary flow expression that caused error in projects created by `expo-cli`.


thymikee:
Yea that was unnecessary, thanks!

Esemesek:
Fixes #31

PR #33, opened 2018-10-15
feat: add instance and renderer back to RenderResults

### LINKED ISSUE #32: render does not return the renderer
The [`render` section of the README](https://github.com/callstack/react-native-testing-library/blob/6097c8f64ce6ad0b3626e2616615d74d73318b22/README.md#render) says:

> For convenience it also returns react-test-renderer's instance and renderer objects, in case you need more control.

But [it doesn't](https://github.com/callstack/react-native-testing-library/blob/6097c8f64ce6ad0b3626e2616615d74d73318b22/src/render.js#L18). But it should! 🙂 Looks like they were removed in e6ac5ad9977b81619ddf1e7cbdeb98cf553c82a8, but I think they would be valuable (at least `renderer`). Was there a reason they were removed?

PR incoming to add them back.

wKovacs64:
Closes #32.

### Summary

The `instance` and `renderer` fields are [documented in the README](https://github.com/callstack/react-native-testing-library/blob/6097c8f64ce6ad0b3626e2616615d74d73318b22/README.md#render) but were removed from `RenderResult` in e6ac5ad9977b81619ddf1e7cbdeb98cf553c82a8. I think they add value and (at least `renderer`) should still be included. This PR re-introduces them to the object returned from `render`.

##### Alternatives:

- only include `renderer` (as `instance` is just `renderer.root` anyway)
- leave them out of `RenderResult` and remove them from the README

### Test plan

N/A

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -16,6 +16,8 @@ export default function render(
   const instance = renderer.root;
 
   return {
+    instance,
+    renderer,
     ...getByAPI(instance),
     ...queryByAPI(instance),
     update: renderer.update,


thymikee:
Od rather remove it from the Readme, as that was the intention, just forgot to update it. What's your use case for exposing the  renderer?

wKovacs64:
Debugging, primarily. react-testing-library's `debug` function is what I wanted but react-native-testing-library's `debug` is shallow so it was not helpful in my case. `renderer.toJSON` was my best bet at the time.

thymikee:
We could add this as an API, how about something like:
```js
degug(<Component />, {deep: true})
```
?

or returning `toJSON` from `render`, if you ever want to snapshot whole component, which actually may make sense for some smaller ones.

thymikee:
Yea, we explicitly don't want users to access `.getInstance()` method. Would you be so kind and update this PR with what we've come up here? 🙂

wKovacs64:
Thanks, sorry I wasn't able to get to the revamp - you beat me to it! 🙂

thymikee:
No worries, thanks for you contribution, much appreciated!

wKovacs64:
Are you still thinking of adding deep debug? I guess that's essentially just a prettified version of toJSON?

thymikee:
I think we can make it a thing, the API however would rather be `debug(<Component />, '', {deep: true})` as we allow second parameter to be a string message.

On the other hand we could check the type of second parameter so you could either use:
```js
debug(<Component />, 'message'); // shorthand, current API
debug(<Component />, {deep: true, message: 'message'}); // debug with more options
```
wdyt? cc @Esemesek

Esemesek:
I don't like those configurations. Methods like `debug` should be short for convenience. We often use those to get a quick feedback, adding configuration as a second parameter will complicate things and will encourage us to add more configurable fields in the future. We could make it happen, but with slight API change:
```jsx
debug(<Component />, 'message'); // defaults to shallow
debug.shallow(<Component />, 'message'); // same as above
debug.deep(<Component />, 'message'); // deep debug
```
Additionally, we won't break existing API.
wdyt?

thymikee:
um, like it!

Esemesek:
Just like `console.log`, `console.warn` etc C:

thymikee:
@wKovacs64 would you like to build it?

wKovacs64:
I don't know jack about Flow, so it'd probably be somewhat of a train wreck. 😉

PR #35, opened 2018-10-16
feat: add toJSON to render; cleanup shallow

thymikee:
### Summary

Exposing `toJSON` so it's possible to snapshot/debug the fully rendered component.
By the way removed the dead code in `shallow` that was not supported (we've never actually wanted to support `shallow(Component)` as it doesn't make sense without passing props.

Fixes #33.

### Test plan

Added a test, which renders funny because of our mocks.

### DIFF
--- a/src/debug.js
+++ b/src/debug.js
@@ -1,7 +1,7 @@
 // @flow
 import * as React from 'react';
 import prettyFormat, { plugins } from 'pretty-format'; // eslint-disable-line import/no-extraneous-dependencies
-import { shallow } from '.';
+import shallow from './shallow';
 
 /**
  * Log pretty-printed shallow test component instance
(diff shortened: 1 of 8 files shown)


src/debug.js
```diff
@@ -1,7 +1,7 @@
 // @flow
 import * as React from 'react';
 import prettyFormat, { plugins } from 'pretty-format'; // eslint-disable-line import/no-extraneous-dependencies
-import shallow from './shallow';
+import { shallow } from '.';
```
Esemesek: Is this a cyclic dependency?

thymikee: hm, yea, reverting

Esemesek:
Can you also update documentation?

PR #36, opened 2018-10-16
fix: use actual react-native instead of homegrown mocks

thymikee:
### Summary

This actually uncovered some bugs, like lack of `onDoublePress` (lol XD) and enabled us to have better snapshots for deep rendering.

### Test plan

no new tests


src/__tests__/__snapshots__/shallow.test.js.snap
```diff
@@ -1,31 +1,23 @@
...
 
 exports[`shallow rendering React elements 1`] = `
-<View
-  testID="root"
->
-  <TouchableOpacity
-    onPress={[Function]}
-    testID="button"
+<Component>
+  <Component
```
thymikee: RN mocks are good at naming things XD

PR #40, opened 2018-10-17
feat: make debug.deep more usable

thymikee:
### Summary

Make `debug.deep` more usable by allowing it to print arbitrary JSON (and actually any) files. This proves handy when we want to debug async component which we want to snapshot after async operations:

```jsx
const { toJSON, getByName } = render(
  <Button onPress={jest.fn} text="Press me" />
);
fireEvent.press(getByName('TouchableOpacity'));
await flushMicrotasksQueue();

debug.deep(toJSON());
```

### Test plan

added one

### DIFF
--- a/src/debug.js
+++ b/src/debug.js
@@ -20,10 +20,19 @@ function debugShallow(
 /**
  * Log pretty-printed deep test component instance
  */
-function debugDeep(instance: React.Element<*>, message?: any) {
-  const { toJSON } = render(instance);
-
-  console.log(format(toJSON()), message || '');
+function debugDeep(
+  instance: React.Element<*> | ?ReactTestRendererJSON,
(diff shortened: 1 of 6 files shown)


src/debug.js
```diff
@@ -20,10 +20,19 @@ function debugShallow(
 /**
  * Log pretty-printed deep test component instance
  */
-function debugDeep(instance: React.Element<*>, message?: any) {
-  const { toJSON } = render(instance);
-
-  console.log(format(toJSON()), message || '');
+function debugDeep(
+  instance: React.Element<*> | ?ReactTestRendererJSON,
+  message?: any
```
Esemesek: Use default parameter here, instead of `message || ''` later on.

PR #41, opened 2018-10-20
Add getInstance helper

benadamstyles:
This allows testing `instance.state` and other advanced patterns using the component instance.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -21,5 +21,6 @@ export default function render(
     update: renderer.update,
     unmount: renderer.unmount,
     toJSON: renderer.toJSON,
+    getInstance: () => renderer.getInstance()
   };
 }


benadamstyles:
(Obviously I'm happy to writes tests and types, I just wanted to get feedback on the idea first!)

thymikee:
Sorry to be a bearer of "bad" news, but this is a design decision behind this library not to export getInstance. Testing internal component instance is basically testing implementation details. If you have parts of the state management that you'd like to test e.g. for edge cases, export that piece of functionality and test in isolation. 

I'm thankful for the PR though, that's great attitude!

benadamstyles:
Fair enough, no problem! Maybe you should change the docs [here](https://github.com/callstack/react-native-testing-library#render) which state:

> For convenience it also returns react-test-renderer's instance and renderer objects, in case you need more control.

That actually isn't the case, but it made me think that you might be open to adding that functionality back in.

thymikee:
Ahh forgot to remove that! Would you like to send a PR?

PR #43, opened 2018-10-21
Documentation update for The Problem section, Enzyme, benefits

### LINKED ISSUE #24: Update README overview comparison to Enzyme
As discussed in this thread:

https://twitter.com/thymikee/status/1050432901945483265

It would be good to both spell out that Enzyme can be set up with deep rendering with `mount()`, which I plan to update documentation for here:

https://github.com/airbnb/enzyme/issues/1375

But it would also be good to highlight the benefits of using react-native-testing-library at the same time. As @thymikee mentioned: 

```Yup, but I don't want to use jsdom as RN is not web. Also Enzyme makes it too easy to write bad tests that assert on implementation details. [...]```

I might be able to help with documentation here soon if no one else wants to pick this up, though I might try out the library myself first when I get a chance so I can better speak to it.

shawnaxsom:
### Summary

Updated to better compare with Enzyme capabilities while highlighting more react-native-testing-library benefits.

It is good for people to know the options (that Enzyme *can* technically be used currently for deep rendering), but also further sell the potential user on the benefits of react-native-testing-library.

This fixes #24.

I left out links to the Enzyme discussions, as it is up to the maintainers to include those and the goal is to highlight benefits while being fair.

Thanks for your consideration, and providing a great alternative to Enzyme!

### Test plan

See https://github.com/airbnb/enzyme/pull/1873, where I have submitted an Enzyme PR to update testing instructions for Enzyme. There I mention how Enzyme does plan an adapter to support deep rendering, but that is still open, while I also show how jsdom can be implemented in Enzyme to use deep rendering (as we accomplished in our React Native boilerplate solution https://github.com/ProminentEdge/mobile-boilerplate).

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,11 @@ _Appreciation notice: This project is heavily inspired by [react-testing-library
 
 ## The problem
 
-You want to write maintainable tests for your React Native components without testing implementation details, but then you're told to use Enzyme, which only supports shallow rendering in React Native environment. And you want to render deep! But doing so with `react-test-renderer` is so painful.
+You want to write maintainable tests for your React Native components without testing implementation details, but then you're told to use Enzyme, which you learn has no React Native adapter, meaning only shallow rendering is supported. And you want to render deep! But deep rendering may otherwise require jsdom (React Native isn't the web!), while doing deep rendering with `react-test-renderer` is so painful.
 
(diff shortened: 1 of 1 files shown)


thymikee:
Thanks @shawnaxsom, appreciated!

Issue #49, opened 2018-10-29
Tests won't compile because library is published as ES6


henrikra:
I am trying to run simple test case in TypeScript project but I am getting 

```bash
> mocha
/myApp/node_modules/react-native-testing-library/src/index.js:2
import render from './render';
^^^^^^
SyntaxError: Unexpected token import
```

This means that you guys are publishing the project code in ES6 even though node_modules should be released as ES5.

Could you compile the library code to ES5 and republish that :)

sibelius:
You should ask Babel-jest to transpile this library instead

henrikra:
I am using mocha. I think the library code should be available as possible like other libraries are doing. If you really really really want to have library code in ES6 then there should be some steps on the README on how to run it. But that doesn't make any sense since that would require different guides to all different tests runners

thymikee:
We mainly target React Native, where projects are released as source. Before you feed this lib to your TS compiler, you'll need to run it through Babel with `metro-react-native-babel-preset`.

I think we could compile it to ES5 though, I'm happy to review and accept the PR with that change :)

thymikee:
Feel free to fix it, this is an open source project and I didn't say no to this kind of improvement 😃.

henrikra:
@thymikee Okay do you have something in mind that you would use to compile this to ES5?

thymikee:
Yup, I'd go with `@babel/preset-env` with a target of Node 8 (that's the minimum required by React Native right now).

thymikee:
Or actually we could just go with `metro-react-native-babel-preset`, already available in dev deps. I think that will be simpler to set up

henrikra:
So should the compilation be in `release-it`? Looks like you guys are using that but I don't have experience with it

thymikee:
Don't worry about `release-it`. There should just be a git-ignored (but not npm-ignored) `build` folder with compiled sources, that would get created/updated on pre-release npm hook.

Jukkish:
@thymikee I made a PR based on your tips, can you check it out?

thymikee:
Just published `@1.4.1-es5.0` with compiled sources. Let me know if it works for you, both TS and Flow users

thymikee:
@henrikra @Jukkish ping ^

PR #50, opened 2018-10-29
Compile to ES5

Jukkish:
### Summary

Fixes https://github.com/callstack/react-native-testing-library/issues/49


thymikee:
Thanks! I've adjusted the PR a bit. We don't use `npm`, so no `package-lock.json`.

Jukkish:
Thanks!

thymikee:
Um, it's not merged yet?

Jukkish:
> Um, it's not merged yet?

Sry, it was an accident :D

Esemesek:
Looks good to me. We still need to investigate if `flow` types are working correctly with this configuration.

thymikee:
They're not. We're gonna need something like https://github.com/Macil/flow-copy-source

henrikra:
Side note: does it make sense to have tests inside src folder? Usually tests are on root level 🤔

thymikee:
Of course it does! The same way as having them in the root level or side-by-side to your source files 😉

henrikra:
Can we get this merged? 😀

thymikee:
Not yet, we need to figure out exporting Flow types, would like to avoid duplication where possible.

PR #62, opened 2018-11-15
fix(breaking): correct faulty implementation of finding by name

thymikee:
### Summary

Migrating from RN 0.57.4 -> 0.57.5 revealed a flaw in the implementation of our helper to get nodes by name. I believe now we should have never test against `node.type` directly if it's a string, as this compares to the host component type and may result in finding more nodes than actually expected (which in RN versions prior to 0.57.5 allowed us to do `getByName('Text')` or `getByName('View')` where they didn't have `type.displayName` nor `type.name` set correctly).

This is breaking but since it fixes the behavior to correct one I'm keen on releasing this as a patch.

Migration path:

- if you can, upgrade to RN 0.57.5 to avoid some manual work

rename:
- `getByName('Text')` -> `getByName(Text)`
- `getByName('View')` -> `getByName(View)`
- `getByName('Image')` -> `getByName(Image)`
- `getByName('Modal')` -> `getByName(Modal)`
- and so on...

While working on this I realized that getting by string name is very fragile in React ecosystem. Tests written this way may break as soon as you (or authors of libraries you use) wrap components with a HOC or `forwardRef`. This is a problem I've seen in Enzyme as well (using `shallow.find` API) but I belittled it.

Since now on we'll be discouraging usage of `getByName(string)` as dangerous/fragile API and introduce a new helper called `getByType` which will do the same as `getByName(ReactComponent)`.

BTW fixed incorrect error message when more than 1 element was found by `getBy*` methods.

### Test plan

Tests green on RN 0.57.5

### DIFF
--- a/src/__tests__/__snapshots__/shallow.test.js.snap
+++ b/src/__tests__/__snapshots__/shallow.test.js.snap
@@ -13,11 +13,11 @@ exports[`shallow rendering React Test Instance 1`] = `
 `;
 
 exports[`shallow rendering React elements 1`] = `
-<Component>
-  <Component
+<View>
+  <Text
     testID="text-button"
   >
     Press me
-  </Component>
-</Component>
+  </Text>
+</View>
 `;
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -57,7 +57,7 @@ test('getByTestId, queryByTestId', () => {
   const component = getByTestId('bananaFresh');
 
   expect(component.props.children).toBe('not fresh');
-  expect(() => getByTestId('InExistent')).toThrow();
+  expect(() => getByTestId('InExistent')).toThrow('No instances found');
 
   expect(getByTestId('bananaFresh')).toBe(component);
   expect(queryByTestId('InExistent')).toBeNull();
@@ -76,7 +76,8 @@ test('getByName, queryByName', () => {
   sameButton.props.onPress();
 
   expect(bananaFresh.props.children).toBe('not fresh');
-  expect(() => getByName('InExistent')).toThrow();
+  expect(() => getByName('InExistent')).toThrow('No instances found');
+  expect(() => getByName(Text)).toThrow('Expected 1 but found 3');
 
   expect(queryByName('Button')).toBe(button);
   expect(queryByName('InExistent')).toBeNull();
@@ -89,7 +90,7 @@ test('getAllByName, queryAllByName', () => {
   expect(text.props.children).toBe('Is the banana fresh?');
   expect(status.props.children).toBe('not fresh');
   expect(button.props.children).toBe('Change freshness!');
-  expect(() => getAllByName('InExistent')).toThrow();
+  expect(() => getAllByName('InExistent')).toThrow('No instances found');
 
   expect(queryAllByName('Text')[1]).toBe(status);
   expect(queryAllByName('InExistent')).toHaveLength(0);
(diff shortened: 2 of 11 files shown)


src/__tests__/render.test.js
```diff
@@ -76,7 +76,8 @@ test('getByName, queryByName', () => {
   sameButton.props.onPress();
 
   expect(bananaFresh.props.children).toBe('not fresh');
-  expect(() => getByName('InExistent')).toThrow();
+  expect(() => getByName('InExistent')).toThrow('No instances found');
+  expect(() => getByName(Text)).toThrow('Expected 1 but found 3');
```
thymikee: fixed errors btw, so it doesn't incorrectly say "No instances found" when there's more than 1

src/helpers/getByAPI.js
```diff
@@ -1,24 +1,44 @@
...
+    const { Text, TextInput } = require('react-native');
+    return (
+      (getNodeByType(node, Text) || getNodeByType(node, TextInput)) &&
+      (typeof text === 'string'
+        ? text === node.props.children
+        : text.test(node.props.children))
+    );
+  } catch (error) {
+    throw new Error(
+      `Currently the only supported library to search by text is \`react-native\`.\n\n${error}`
```
Esemesek: Can you extract constructing this string to a method? Or even create custom Error?

src/helpers/getByAPI.js
```diff
@@ -1,24 +1,44 @@
 // @flow
 import * as React from 'react';
+import prettyFormat from 'pretty-format';
 import ErrorWithStack from './errorWithStack';
 
+const getNodeByType = (node, type) => node.type === type;
```
Esemesek: I think it will be better to name `getNodeByName` and `getNodeByType` as `filterNodeByName` and `filterNodeByType`. The `getNode` method indicates that it returns `Node`, but in fact it returns `Boolean`

src/helpers/getByAPI.js
```diff
@@ -1,24 +1,44 @@
...
-  node.type.name === name ||
-  node.type.displayName === name ||
-  node.type === name;
+  typeof node.type !== 'string' &&
+  (node.type.displayName === name || node.type.name === name);
+
+const getNodeByText = (node, text) => {
+  try {
+    // eslint-disable-next-line
+    const { Text, TextInput } = require('react-native');
```
Esemesek: As I understand, we make this `require` call here to make lib compatible with other environments?

src/helpers/getByAPI.js
```diff
@@ -1,24 +1,44 @@
...
+    return (
+      (getNodeByType(node, Text) || getNodeByType(node, TextInput)) &&
+      (typeof text === 'string'
```
Esemesek: `typeof something === 'string'` is used couple of times throughout the code - can you extract it to method?

src/__tests__/__snapshots__/shallow.test.js.snap
```diff
@@ -13,11 +13,11 @@ exports[`shallow rendering React Test Instance 1`] = `
 `;
 
 exports[`shallow rendering React elements 1`] = `
-<Component>
-  <Component
+<View>
```
Esemesek: This looks much better C:

src/helpers/getByAPI.js
thymikee: There may be different text primitives in different environments (like ReactVR). I stand that it's safer to assert on right primitives instead of using string-based comparisons, where the display name may change over time.

src/helpers/getByAPI.js
thymikee: Wish I could, but Flow looses type information because of fn extraction :<

thymikee:
Addressed, thanks!

PR #64, opened 2018-11-18
feat: add getByType helper

thymikee:
### Summary

Adds new `getByType`, `getAllByType` render helper and `query*` equivalents to find components by type only. This is going to be the method where we'll point `getByName` when deprecation.

### Test plan

Included.


thymikee:
I'll add the docs when we'll mark `getByName` as deprecated in a followup PR.

PR #93, opened 2018-12-15
Fix link for api docs in read me file

alinaMihai:


thymikee:
Oh, I read it wrong. The file should be renamed to API.md (just did). Sorry for the confusion!

alinaMihai:
OK, thank you for the quick feedback !

PR #100, opened 2018-12-22
Add fireEvent.blur support

danahartweg:
### Summary
There is not yet support for calling text input blur events via `fireEvent`. This PR adds that support.

### Test plan
+ Query the test instance for a react native `TextInput`
+ Call `fireEvent.blur()` on the `TextInput` test instance
+ Verify the supplied handler is called properly

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -200,6 +200,25 @@ const { getByTestId } = render(
 fireEvent.changeText(getByTestId('text-input'), CHANGE_TEXT);
 ```
 
+### `fireEvent.blur: (element: ReactTestInstance) => void`
+
+Invokes `blur` event handler on the element or parent element in the tree.
+
+```jsx
+import { View, TextInput } from 'react-native';
+import { render, fireEvent } from 'react-native-testing-library';
+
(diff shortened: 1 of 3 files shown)


thymikee:
Thanks for your PR, it's greatly appreciated. However, with this mindset we would have to implement every event out there, which we'd like to avoid because of pain to maintain it. Especially that this library aims to support any React application, not only React Native.

We try to keep the `fireEvent` API as small as possible, because you can always fire:
```js
fireEvent(getByTestId('id'), 'blur')
```
In a way of exception the `press` and `changeText` are exposed, because they're used pretty often, at least in our experience. Maybe we can document the generic `fireEvent` somehow better, perhaps add more examples of using it with various events?

danahartweg:
@thymikee that makes a lot of sense! Honestly, I should have come to that conclusion when diving into the code, I just didn't even think about it since there were a couple of them aliased.

I'll close this out and add some additional documentation instead =)

thymikee:
>I'll close this out and add some additional documentation instead =)

That would be lovely, thanks!

PR #101, opened 2018-12-22
Add getByPlaceholder query functionality

danahartweg:
### Summary
There is a helper that exists in `react-testing-library` to query inputs by their `placeholder`. This PR adds that support.

You **could** achieve this by querying for a `placeholder` prop, or querying for react native `TextInput`s manually and then checking their props, but it felt cleaner to have this contained in the testing library itself. It will also allow you to bypass any other instances that may have a `placeholder` prop that aren't `TextInput`s.

If we don't want to expand the surface of the library further, I completely understand that and am fine with this being closed. Just seems like it may be useful and save some time for several scenarios =)

### Test plan
+ Render a `TextInput` with a `placeholder` prop
+ Query the test instance via `getByPlaceholder` for the input
+ Verify the test instance is found

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -56,7 +56,7 @@ A method returning an array of `ReactTestInstance`s with matching text – may b
 
 A method returning a `ReactTestInstance` for a `TextInput` with a matching placeholder – may be a string or regular expression. Throws when no matches.
 
-### `getAllByPlaceholder: (text: string | RegExp)`
+### `getAllByPlaceholder: (placeholder: string | RegExp)`
 
 A method returning an array of `ReactTestInstance`s for `TextInput`'s with a matching placeholder – may be a string or regular expression.
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -52,6 +52,14 @@ A method returning a `ReactTestInstance` with matching text – may be a string 
 
 A method returning an array of `ReactTestInstance`s with matching text – may be a string or regular expression.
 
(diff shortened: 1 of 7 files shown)


docs/API.md
```diff
@@ -52,6 +52,14 @@ A method returning a `ReactTestInstance` with matching text – may be a string
 
 A method returning an array of `ReactTestInstance`s with matching text – may be a string or regular expression.
 
+### `getByPlaceholder: (placeholder: string | RegExp)`
+
+A method returning a `ReactTestInstance` for a `TextInput` with a matching placeholder – may be a string or regular expression. Throws when no matches.
+
+### `getAllByPlaceholder: (text: string | RegExp)`
```
thymikee: Can we change this `text` to `placeholder`? Or the other way around if you feel like `text` is nicer :)
```suggestion
### `getAllByPlaceholder: (placeholder: string | RegExp)`
```

danahartweg: Well crap, I thought I grabbed all of them!

thymikee:
Mind adding a test here as well? https://github.com/callstack/react-native-testing-library/blob/master/typings/__tests__/index.test.tsx

thymikee:
Thank you so much for this contribution!

danahartweg:
You're very welcome! As I use the library more, I'll be sure to keep a list of things that could be improved / added, and then get to discussing or implementing them.

PR #121, opened 2019-02-22
Esemesek/templates

Esemesek:
We were missing those templates.


Esemesek:
Messed up my branch, but it should be fine C:

thymikee:
Btw now we can enforce some of those on an organization level! :)

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,22 @@
+---
+name: 🐛 Report a bug
+about: Report a reproducible or regression bug in React Native Testing Library'
+labels: 'bug'
```
thymikee: Change to "bug report", they're usually not bugs

PR #132, opened 2019-03-13
Allow for text to match children who are composed as array

eliperkins:
### Summary

Given a component that renders text by composing together literal text with an inline expression for a dynamic variable, React Native will render this as a `<Text>` element with multiple children. For example:

```js
const BananaCounter = ({ numBananas }) => (
  <Text>There are {numBananas} bananas in the bunch</Text>
);

const { toJSON, debug } = render(<BananaCounter numBananas={3} />);

debug();
/*
      <Text>
        There are
        3
         bananas in the bunch
      </Text>
*/

expect(toJSON()).toMatchInlineSnapshot(`
<Text>
  There are 
  3
   bananas in the bunch
</Text>
`);
```

This makes sense, as the component is a:
- literal string
- a dynamic evaluation
- literal string

Unfortunately, this means that writing a test that finds an element based on that dynamic evaluation fails when using `getByText`.

```js

const { getByText } = render(<BananaCounter numBananas={3} />);
expect(getByText('There are 3 bananas in the bunch')).toBeTruthy(); // Fails
```

This is because we compare the given test string directly against `children`. https://github.com/callstack/react-native-testing-library/blob/ce3bf28f308728672bc5502e120048821c60218b/src/helpers/getByAPI.js#L23-L24

This results in comparing `'There are 3 bananas in the bunch' === ['There are ' 3, ' bananas in the bunch']`, which of course will fail.

The solution is to join children and compare the joined result against the given text string.

### Test plan

A test for this new functionality is provided!

### DIFF
--- a/src/helpers/getByAPI.js
+++ b/src/helpers/getByAPI.js
@@ -17,12 +17,20 @@ const getNodeByText = (node, text) => {
   try {
     // eslint-disable-next-line
     const { Text, TextInput } = require('react-native');
-    return (
-      (filterNodeByType(node, Text) || filterNodeByType(node, TextInput)) &&
-      (typeof text === 'string'
-        ? text === node.props.children
-        : text.test(node.props.children))
-    );
+    const isTextNode =
+      filterNodeByType(node, Text) || filterNodeByType(node, TextInput);
+    if (isTextNode) {
+      const textChildren = React.Children.map(node.props.children, child =>
+        child.toString()
+      );
+      if (textChildren) {
+        const textToTest = textChildren.join('');
+        return typeof text === 'string'
+          ? text === textToTest
+          : text.test(textToTest);
+      }
+    }
+    return false;
   } catch (error) {
     throw createLibraryNotSupportedError(error);
   }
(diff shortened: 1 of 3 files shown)


src/helpers/getByAPI.js
```diff
@@ -17,12 +17,23 @@ const getNodeByText = (node, text) => {
...
     const { Text, TextInput } = require('react-native');
-    return (
-      (filterNodeByType(node, Text) || filterNodeByType(node, TextInput)) &&
-      (typeof text === 'string'
-        ? text === node.props.children
-        : text.test(node.props.children))
-    );
+    const isTextNode =
+      filterNodeByType(node, Text) || filterNodeByType(node, TextInput);
+    if (isTextNode) {
```
eliperkins: I don't love how this came out, so I'd love some feedback here!

thymikee: I think that's fine. Can we change the naming to `isTextComponent`?

eliperkins: Done in 7a1a3f4

Esemesek:
@eliperkins I think we should also mention this behavior in the documentation. Can you also make changes there?

eliperkins:
Thanks y'all! 🎊

Esemesek:
Thank you too ❤️ ❤️ ❤️

thymikee:
Available in 1.6.1, thank you :)

PR #137, opened 2019-03-23
Add act to update call to support hooks

brmenchl:
### Summary
Currently `render` and `fireEvent` are wrapped in `act` to support hooks. This pull request wraps the update call in `act` to support testing component updates (e.g. prop changes).

### Test plan
* I have added a test for update in `__tests__/act.test.js`.
* A useEffect hook should run again on `update` calls (if the hook dependency array is satisfied).

### DIFF
--- a/src/__tests__/act.test.js
+++ b/src/__tests__/act.test.js
@@ -28,6 +28,14 @@ test('render should trigger useEffect', () => {
   expect(effectCallback).toHaveBeenCalledTimes(1);
 });
 
+test('updaate should trigger useEffect', () => {
+  const effectCallback = jest.fn();
+  const { update } = render(<UseEffect callback={effectCallback} />);
+  update(<UseEffect callback={effectCallback} />);
+
+  expect(effectCallback).toHaveBeenCalledTimes(2);
+});
+
 test('fireEvent should trigger useState', () => {
   const { getByTestId } = render(<Counter />);
   const counter = getByTestId('counter');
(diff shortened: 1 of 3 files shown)


src/__tests__/act.test.js
```diff
@@ -28,6 +28,14 @@ test('render should trigger useEffect', () => {
   expect(effectCallback).toHaveBeenCalledTimes(1);
 });
 
+test('updaate should trigger useEffect', () => {
```
thymikee: typo: `update`

src/render.js
```diff
@@ -46,6 +46,15 @@ function renderWithAct(
   return ((renderer: any): ReactTestRenderer);
 }
 
+function updateWithAct(renderer: ReactTestRenderer) {
+  function updateImpl(component: React.Element<any>) {
```
thymikee: we can `return` here :)

src/__tests__/act.test.js
brmenchl: oof good catch

brmenchl:
Thanks for the response! I made a couple updates, let me know if there's anything else outstanding.

Esemesek:
Thank you for PR ❤️ ❤️ ❤️

brmenchl:
Of course! I've got a couple more coming up in the next few days, more type inference for the typescript typings file and a hook specific API a la react-hooks-testing-library (this might require some discussion).

PR #138, opened 2019-03-25
Feat: adding renderHook to test hooks in isolation

brmenchl:
### Summary

Taking a lot of inspiration from https://github.com/mpeyper/react-hooks-testing-library/, this PR adds a `renderHook` function to render React hooks in isolation. This will remove the need to render a component just to test the behavior and return values for custom hooks.

Under the hood, `renderHook` renders a minimal component called the `HookTestHarness` that executes the hook under test, stores the hook return value, and ultimately renders nothing. 

I made this light wrapper for use in our project at Peloton, and it's helped us a ton so far. We have some integration tests that include both a view component and its associated hook, but if the point of hooks is to isolate and share logic, we should have a simple way of testing them without having to create numerous fake components just to successfully execute them.

This is the super minimal structure we're using right now, I'm open to suggestions for more features and improvements.

### Usage
```javascript
const { getValue, rerender, unmount } = renderHook(myHook, firstParam, secondParam, ...);
```

* API:
  * `renderHook: (hook: (...params: HP) => R, initialParams: ...HP) => RenderHookAPI`
  * `RenderHookAPI`:
    * `getValue: () => any` - returns the latest return value from the hook under test.
    * `rerender: (...newParams: Array<any>) => void` - updates the `HookTestHarness` with the new passed in props. This will re-execute the hook under test with new params.
    * `unmount: () => void` - unmounts the `HookTestHarness`, triggering any hook cleanup behavior.

### Test plan
* I have included tests for all API methods, typescript types, and typings tests. Tests can be found in `__tests__/renderHook.test.js`
* I'm not super familiar with advanced Flow typing, so I wasn't able to get as much type inference as I was in the Typescript types. In Typescript, the RenderHookAPI object is type safe with regards to the return value and parameters of the hook under test.

### DIFF
--- a/src/__tests__/renderHook.test.js
+++ b/src/__tests__/renderHook.test.js
@@ -1,0 +1,63 @@
+import { useState, useEffect } from 'react';
+import { renderHook } from '../renderHook';
+
+describe('renderHook', () => {
+  test('should call the hook', () => {
+    const spy = jest.fn();
+    renderHook(spy);
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+
+  test('should trigger cleanup behavior on unmount', () => {
+    const hookCleanup = jest.fn();
+    const { unmount } = renderHook(() => {
+      useEffect(() => {
+        return hookCleanup;
+      });
+    });
+    expect(hookCleanup).not.toHaveBeenCalled();
+    unmount();
+    expect(hookCleanup).toHaveBeenCalledTimes(1);
+  });
+
+  test('should rerun hook on rerender', () => {
+    let renderCount = 0;
+    const { rerender } = renderHook(() => {
+      useEffect(() => {
+        renderCount++;
+      });
+    });
+
+    expect(renderCount).toBe(1);
+    rerender();
+    expect(renderCount).toBe(2);
+  });
+
+  test('should pass new parameters to hook on rerender', () => {
+    let passedParameters;
+    const { rerender } = renderHook(parameters => {
+      useEffect(() => {
+        passedParameters = parameters;
+      });
+    }, 'banana');
+
+    expect(passedParameters).toBe('banana');
+    rerender('split');
+    expect(passedParameters).toBe('split');
+  });
+
+  test('getValue should return latest hook execution result', () => {
+    const { getResult } = renderHook(() => {
+      const [state, setState] = useState(0);
+
+      return {
+        state,
+        setState,
+      };
+    });
+
+    expect(getResult().state).toBe(0);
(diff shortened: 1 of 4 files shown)


src/__tests__/renderHook.test.js
```diff
@@ -0,0 +1,63 @@
...
+describe('renderHook', () => {
+  test('should call the hook', () => {
+    const spy = jest.fn();
+    renderHook(spy);
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+
+  test('should trigger cleanup behavior on unmount', () => {
+    const hookCleanup = jest.fn();
+    const { unmount } = renderHook(() => {
```
thymikee: can you extract these inline hooks to named custom ones? tests serve as documentation and it would be easier for users to adapt if they see how to test imported hook

src/renderHook.js
```diff
@@ -0,0 +1,32 @@
+// @flow
+import * as React from 'react';
+import render from './render';
+
+export const renderHook = <THook: (...arg: Array<mixed>) => mixed>(
```
thymikee: the `renderHook` is not exported from package entry point

src/renderHook.js
```diff
@@ -0,0 +1,32 @@
...
+  const getResult = () => result;
+
+  const { unmount, update } = render(
+    <HookTestHarness parameters={initialParams} />
+  );
+
+  return {
+    getResult,
+    unmount,
+    rerender: (...newParams: HookParameters) => {
```
thymikee: how about `update` instead of `rerender`?

src/__tests__/renderHook.test.js
```diff
@@ -0,0 +1,63 @@
...
+      });
+    }, 'banana');
+
+    expect(passedParameters).toBe('banana');
+    rerender('split');
+    expect(passedParameters).toBe('split');
+  });
+
+  test('getValue should return latest hook execution result', () => {
+    const { getResult } = renderHook(() => {
```
thymikee: I think it would be nice to have a one kinda-complex hook that we could test from every angle, which would help us visualize a proper API for testing it.

thymikee:
Ok, this stalled for a while and now I believe [react-hooks-testing-library](https://github.com/mpeyper/react-hooks-testing-library) is a proper way to test custom hooks – it's quite specialized in what it does and also it doesn't have deps on `react-dom` (or anything unsupported by React Native) so it's safe to use in this context. 

Thanks for your PR though, it's greatly appreciated 🙇

Issue #99, opened 2018-12-21
jest-dom extend-expect


maitriyogin:
Hi,
love this library!
just wondering if your planning to get the jest-dom/extend-expect functions converted to Native?
I have a bunch of tests where I introspect text results ...
I can do a getByText but in some cases I need to get all text for a parent Text component that contains a bunch of child Text components.
Thanks for your help,
/Stephen.

thymikee:
Sounds like cool thing to have. Here's a list of all matchers: https://github.com/gnapse/jest-dom#custom-matchers

Does anybody feel like implementing this for native? :)

brooksbecton:
I started something here: https://github.com/brooksbecton/react-native-jest-expect-extensions
as a proof of concept with a `.toBeDisabled` extension

Would this be part of 'react-native-testing-library' or would it be a separate repo?

thymikee:
I think @Esemesek started something around that as well, so feel free to collaborate :). I'd prefer to have the matchers live in this repository.

It's important to make them output high-quality error messages – I encourage you to take a look at Jest matchers recently getting improved by @pedrottimark, e.g. https://github.com/facebook/jest/blob/7e4e170023b29f1e78dbe757b12b291443a53c6a/packages/expect/src/matchers.ts#L200-L221.

brooksbecton:
I've got a few extensions going. Were ya'll wanting all of them in before I PR? 

Also, where would you want these to go in the project?

thymikee:
I'd create a directory `src/custom-matchers` and store them there.

codenaz:
Use @testing-library/jest-native. Then import like this
 ```import '@testing-library/jest-native/extend-expect';```

thymikee:
Closing in favor of #298. I see some of you already use `@testing-library/jest-native` so I hope it will only be a matter of documenting it and providing some tests that we're compatible :)

PR #141, opened 2019-03-28
Feat: Custom matchers

brooksbecton:
### Summary

Begins work on custom jest matchers as part of issue #99 

### Test plan

`yarn test`


Esemesek:
@thymikee I don't think we want this `package-lock.json` committed.

src/__tests__/custom-matchers.test.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
```
thymikee: Please add `// @flow` pragma

src/custom-matchers/extendExpect.js
```diff
@@ -0,0 +1,3 @@
+import extensions from './index';
```
thymikee: `// @flow`

src/custom-matchers/index.js
```diff
@@ -0,0 +1,4 @@
+import { toBeDisabled } from './toBeDisabled';
```
thymikee: `// @flow`

src/__tests__/custom-matchers.test.js
```diff
@@ -0,0 +1,65 @@
...
+import { Button, Text, TextInput, View } from 'react-native';
+import { create } from 'react-test-renderer';
+
+describe('custom-matchers', () => {
+  function render(node) {
```
thymikee: can we use `render` from this library?

brooksbecton:
- Removed the package-lock :sweat_smile: 
- Added `// @flow`
- Using `react-native-testing-library`'s render function 

I found Esemesek's branch with the custom-matchers here: https://github.com/callstack/react-native-testing-library/tree/Esemesek/custom-matchers/src/matchers

and I used most of it's .toBeEmpty logic because it is way more thorough

brooksbecton:
Removed package-lock.json. 

Were ya'll wanting to get most of the matchers that are in jest-dom in this PR?

thymikee:
I'd rather like to start small :)

src/__tests__/custom-matchers.test.js
```diff
@@ -0,0 +1,125 @@
...
+    describe('Button', () => {
+      const BUTTON_EXPLICIT_DISABLED = 'button-explicit-disabled';
+      const BUTTON_IMPLICIT_DISABLED = 'button-implicit-disabled';
+      const BUTTON_EXPLICIT_ENABLED = 'button-explicit-enabled';
+      const BUTTON_IMPLICIT_ENABLED = 'button-implicit-enabled';
+      const TestComponent = () => (
+        <View>
+          <Button
+            testID={BUTTON_EXPLICIT_DISABLED}
+            disabled={true}
```
thymikee: This case can be removed, it's the same as below from jsx point of view

src/custom-matchers/toBeEmpty.js
```diff
@@ -0,0 +1,28 @@
+// @flow
+
+function isEmpty(instance: ReactTestInstance) {
+  if (instance.props.children instanceof Array) {
```
thymikee: We should use React.Children for iteration over children

brooksbecton: Are you saying we should check to see if it is instanceof React.Children? 
I don't understand

src/__tests__/custom-matchers.test.js
brooksbecton: Removed the 'BUTTON_EXPLICIT_DISABLED' test case

thymikee:
Thank you so much for your time! I'll close this in favor of #298 though. We're going to use `@testing-library/jest-native`. I hope it will only be a matter of documenting it and providing some tests that we're compatible :). If we want to add more matchers, we can contribute it back there :)

Issue #111, opened 2019-02-06
Wrap render calls with act() for better Hooks integration


thymikee:
Hooks work fine even without it, but you'll likely need to add asynchrony to your code. With `act()` you'll be able to achieve the same results in synchronous manner.

See: https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks

ccfz:
It would be great if `update` is also be wrapped with act().

Esemesek:
@ccfz Would appreciate a PR with an example test for this functionality.

PR #146, opened 2019-04-03
Add updateWithAct to render

ccfz:
### Summary

To test behaviour that only occurs when props are updated it is
necessary to call update on the test instance. However, if the
behaviour under test is triggered by a hook, e.g. useEffect then
the test will fail because the hook is only triggered after the test
has run.

https://github.com/callstack/react-native-testing-library/issues/111

### Test plan

```javascript
describe('when FloatLabel becomes active', () => {
  it('updates the Text color from dark to light', () => {
    const wrapper = render(<FloatLabel active={false} />)

    expect(expect(text.props.style).toMatchObject({ color: 'white' });)

    act(() => wrapper.update(<FloatLabel active />));
   // with the changes it will simply work by calling update

    expect(text.props.style).toMatchObject({ color: 'white' });
  });
});
```

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -26,11 +26,15 @@ export default function render(
   return {
     ...getByAPI(instance),
     ...queryByAPI(instance),
-    update: renderer.update,
+    update: (element) => updateWithAct(element, renderer),
     unmount: renderer.unmount,
     toJSON: renderer.toJSON,
     debug: debug(instance, renderer),
   };
+}
+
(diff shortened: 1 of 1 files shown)


thymikee:
Thank you! If you rebased to latest master, you'd see this is already fixed by https://github.com/callstack/react-native-testing-library/pull/137 and released as 1.6.2 just now :)

Looking forward to future contributions!

PR #178, opened 2019-05-27
feat: add a11y queries

Esemesek:
### Summary

Addition of accessibility queries to the API. Available queries:
- `byA11yLabel`, `byAccessibilityLabel`
- `byA11yHint`, `byAccessibilityHint`
- `byA11yStates`, `byAccessibilityStates`
- `byA11yRole`, `byAccessibilityRole`

### TODO:
- [x] - Add documentation

### Test plan

Unit tests included in this PR.


src/helpers/makeQuery.js
```diff
@@ -0,0 +1,85 @@
+// @flow
+import { ErrorWithStack } from './errors';
+
+const prepareErrorMessage = error =>
```
Esemesek: This is copied and pasted. Import this thing instead.

thymikee: Fixed that for you

src/helpers/a11yAPI.js
```diff
@@ -0,0 +1,98 @@
...
+  // $FlowFixMe
+  return !matcher.some(e => !prop.includes(e));
+}
+
+const a11yAPI = (instance: ReactTestInstance): A11yAPI =>
+  ({
+    ...makeQuery(
+      'accessibilityLabel',
+      {
+        getBy: 'getByA11yLabel',
```
Esemesek: We could also replace `strings` with `Array<string>` to have more than one alias for a query.
E.g. `getByA11yLabel` and `getByAccessibilityLabel`. @thymikee wdyt?

thymikee: I'm down 👍

PR #198, opened 2019-07-03
 feat: add getAllByTestId and queryAllByTestId queries

### LINKED ISSUE #161: getAllByTestId is missing
## Versions

    "react": "16.8.3",
    "react-native": "0.59.0",
    "react-native-testing-library": "^1.6.0",
    "react-test-renderer": "^16.8.4"

## Description
getAllByTestId appears to be missing, it is listed as a query in the [docs](https://callstack.github.io/react-native-testing-library/docs/api-queries#bytestid).

Screenshot 2019-04-24 at 13 28 58

### src: 
```
import React from 'react';
import { Text, View } from 'react-native';

const OptionList = ({ options }) => {
  const renderOptions = (optionList) => {
    return optionList.map(visibleOption => (
      <Text
        key={option.id}
        testID="optionContainer"
      >
        {option.text}
      </Text>
    ));
  };

  return (
    <View>
      {renderOptions(options)}
    </View>
  );
};

export default OptionList;
```

### spec:
```
import React from 'react';
import { Text } from 'react-native';
import { render } from 'react-native-testing-library';

import OptionList from 'src/components/OptionList';

const buildTestProps = props => ({
  options: [
    { id: 1, text: 'Frodo Card' },
    { id: 2, text: 'Gandalf Card' },
    { id: 3, text: 'Sauron Card' },
    { id: 4, text: 'Elrond Card' },
    { id: 5, text: 'Pippin Card' }
  ],
  ...props
});

let props;
let wrapper;

beforeAll(() => {
  props = buildTestProps();
  wrapper = render(<SwipeSelect {...props} />);
});

describe('SwipeSelect', () => {
  it('renders the options', () => {
    expect(wrapper.getAllByTestId(Text).length).toEqual(5);
  });
});

```

OzzieOrca:
### Summary
Implements `getAllByTestId` and `queryAllByTestId` queries.
Closes #186, Resolves #161

### Test plan
I wrote a test and added two `<Text>` components to the component under test as a simple, contrived example. If you want me to think of a more realistic use case or you want to suggest one I'd be happy to try that. Maybe buttons in a loop or a child component or something. But I don't want to add complexity to the example component.

### DIFF
--- a/src/__tests__/__snapshots__/render.test.js.snap
+++ b/src/__tests__/__snapshots__/render.test.js.snap
@@ -45,6 +45,16 @@ exports[`debug 1`] = `
       Change freshness!
     </Text>
   </View>
+  <Text
+    testID=\\"duplicateText\\"
+  >
+    First Text
+  </Text>
+  <Text
+    testID=\\"duplicateText\\"
+  >
+    Second Text
+  </Text>
 </View>"
 `;
 
@@ -93,6 +103,16 @@ exports[`debug changing component: bananaFresh button message should now be "fre
       Change freshness!
     </Text>
   </View>
+  <Text
+    testID=\\"duplicateText\\"
+  >
+    First Text
+  </Text>
+  <Text
+    testID=\\"duplicateText\\"
+  >
+    Second Text
+  </Text>
 </View>"
 `;
 
@@ -128,6 +148,16 @@ exports[`debug: shallow 1`] = `
   >
     Change freshness!
   </Button>
+  <Text
+    testID=\\"duplicateText\\"
+  >
+    First Text
+  </Text>
+  <Text
+    testID=\\"duplicateText\\"
+  >
+    Second Text
+  </Text>
 </View>"
 `;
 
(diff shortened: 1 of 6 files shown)


OzzieOrca:
Just noticed flow is yelling at me. Ya I'll look at the TS types. Forgot about them...

OzzieOrca:
You have any suggestions for fixing the flow error? I'm not very familiar with flow. Seems like it doesn't know that `instance.findAllByProps` returns `ReactTestInstance[]`.

thymikee:
You'll need too add a type annotation to filter that it returns a single ReactTestInstance, something like `filter<ReactTestInstance>()`

OzzieOrca:
That is a very unhelpful error telling you to specify a return type. I tried something like that and it wasn't happy...

thymikee:
Would be nice to also update this file 😅 https://github.com/callstack/react-native-testing-library/blob/master/typings/__tests__/index.test.tsx

OzzieOrca:
Didn't realize what that was doing...

thymikee:
Thank you so much!

OzzieOrca:
Any chance you could create a release for this? I could pin it to a master commit for a while but I'd rather be done with it. But I realize you probably don't want a new release for every little thing. Do what you need to.

thymikee:
We'll publish it tomorrow :)

thymikee:
Released 1.10

PR #223, opened 2019-09-24
docs: add fireEvent.scroll example for FlatList

ajubin:
### Summary

When testing scroll on a flatlist, I was getting the error : `TypeError: cannot read property 'height' of undefined`

#### Error thrown
image

#### Component tested

```
export class InfiniteScrollFlatList<Item> extends PureComponent<Props<Item>> {
  private handleLoadMore = () => {
    if (!this.props.isLoadingMore && this.props.nextPage) {
      this.props.fetchMore({ page: this.props.nextPage });
    }
  };

  private renderListFooter = () => {
    if (this.props.isLoadingMore) return <Loader />;
    return null;
  };

  public render() {
    return (
      <FlatList
        {...this.props}
        onEndReachedThreshold={0.2}
        onEndReached={this.handleLoadMore}
        ListFooterComponent={this.renderListFooter}
      />
    );
  }
}
```

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -211,8 +211,10 @@ fireEvent.changeText(getByTestId('text-input'), CHANGE_TEXT);
 
 Invokes `scroll` event handler on the element or parent element in the tree.
 
+#### On a `ScrollView`
+
 ```jsx
-import { ScrollView, TextInput } from 'react-native';
+import { ScrollView, Text } from 'react-native';
 import { render, fireEvent } from 'react-native-testing-library';
 
 const onScrollMock = jest.fn();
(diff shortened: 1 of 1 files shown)


thymikee:
Hey! Thanks for sending that. The example though is about `ScrollView`, not `FlatList`. Could you verify that the scroll event is the same for scroll view and flat list?

ajubin:
Hello,

My example here is about `FlatList`. The `contentSize` and the `layoutMeasurement` are not needed for `ScrollView`. Should I had another example ?

thymikee:
Please add another example then :) Feel free to only mention the scroll event data differs between components, and e.g. for FlatList it looks like this.

ajubin:
@thymikee I've updated my pr

Issue #234, opened 2019-11-18
Automatic unmount of rendered components at end of each test


aleclarson:
## Describe the Feature
I'm testing hooks that modify the global state of a library, so they need to be unmounted before the next test begins. Currently, I'm using this workaround:
```ts
import { render as renderImpl, RenderAPI } from 'react-native-testing-library'

const rendered: RenderAPI[] = []
const render: typeof renderImpl = (...args) => {
  const elem = renderImpl(...args)
  rendered.push(elem)
  return elem
}

beforeEach(() => {
  rendered.forEach(elem => elem.unmount())
  rendered.length = 0
})
```

Ideally, `react-native-testing-library` would export a function that lets developers opt-in to this behavior, like the following:

```ts
import { setAutoUnmount } from 'react-native-testing-library'

setAutoUnmount(true)
``` 

Then I wouldn't need to maintain this code myself. :)

## Possible Implementations
See the workaround.

## Related Issues
None.

thymikee:
> so they need to be unmounted before the next test begins

Um, why not leveraging Jest APIs then?:

```js
describe('testing hook that messes with globals', () => {
  let component;
  beforeEach(() => component.unmount())
	
  test('1', () => {
    component = render(element1)
  })

  test('2', () => {
    component = render(element2)
  })
})
```

aleclarson:
Good question! That's certainly a valid solution, but all of my tests are actually ignoring the return value of `render`, so it's merely bloat to assign to `component` when I'm not using it.

thymikee:
I see that RTL has something similar built-in and it's called `cleanup`: https://testing-library.com/docs/react-testing-library/api#cleanup 

Could you evaluate if this is the same use case and we could mimic its behavior in RNTL?

aleclarson:
That looks perfect! nice find

thymikee:
I'd be happy to get this onboard and turn on by default in v2

aleclarson:
v1 #237
v2 #238

thymikee:
Closed via #238

PR #235, opened 2019-11-18
feat: add "autoUnmount" function

aleclarson:
### Summary

See #234

### Test plan

None yet

### DIFF
--- a/src/autoUnmount.js
+++ b/src/autoUnmount.js
@@ -1,0 +1,12 @@
+import render from './render';
+
+export default function autoUnmount() {
+  let queue;
+  beforeEach(() => {
+    queue = render.unmountQueue = [];
+  });
+  afterEach(() => {
+    queue.forEach(unmount => unmount());
+    render.unmountQueue = null;
+  });
+}
--- a/src/index.js
+++ b/src/index.js
@@ -1,16 +1,18 @@
 // @flow
 import act from './act';
-import render from './render';
-import shallow from './shallow';
-import flushMicrotasksQueue from './flushMicrotasksQueue';
+import autoUnmount from './autoUnmount';
 import debug from './debug';
 import fireEvent from './fireEvent';
+import flushMicrotasksQueue from './flushMicrotasksQueue';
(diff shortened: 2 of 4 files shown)


src/autoUnmount.js
```diff
@@ -0,0 +1,13 @@
...
+export default function autoUnmount() {
+  let queue;
+  beforeEach(() => {
+    queue = render.unmountQueue = [];
+  });
+  afterEach(() => {
+    render.unmountQueue = null;
+    queue.forEach(unmount => unmount());
+    queue.length = 0;
+  });
```
aleclarson: This lets us call `autoUnmount` within a `describe` block to enable auto-unmounting for a subset of tests.

You can also call it from the top-level if you want auto-unmounting for all tests in the file.

thymikee:
Thanks for your PR, appreciate the attitude.

However, I'm not entirely sure about this feature. If you can have this functionality with <10 LOC, why adding it to the library and make us maintain it? I'd rather like to first prove that such a feature is more commonly requested by community first (e.g. in form of many upvotes on the original issue, or a library), then think about implementation. Especially that it's side-effectful and bound to the whole `render` module, making it non composable and available for every test case, even if you don't want it. 

Let's close it for now and keep the discussion going.

aleclarson:
> I'd rather like to first prove that such a feature is more commonly requested by community first

Good point! Another counterpoint is that it's using Jest-specific globals.

> Especially that it's side-effectful and bound to the whole `render` module, making it non composable and available for every test case, even if you don't want it.

Nope, as I tried to clarify in the OP, you can call it inside `describe` blocks too. It unplugs from the `render` module at the end.

thymikee:
> Nope, as I tried to clarify in the OP, you can call it inside describe blocks too. It unplugs from the render module at the end.

Missed that, thanks for point out. Still, I'm not convinced it's a good API and worth getting it into the library, since it can be done entirely in user code. I believe in small API surface area and won't add anything that doesn't add clear value for lots of users.

PR #237, opened 2019-11-19
feat: add "cleanup" function

aleclarson:
### Summary

https://github.com/callstack/react-native-testing-library/issues/234#issuecomment-555725623

Every `render` call stores its `unmount` function in a global `Set` called `cleanup.queue`. This is how RTL does it (except it doesn't expose the queue). I exposed the queue on `cleanup` because it was easiest. You can adjust that in a follow-up commit if you want.

The `cleanup` function is **never** called automatically in this PR.

### Test plan

https://github.com/callstack/react-native-testing-library/commit/f08f132a202eaaf88019dab7c40c855105afd2af

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -115,6 +115,45 @@ toJSON(): ReactTestRendererJSON | null
 
 Get the rendered component JSON representation, e.g. for snapshot testing.
 
+## `cleanup`
+
+```ts
+const cleanup: () => void
+```
+
+Unmounts React trees that were mounted with `render`.
+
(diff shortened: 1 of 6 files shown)


aleclarson:
ping @thymikee

aleclarson:
Can this be published to npm now?

thymikee:
@aleclarson yup, just released v1.12. Thanks for this contribution!

hossamnasser938:
Should we update the readme to say that `cleanup` now exists?

PR #254, opened 2020-02-20
feat(website): add DocSearch as recommended by docusaurus.

s-pace:
👋 team,

I am working on DocSearch. [We have an integration for Docusaurus](https://docusaurus.io/docs/en/search#enabling-the-search-bar). We thought it is a shame you can not also used this search that you can [find on reactjs for example](https://reactjs.org/).

This PR will add DocSearch to the documentation website. It will allow an user to have a learn-as-you-type experience by displaying results thanks to a dropdown in a live way.

Let me know if you need anything.

### DIFF
--- a/website/siteConfig.js
+++ b/website/siteConfig.js
@@ -84,6 +84,12 @@ const siteConfig = {
   // template. For example, if you need your repo's URL...
   repoUrl,
   usePrism: ['jsx'],
+  
+  // search
+  algolia: {
+    apiKey: 'cd9b8b73f97b64ed04570e41c507683f',
+    indexName: 'react-native-testing-library',
+  },
 };
 
 module.exports = siteConfig;


website/siteConfig.js
```diff
@@ -84,6 +84,12 @@ const siteConfig = {
   // template. For example, if you need your repo's URL...
   repoUrl,
   usePrism: ['jsx'],
+  
+  // search
+  algolia: {
+    apiKey: 'cd9b8b73f97b64ed04570e41c507683f',
```
thymikee: where this key comes from?

s-pace: This is randomly created during the application https://community.algolia.com/docsearch/apply.html

Feel free to provide me your email address so that I grant you access to the analytics we provide for free.

thymikee: Thanks! Just applied for a key, it may be interesting to see the analytics :)

s-pace: Great, invited!

thymikee:
For some reasons I can't push to your remote, but I'd like to keep your attribution. Here's my API key, feel free to update it: `c8b531ec509d4936fad12fa9bd078429`

s-pace:
@thymikee thanks for the update. What is this key? You can use the one I have set here as it is a public one. You will have the same analytics. This PR is ready.

thymikee:
Hey @s-pace, how are you doing? :) I recently noticed that the searches are a bit out of date. We've released a new version and added a few APIs and they don't show up. How can we trigger re-crawling the searching index?

s-pace:
👋 @thymikee ,

Should be good now :)

thymikee:
@s-pace thanks, works like a charm 👍

PR #259, opened 2020-03-16
Implement a11yState and a11yValue queries

Esemesek:


package.json
```diff
@@ -15,25 +15,25 @@
     "integration"
   ],
   "devDependencies": {
-    "@babel/cli": "^7.1.2",
-    "@babel/core": "^7.1.2",
```
thymikee: can we bump dev deps in a separate PR to make it smaller?

Esemesek: Yup

thymikee:
Can you make sure CI passes?

Esemesek:
That is what I am working on right now :P

Esemesek:
Will split this thing into couple of PRs

thymikee:
Thanks, looking forward to it!

PR #260, opened 2020-03-17
feat: byA11yState matcher

Esemesek:
### Summary

Introduction of `byA11yState` matcher and improvement of types.

### Test plan

Added UT and Typescript tests

### DIFF
--- a/src/helpers/a11yAPI.js
+++ b/src/helpers/a11yAPI.js
@@ -70,14 +70,11 @@ export function matchArrayValue(
 }
 
 export function matchObject<T: {}>(prop?: T, matcher: T): boolean {
-  return (
-    Boolean(prop) &&
-    Object.keys(matcher).length !== 0 &&
-    // $FlowFixMe Already checked
-    Object.keys(prop).length !== 0 &&
-    // $FlowFixMe Already checked
-    !Object.keys(matcher).some(key => prop[key] !== matcher[key])
-  );
+  return prop
+    ? Object.keys(matcher).length !== 0 &&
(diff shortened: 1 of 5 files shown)


src/helpers/a11yAPI.js
```diff
@@ -58,6 +69,17 @@ export function matchArrayValue(
...
 
+export function matchObject<T: {}>(prop?: T, matcher: T): boolean {
+  return (
+    Boolean(prop) &&
+    Object.keys(matcher).length !== 0 &&
+    // $FlowFixMe Already checked
+    Object.keys(prop).length !== 0 &&
+    // $FlowFixMe Already checked
+    !Object.keys(matcher).some(key => prop[key] !== matcher[key])
+  );
```
thymikee: This would get rid of fixmes
```suggestion
  return prop
    ? Object.keys(matcher).length !== 0 &&
        Object.keys(prop).length !== 0 &&
        !Object.keys(matcher).some(key => prop[key] !== matcher[key])
    : false;
```

src/types.flow.js
```diff
@@ -0,0 +1,40 @@
+// @flow
+
+export type A11yRole =
```
thymikee: Can we add a reference link? https://reactnative.dev/docs/accessibility#accessibilityrole-ios-android
Ideally we could import directly from `react-native`, but I'm not sure if it's exported and if it would be backward compatible with older RN versions. This applies to TS as well

thymikee:
cc @petermakowski

PR #261, opened 2020-03-17
imp: Re-alias deprecated matchers and warn

Esemesek:
### Summary

Re-alias every deprecated method with `UNSAFE_` prefix and add a warning.
For now, I decided not to include any warnings for re-aliased functions (but we should definitely include some messages after we delete original methods)

cc @thymikee 

### Test plan

Everything is logged correctly.

https://app.circleci.com/pipelines/github/callstack/react-native-testing-library/758/workflows/e3d488a4-dd12-4e68-9bae-7cbc7951bd96/jobs/2865/parallel-runs/0/steps/0-102

### DIFF
--- a/src/helpers/errors.js
+++ b/src/helpers/errors.js
@@ -13,30 +13,6 @@ export const createLibraryNotSupportedError = (error: Error) =>
     `Currently the only supported library to search by text is "react-native".\n\n${error.message}`
   );
 
-const warned = {
-  getByName: false,
-  getAllByName: false,
-  queryByName: false,
-  queryAllByName: false,
-};
-
-export const logDeprecationWarning = (
-  deprecatedFnName: string,
-  alternativeFnName: string
-) => {
-  if (warned[deprecatedFnName]) {
-    return;
-  }
-  console.warn(`Deprecation Warning:
-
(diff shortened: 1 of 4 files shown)


src/helpers/queryByAPI.js
```diff
@@ -82,20 +84,23 @@ export const queryByTestId = (instance: ReactTestInstance) =>
     }
   };
 
-export const queryAllByName = (instance: ReactTestInstance) => (
-  name: string | React.ComponentType<any>
-) => {
-  logDeprecationWarning('queryAllByName', 'getAllByName');
```
thymikee: we've long-deprecated ByName matchers, so I'd like to actually remove this API, not move it to UNSAFE_

src/helpers/errors.js
```diff
@@ -13,37 +13,44 @@ export const createLibraryNotSupportedError = (error: Error) =>
...
+  queryByType: false,
+  queryAllByType: false,
 };
 
-export const logDeprecationWarning = (
-  deprecatedFnName: string,
-  alternativeFnName: string
-) => {
-  if (warned[deprecatedFnName]) {
+export function printDeprecationWarning(functionName: string) {
```
thymikee: let's rename this to `printMigrationToUnsafeWarning` and `logDeprecationWarning` -> `printDeprecationWarning`, essentially keeping them both

PR #269, opened 2020-03-18
chore: fix inconsistent flow fails on CI

Esemesek:
### Summary

Let's try to fix this flow issues on CI.

### Test plan

Flow passing 10/10 on CI.


Esemesek:
It seems to fix the issue on CI.

Issue #28, opened 2018-10-12
fireEvent seems to always trigger event (on disabled TouchableOpacity)


fwal:
I'm trying to figure out why it seems like the `fireEvent.press` function always seem to trigger a callback even though it shouldn't be registered. My use case is that I'm trying to write a test that makes sure that `onPress` isn't called if the component is "disabled".

Trying to mimic the behaviour below...
```js
  const MyComponent = _props => (
    <View>
      <TouchableHighlight testID={'touchable'}>
        <Text>{'Foo'}</Text>
      </TouchableHighlight>
    </View>
  )

  const pressedCallback = jest.fn()
  const { getByTestId } = render(<MyComponent onPress={() => pressedCallback('bar')} />)
  fireEvent.press(getByTestId('touchable'))
  expect(pressedCallback).not.toBeCalled() //Expected mock function not to be called but it was called with: ["bar"]
```

fwal:
Hmm seems like `fireEvent` traverses up the tree and call the function if found, that would make my use-case a bit trickier... https://github.com/callstack/react-native-testing-library/blob/master/src/fireEvent.js#L18

thymikee:
Yup, that's a desired behavior – user may press a text, which in fact may trigger a parent press event handler.

fwal:
Any idea on how I could test my behaviour? I basically don't set `onPress` on the touchable component if my prop `disabled` is true...

thymikee:
cc @Esemesek

thymikee:
I think it makes sense to stop bubbling just before root component, that would prevent this and seems like a sane solution

fwal:
Ah I ~could~ probably should swallow the `onPress` event in the TouchableHighlight if disabled is true rather than omitting to set the prop... 🤔

fwal:
For example this works:
```js
  const MyComponent = ({ disabled, onPress }) => (
    <View>
      <TouchableHighlight onPress={disabled ? () => {} : onPress} testID={'touchable'}>
        <Text>{'Foo'}</Text>
      </TouchableHighlight>
    </View>
  )
```

But then again I'm changing implementation to satisfy the test 😅

thymikee:
I usually do both just in case, but I guess we could detect the "disabled" prop on the Core RN components like `Touchable*`.

fwal:
[...]

PR #273, opened 2020-04-03
chore: bump RN to 0.62

thymikee:
### Summary

Blocked by https://github.com/facebook/react-native/pull/28520

### Test plan

CI passes


src/__tests__/fireEvent.test.js
```diff
@@ -57,7 +57,9 @@ describe('fireEvent', () => {
   });
 
   test('should throw an Error when event handler was not found', () => {
-    const { getByTestId } = render(<WithoutEventComponent />);
+    const { getByTestId } = render(
+      <WithoutEventComponent onPress={() => 'this is not passed to children'} />
```
thymikee, line 61: Based on https://github.com/callstack/react-native-testing-library/issues/28 I've added this instead of the `should not bubble event to root element` test that started to fail

src/__tests__/fireEvent.test.js
```diff
@@ -78,18 +80,6 @@ describe('fireEvent', () => {
...
-    const onPressMock = jest.fn();
-    const { getByTestId } = render(
-      <TouchableOpacity onPress={onPressMock}>
-        <Text testID="test">Content</Text>
-      </TouchableOpacity>
-    );
-
-    expect(() => fireEvent.press(getByTestId('test'))).toThrow();
-    expect(onPressMock).not.toHaveBeenCalled();
-  });
```
thymikee, line 92: This test seems fishy and not testing what it's supposed to. Not really sure why it didn't fail earlier, since the `test` testID is there, we can press it, and closest `onPress` is bound, so it should be called.

PR #277, opened 2020-04-17
docs: add example of testing React Navigation

cross19xx:
### Summary

This PR includes
- setup an example of testing React Navigation navigator with default screen
- integration with the website

Part of #280


website/i18n/en.json
```diff
@@ -21,6 +24,7 @@
     },
     "categories": {
       "Introduction": "Introduction",
+      "Examples": "Examples",
```
thymikee, line 27: let's put Examples below API Reference

docs/ReduxIntegration.md
```diff
@@ -0,0 +1,12 @@
+---
+id: redux-integration
+title: Redux Integration
+---
+
+Codesandbox provided as proof.
+
+<iframe
+  src="https://codesandbox.io/embed/new?codemirror=1"
```
thymikee, line 9: I requested creating an account, will update this PR once done

thymikee: We have an account now. So a step forward to this, would be for you to create a codesandbox with Redux example, which I could then fork to the official account

website/i18n/en.json
cross19xx: Done

thymikee:
We'll also need to set up running tests for examples, let's do it in a followup

PR #284, opened 2020-04-23
chore: update GettingStarted example with something you can run

thymikee:
### Summary

The example we had was based on some 3rd party code without source. Let's change to so it's easier for people to give the library a try. I've put the implementation into our test suite.

### Test plan

Added a new test for QuestionBoard from the docs.


thymikee:
CI is failing because of GitHub issues, will try tomorrow

cross19xx:
It shows here that CI passed.

Issue #275, opened 2020-04-14
Ability to query by text that is not a single node.


Trancever:
## Describe the Feature

It is very common to wrap one text into another to apply some additional styles. Unfortunately, this doesn't work well with testing. Currently, there is no way to query the whole text e.g.

```
<Text>Hello <Text>World</Text></Text>
```

currently `queryByText('Hello World')` won't find the text, but it would be really handy if it finds it.

There is also an option to change the structure of rendered components and render them one by one e.g.

```
<Text>Hello </Text><Text>World</Text>
```

but even though we do it, we have to query each text separately which doesn't make sense in most cases.

I believe `queryByText('Hello World')` should also somehow find the text that is divided into two sibling components. 

## Possible Implementations
I have no idea, but react-testing-library supports it by accepting custom matching function in getByText. Maybe it is possible to implement it in a similar way?

PR #290, opened 2020-04-28
Fixed an issue where xByText would call toString() on objects and only match on `[object Object]`

Vannevelj:
### Summary

This fixes the problem as described in https://github.com/testing-library/native-testing-library/issues/121. In short, `getByText` and `queryByText` would not find a match when the `<Text>` component has a nested component inside of it. This was caused by calling `toString()` on all components it found, meaning something like an `<Image>` would result in `[object Object]`.

A workaround would be to use a regex match for your text instead so you could decide that that is better than this proposed fix, although I would recommend detailing that somewhere as this isn't easy to debug.

### Test plan

I've provided a few unit tests and made sure the code supports a scenario in an existing unit test. That being said, I'm not 100% confident that this change will behave as intended for all scenarios (i.e. we now explicitly take care of `string` and `number` -- are there other data types that should be included)?

### DIFF
--- a/src/__tests__/queryByApi.test.js
+++ b/src/__tests__/queryByApi.test.js
@@ -1,0 +1,49 @@
+// @flow
+import React from 'react';
+import { Text, Image } from 'react-native';
+import { render } from '..';
+
+test('queryByText nested <Image> in <Text> at start', () => {
+  expect(
+    render(
+      <Text>
+        <Image source={{}} />
+        Hello
+      </Text>
+    ).queryByText('Hello')
+  ).toBeTruthy();
+});
+
+test('queryByText nested <Image> in <Text> at end', () => {
+  expect(
+    render(
+      <Text>
+        Hello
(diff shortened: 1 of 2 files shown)


thymikee:
Is this anyhow related to https://github.com/callstack/react-native-testing-library/issues/275?

Vannevelj:
It's related but not the same. The issue I describe here works on a single level, the direct children: with this fix, the Text node and it's siblings are getting an improved string representation. 

The issue described there works on descendants. If you want to fix that, you'd need to go down in the component tree and form a string representation of all components in it. 

The big difference is that I consider a nested component as an empty string whereas in order to fix that issue, you would need to go into the nested component (recursively) and build up a string representation.

I thought it best to keep the change contained to one issue at first but it shouldn't be terribly hard to expand for his scenario. I wasn't quite sure either what the desired functionality was there so I didn't attempt to fix it.

PR #291, opened 2020-05-04
docs: add Redux example

cross19xx:
Examples to include the following:
- [x] Integration with Redux


thymikee:
Yay! Now that we have multiple examples, we need to come up with a way to test them. I think it makes sense to make RNTL a Yarn workspace so that we can install deps once and use Jest's `project` config. The structure would look like this:
```sh
/ # root
  packages/
    react-native-testing-library/ # whole lib moved here
  examples/
    redux/
    react-navigation/
    ...
```

examples/redux/App.test.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { cleanup, fireEvent, render } from 'react-native-testing-library';
+import App from './App';
```
thymikee, line 3: We usually don't test the root App component. I mean, we can, but that's easy and requires no setup. 
What we want to test is a component that renders _inside_ the App, that implicitly relies on Redux. A component that doesn't have Provider inside, hence we need to provide such in a test.

examples/redux/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "redux-example",
+  "description": "Testing Redux interactions with RNTL",
+  "version": "0.0.1",
+  "scripts": {
+    "test": "jest"
+  },
+  "dependencies": {
+    "expo": "~37.0.3",
+    "expo-updates": "~0.1.0",
```
thymikee, line 10: Let's try to be expo-independent. We don't need the apps to be runnable, only tests. We don't have capacity to keep all of that updated and upgrade all examples

examples/redux/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "react": "~16.9.0",
+    "react-dom": "~16.9.0",
+    "react-native": "~0.61.5",
+    "react-native-gesture-handler": "~1.6.0",
+    "react-native-reanimated": "~1.7.0",
+    "react-native-screens": "~2.2.0",
+    "react-native-unimodules": "~0.9.0",
+    "react-native-web": "~0.11.7",
```
thymikee, line 18: redux example doesn't need all these deps, please remove

thymikee:
Let's make this PR only about Redux interaction, remove anything that's not specific to the tests. We need to keep examples short, scoped and to the point to avoid maintenance burden. It's crucial

examples/redux/App.test.js
cross19xx: This has been treated. What I did was to create test cases for the different components that use Redux. For each test component, I created an instance of a store I wanted.

cross19xx:
> Yay! Now that we have multiple examples, we need to come up with a way to test them. I think it makes sense to make RNTL a Yarn workspace so that we can install deps once and use Jest's `project` config. The structure would look like this:
> 
> ```shell
> / # root
>   packages/
>     react-native-testing-library/ # whole lib moved here
>   examples/
>     redux/
>     react-navigation/
>     ...
> ```

This I believe will be cool. But do you agree this has to be done later?

examples/redux/package.json
cross19xx: Done 😉

examples/redux/package.json
cross19xx: Expo has been expelled

PR #293, opened 2020-05-08
feat: query nested text nodes

### LINKED ISSUE #275: Ability to query by text that is not a single node.

marchenk0va:
### Summary

Support `*ByText` to match text for nested `Text` components.
Closes #275

### Test plan

- Nest node into `Text`, eg `<Text/>RN Testing <Text/>Library</Text></Text>`
- Check the existence of 'RN Testing Library' with `getAllByText`

### DIFF
--- a/src/__tests__/queryByApi.test.js
+++ b/src/__tests__/queryByApi.test.js
@@ -3,56 +3,73 @@ import React from 'react';
 import { Text, Image } from 'react-native';
 import { render } from '..';
 
-test('queryByText nested <Image> in <Text> at start', () => {
+// test('queryByText nested <Image> in <Text> at start', () => {
+//   expect(
+//     render(
+//       <Text>
+//         <Image source={{}} />
+//         Hello
+//       </Text>
+//     ).queryByText('Hello')
+//   ).toBeTruthy();
+// });
+
+// test('queryByText nested <Image> in <Text> at end', () => {
+//   expect(
+//     render(
+//       <Text>
+//         Hello
+//         <Image source={{}} />
+//       </Text>
+//     ).queryByText('Hello')
+//   ).toBeTruthy();
+// });
+
+// test('queryByText nested <Image> in <Text> in middle', () => {
+//   expect(
+//     render(
+//       <Text>
+//         Hello
+//         <Image source={{}} />
+//         World
+//       </Text>
+//     ).queryByText('HelloWorld')
+//   ).toBeTruthy();
+// });
+
+// test('queryByText not found', () => {
+//   expect(
+//     render(
+//       <Text>
+//         Hello
+//         <Image source={{}} />
+//       </Text>
+//     ).queryByText('SomethingElse')
+//   ).toBeFalsy();
+// });
+
+test('queryByText nested multiple <Text> in <Text>', () => {
   expect(
     render(
       <Text>
-        <Image source={{}} />
-        Hello
+        Hello{' '}
+        <Text>
+          World<Text>!</Text>
+        </Text>
       </Text>
-    ).queryByText('Hello')
+    ).queryByText('Hello World!')
   ).toBeTruthy();
 });
 
-test('queryByText nested <Image> in <Text> at end', () => {
-  expect(
-    render(
-      <Text>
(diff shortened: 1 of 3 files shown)


thymikee:
Awesome! Can we add a test please? :)

src/helpers/getByAPI.js
```diff
@@ -36,17 +36,37 @@ const getNodeByText = (node, text) => {
...
     if (typeof child === 'number') {
-      return child.toString();
+      return textContent.push(child.toString());
     }
 
-    return '';
+    if (child.props.children) {
+      const { children } = child.props;
+
+      if (children instanceof Array) {
```
thymikee, line 53: Don't use `instanceof` to check for array. Use `.length`

src/helpers/getByAPI.js
```diff
@@ -36,17 +36,37 @@ const getNodeByText = (node, text) => {
...
+      return textContent.push(child);
     }
 
     if (typeof child === 'number') {
-      return child.toString();
+      return textContent.push(child.toString());
     }
 
-    return '';
+    if (child.props.children) {
```
thymikee, line 50: Here, `child` can be `NaN`, `null`, `undefined`, `Symbol`. None of these have `props` member. We should guard against it

thymikee: This will also check only one level of text, and you can nest more or components that wrap Text host component. If we want to get it right, I feel we should make `getChildrenAsText` recursive

src/helpers/getByAPI.js
```diff
@@ -36,17 +36,37 @@ const getNodeByText = (node, text) => {
...
+      const { children } = child.props;
+
+      if (children instanceof Array) {
+        children.forEach(node => {
+          // eslint-disable-next-line
+          const { Text } = require('react-native');
```
thymikee, line 56: `Text` is retrieved from the caller, let's pass it through function parameters, instead of requiring it in yet another place.

src/__tests__/queryByApi.test.js
```diff
@@ -47,3 +47,13 @@ test('queryByText not found', () => {
     ).queryByText('SomethingElse')
   ).toBeFalsy();
 });
+
+test('queryByText nested <Text> in <Text>', () => {
+  expect(
+    render(
+      <Text>
+        Hello <Text>World!</Text>
+      </Text>
```
thymikee, line 56: I'd also like these scenarios:
```jsx
<Text>
   Hello <Text>World<Text>!</Text></Text>
</Text>
```

```jsx
<Text>
   Hello <CustomText>World!</CustomText> // where CustomText renders Text 
</Text>
```

src/__tests__/queryByApi.test.js
```diff
@@ -3,56 +3,73 @@ import React from 'react';
...
-  expect(
-    render(
-      <Text>
-        <Image source={{}} />
-        Hello
-      </Text>
-    ).queryByText('Hello')
-  ).toBeTruthy();
-});
+// test('queryByText nested <Image> in <Text> at start', () => {
```
thymikee, line 6: FYI: if you want to skip all other tests than your current one, you can use `test.only` on that test to focus only there :)

marchenk0va: oops, thanks😅

thymikee:
Thanks once again! @Trancever will be happy :D

PR #226, opened 2019-10-16
docs: migrate Docusaurus to v2

Croftyland:
### Summary

Migration from v1 to v2
### Test plan

Hi, @yangshun,  please check [react native testing library ](https://github.com/callstack/react-native-testing-library/pull/226/files/2fa4038e59ef67f365bd25ecbe678d79d75897c6#diff-49a4483e3b8351e5a3689119bc2da52c) as an example of resolved issue "migrate react native testing library to v2 #226 "

PR #299, opened 2020-05-11
docs: migrate to Docusaurus v2

### LINKED ISSUE #225: Migration Website to Docusaurus 2 Alpha
## Describe the Feature

Hi there! The Docusaurus team is currently developing Docusaurus 2 and are almost done with an MVP that doesn't include translations and versioning features, which means there's feature parity with your website's current usage of Docusaurus.

Docusaurus 2 brings about many improvements:

- Client-side rendering with prerendering (site renders without JavaScript!)
- More flexible appearance - Customizable layouts, CSS modules, etc
- Embeddable interactive React components within markdown via MDX

I will be your POC in helping you throughout the migration process. Let me know what you think!

cc @yangshun @endiliey

cross19xx:
### Summary

Migration of the website from Docusaurus v1 to Docusaurus v2.0.0-alpha.54

Closes #225
Closes #226


thymikee:
Does it make #226 obsolete?

cross19xx:
In a way, yes.

This was not just changing the versions, but a few details were modified to fit the version

thymikee:
cc @yangshun anything obvious we missed during the migration? :)

website/src/components/Feature.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import styles from '../css/index.module.css';
```
yangshun, line 5: You can use `@site/src/css/index.module.css` instead of relative paths so that if you move this component around it wouldn't break :)

website/src/pages/index.js
```diff
@@ -0,0 +1,76 @@
...
+        title={feature.title}
+        description={feature.description}
+        imgUrl={feature.imgUrl}
+      />
+    ));
+  };
+
+  return (
+    <Layout title={`${siteConfig.title}`} description={siteConfig.tagline}>
+      <main className="home-container">
```
yangshun, line 51: This class seems unused

website/src/components/Feature.js
thymikee: It's only one level up, so I don't mind, but thanks for suggestion!

.gitignore
```diff
@@ -7,3 +7,4 @@ build
 
 # Ignore lock files in examples for now
 examples/**/yarn.lock
+.docusaurus
```
thymikee, line 10: We should be able to remove that entry, since it's present in `website/.gitignore`
```suggestion
```

Issue #270, opened 2020-03-18
getAllByTestId returns empty list while getByTestId returns an item


PMabMI:
## Versions

react: 16.9.0
react-native: 0.61.5
react-native-testing-library: 1.13.0
react-test-renderer: 16.9.0

## Description

When attempting to find multiple items with the same testID property (FlatList render item), I am unable to fetch any test components using getAllByTestId. 

getByTestId does return the first item with that test ID in the virtual DOM.

I am able to get all of the test items by using the test renderer directly and doing a findAllByProp query from there.

## Reproducible Demo

```
import { render, RenderAPI } from 'react-native-testing-library';
import { FlatList, Text } from 'react-native';
import { create } from 'react-test-renderer';

class ListItem extends React.Component<{ testID: string }> {
  render() {
    return <Text>foo</Text>;
  }
}

it('should have two items', () => {
  const renderer = render((
    <FlatList data={[{}, {}]} renderItem={() => <ListItem testID="listItem" />} />
  ));
  expect(renderer.getAllByTestId('listItem')).toHaveLength(2);
});

it('should have two items', () => {
  const renderer = create((
    <FlatList data={[{}, {}]} renderItem={() => <ListItem testID="listItem" />} />
  ));
  expect(renderer.root.findAllByProps({ testID: 'listItem' })).toHaveLength(2);
});
```

thymikee:
There's a bug in how `getByTestId` works. `getAllByTestId` is correct. Will leave that open so it stays clear. Fixing that means a breaking change.

thymikee:
In the meantime, do yourself a favor and, please, never use `ById` queries. You don't need them really.

PMabMI:
Ok, I looked at the source code and see that there's a filter where the type of "type" should be string. I'm new to React and React Native, so forgive me if this is a dumb question, but all of the "type"s that I have in my test are "object" and "function". What am I doing wrong here, as I am unable to get any of the test instances from the queries that have this filter?

thymikee:
The type of string for a react element means it's "host" element - platform specific, like div or View. The purpose of testID is that it propagates to the native view, so it's accessible by e2e test frameworks. We don't do that filtering in getById and that's wrong. 

Instead of grabbing components by id you should check something observable - a text, a value or accessibility props. Use these marchers. If you can't grab something using those, your users can't do it as well. Keep that in mind

PMabMI:
[...]

PR #304, opened 2020-05-13
feat: add `findBy*` queries

mdjastrzebski:
### Summary

Added findBy queries for original queries (ByText, ByPlaceholder, ByDisplayValue, ByTestId) as well as for a11y queries. 

*Note:* in case of `findByTestId` I used custom finding algorithm because current implementation of `getByTestId` is [broken](https://github.com/callstack/react-native-testing-library/issues/270).

### Test plan
Added units tests for all of the new queries.


docs/Queries.md
```diff
@@ -24,6 +24,17 @@ title: Queries
```
thymikee, line 24: can you rebase to master? `docs` were moved to `website/docs`

src/helpers/getByAPI.js
```diff
@@ -157,6 +157,23 @@ export const getByTestId = (instance: ReactTestInstance) =>
     }
   };
 
+export const fixedGetByTestId = (instance: ReactTestInstance) =>
+  function getByTestIdFn(testID: string) {
+    try {
+      const results = getAllByTestId(instance)(testID);
+      if (results.length === 1) return results[0];
+      else
```
thymikee, line 165: can we use some old fashioned curly braces with this if statement? :)

docs/Queries.md
mdjastrzebski: Done :-)

mdjastrzebski:
> Missing updates to `typings/index.d.ts` for TS users, but from a quick look it looks good!

I've added typescipt type definitions & tests for that.

typings/index.d.ts
```diff
@@ -57,61 +71,147 @@ export interface QueryByAPI {
...
-type GetReturn = ReactTestInstance;
-type GetAllReturn = Array<ReactTestInstance>;
-type QueryReturn = ReactTestInstance | null;
-type QueryAllReturn = Array<ReactTestInstance> | [];
+export interface WaitForOptions {
+  timeout: number;
+  interval: number;
+}
+
+export interface FindByAPI {
```
thymikee, line 91: I don't think we need to export this interface, what for?

mdjastrzebski: Same happens for GetByAPI & QueryByAPI interfaces. Do you things it's good idea to remove all of these exports?

thymikee: If we don't use them, then let's remove. If they're useful for somebody, they'll send a PR to export it :)

mdjastrzebski: I've removed exports for all three of them

mdjastrzebski:
Reversed prettier run on typings files in order to remove noise from the PR :-)

mdjastrzebski:
I've added overlapping parts with `within` operator which has been merged to master.

src/__tests__/a11yAPI.test.js
```diff
@@ -62,74 +64,127 @@ function Section() {
...
   );
   const button = queryByA11yLabel(/button/g);
   expect(button && button.props.accessibilityLabel).toEqual(BUTTON_LABEL);
+
+  const asyncButton = await findByA11yLabel(BUTTON_LABEL);
+  expect(asyncButton.props.accessibilityLabel).toEqual(BUTTON_LABEL);
+
   expect(() => getByA11yLabel(NO_MATCHES_TEXT)).toThrow(NO_INSTANCES_FOUND);
   expect(queryByA11yLabel(NO_MATCHES_TEXT)).toBeNull();
+  await expect(
```
thymikee, line 83: Can we move all async queries to the bottom of these tests? There's a possibility that a sync query would regress and doing async stuff in-between could hide that.

website/docs/Queries.md
```diff
@@ -24,6 +24,17 @@ title: Queries
...
+### findBy
+
+`findBy` queries return a promise which resolves when a matching element is found. The promise is rejected if no elements match or if more than one match is found after a default timeout of 4500ms. If you need to find more than one element, then use `findAllBy`.
+
+### findAllBy
+
+`findAllBy` queries return a promise which resolves to an array when any matching elements are found. The promise is rejected if no elements match after a default timeout of 4500ms.
+
+
+*Note:* `findBy` and `findAllBy` queries accept optional `waitForOptions` object argument which can contain `timeout` and `interval` properies which have the same meaning as respective arguments to [`waitForElement`](https://callstack.github.io/react-native-testing-library/docs/api#waitforelement) function.
```
thymikee, line 36:
```suggestion
:::info
`findBy` and `findAllBy` queries accept optional `waitForOptions` object argument which can contain `timeout` and `interval` properies which have the same meaning as respective arguments to [`waitForElement`](https://callstack.github.io/react-native-testing-library/docs/api#waitforelement) function. 
:::
```

src/__tests__/a11yAPI.test.js
mdjastrzebski: Done

PR #305, opened 2020-05-14
Added usage notes for ByTestId queries

### LINKED ISSUE #303: Update docs to mention buggy behavior of `getByTestId` and `queryByTestId`
## Describe the Feature

Add a warnings under `getByTestId` and `queryByTestId` queries: 

```
> Current implementation has a serious flaw, which results in finding more IDs than there really is in native host React components. Fixing it may break some of your tests so we'll do it in next major release (v2). As a mitigation, please use `getAllByTestId('your-id')[0]` or migrate off of testing with testID, which is considered unsafe.
```

And a second one, a bit unrelated by why not:

```
> Please be mindful when using this API and **treat it as an escape hatch**. Your users can't interact with `testID` anyhow, so you may end up writing tests that provide false sense of security. Favor text and accessibility queries instead. 
```

## Related Issues

See https://github.com/callstack/react-native-testing-library/issues/270

mdjastrzebski:
### Summary

Based on [this issue](https://github.com/callstack/react-native-testing-library/issues/303) I've updated the docs for ByTestId queries.

Closes #303

### Test plan
Not applicable.

### DIFF
--- a/website/docs/Queries.md
+++ b/website/docs/Queries.md
@@ -91,7 +91,9 @@ const { getByTestId } = render(<MyComponent />);
 const element = getByTestId('unique-id');
 ```
 
-*Note:* Please be mindful when using these API and **treat it as an escape hatch**. Your users can't interact with `testID` anyhow, so you may end up writing tests that provide false sense of security. Favor text and accessibility queries instead. 
+:::caution
+Please be mindful when using these API and **treat it as an escape hatch**. Your users can't interact with `testID` anyhow, so you may end up writing tests that provide false sense of security. Favor text and accessibility queries instead. 
+:::
 
 :::danger
 Current implementation of `getByTestId` and `queryByTestId` has a serious flaw, which results in finding more IDs than there really would be present in native React host components. Fixing it may break some of your tests so we'll do it in next major release (v2). As a temporary workaround, please use `getAllByTestId('your-id')[0]` or `queryAllByTestId('your-id')[0]` or migrate off testing with testID, which is considered to be an escape hatch.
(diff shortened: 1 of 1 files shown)


website/docs/Queries.md
```diff
@@ -91,6 +91,10 @@ const { getByTestId } = render(<MyComponent />);
 const element = getByTestId('unique-id');
 ```
 
+*Note:* Please be mindful when using these API and **treat it as an escape hatch**. Your users can't interact with `testID` anyhow, so you may end up writing tests that provide false sense of security. Favor text and accessibility queries instead. 
+
+*Warning:* Current implementation of `getByTestId` and `queryByTestId` has a serious flaw, which results in finding more IDs than there really would be present in native React host components. Fixing it may break some of your tests so we'll do it in next major release (v2). As a temporary workaround, please use `getAllByTestId('your-id')[0]` or `queryAllByTestId('your-id')[0]` or migrate off testing with testID, which is considered to be an escape hatch.
```
thymikee, line 96:
```suggestion
:::danger
Current implementation of `getByTestId` and `queryByTestId` has a serious flaw, which results in finding more IDs than there really would be present in native React host components. Fixing it may break some of your tests so we'll do it in next major release (v2). As a temporary workaround, please use `getAllByTestId('your-id')[0]` or `queryAllByTestId('your-id')[0]` or migrate off testing with testID, which is considered to be an escape hatch.
:::
```

website/docs/Queries.md
```diff
@@ -91,6 +91,10 @@ const { getByTestId } = render(<MyComponent />);
 const element = getByTestId('unique-id');
 ```
 
+*Note:* Please be mindful when using these API and **treat it as an escape hatch**. Your users can't interact with `testID` anyhow, so you may end up writing tests that provide false sense of security. Favor text and accessibility queries instead.
```
thymikee, line 94:
```suggestion
:::caution
Please be mindful when using these API and **treat it as an escape hatch**. Your users can't interact with `testID` anyhow, so you may end up writing tests that provide false sense of security. Favor text and accessibility queries instead. 
:::
```

mdjastrzebski:
Docosaurus preview:

image

Issue #297, opened 2020-05-11
Ability to query children of given element


mdjastrzebski:
## Describe the Feature
In certain scenarios it is useful to restrict query to only a part of component tree.
Sample use cases:
1. when displaying a `FlatList` of items we might want to find a certain item and query for some views inside it
2. when rendering multi-screen integration test using react-navigation and navigating from list screen to details screen. The global query still returns elements from hidden master screen. Forcing test code to use getAllBy* with arbitrary indices instread of getBy* queries.

## Possible Implementations
DOM/React Testing Library provides [`within`/`getQueriesForElement`](https://testing-library.com/docs/dom-testing-library/api-helpers#within-and-getqueriesforelement-apis) for such use case. 

```typescript
const boundQueries = within(getByA11yHint('Some Item'));
expect(boundQueries.getByText('Some text')).toBeTruthy();
```

Alternative would be to provide query functions accepting container element as one of their parameters e.g. `getByText(container, text)` as in DOM Testing Library.

## Related Issues
Not found.

thymikee:
I think I like the `within` helper more, let's implement it 👍

PR #306, opened 2020-05-14
`within` operator for bound queries

mdjastrzebski:
### Summary

This commit implements #297. New `within` operator, inspired by RTL operator of the same name exposes queries bound to given element. 

Available queries are the same as those from `render` (so getBy, queryBy, incl. A11y). Additional `render` specific operations like `update`, `unmount`, `debug`, `toJSON` are NOT included.

### Test plan
Added unit test for within operator as well as typescript type definitions tests.

### DIFF
--- a/src/__tests__/within.test.js
+++ b/src/__tests__/within.test.js
@@ -1,8 +1,7 @@
 // @flow
 import React from 'react';
 import { View, Text, TextInput } from 'react-native';
-import { render } from '..';
-import { within } from '../within';
+import { render, within } from '..';
 
 test('within() exposes basic queries', () => {
   const rootQueries = render(
--- a/src/__tests__/within.test.js
+++ b/src/__tests__/within.test.js
@@ -6,11 +6,11 @@ import { render, within } from '..';
 test('within() exposes basic queries', () => {
   const rootQueries = render(
     <View>
-      <View testID="first">
+      <View accessibilityHint="first">
         <Text>Same Text</Text>
         <TextInput value="Same Value" placeholder="Same Placeholder" />
       </View>
-      <View testID="second">
+      <View accessibilityHint="second">
         <Text>Same Text</Text>
         <TextInput value="Same Value" placeholder="Same Placeholder" />
       </View>
@@ -21,7 +21,7 @@ test('within() exposes basic queries', () => {
   expect(rootQueries.getAllByDisplayValue('Same Value')).toHaveLength(2);
   expect(rootQueries.getAllByPlaceholder('Same Placeholder')).toHaveLength(2);
 
-  const firstQueries = within(rootQueries.getAllByTestId('first')[0]);
+  const firstQueries = within(rootQueries.getByA11yHint('first'));
   expect(firstQueries.getAllByText('Same Text')).toHaveLength(1);
   expect(firstQueries.getByText('Same Text')).toBeTruthy();
   expect(firstQueries.queryAllByText('Same Text')).toHaveLength(1);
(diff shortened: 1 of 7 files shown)


typings/index.d.ts
```diff
@@ -1,5 +1,9 @@
 import * as React from 'react';
-import { AccessibilityState, AccessibilityStates, AccessibilityRole } from 'react-native';
+import {
```
thymikee, line 2: can we extract the Prettier run over `typings/` in a separate PR so that this one and #304 are shorter? :)

src/__tests__/within.test.js
```diff
@@ -0,0 +1,72 @@
+// @flow
+import React from 'react';
+import { View, Text, TextInput } from 'react-native';
+import { render } from '..';
+import { within } from '../within';
```
thymikee, line 5: we import from index.js for a reason – `within` is not exposed to the user currently, please import it there and then, in this test, from there :)

src/__tests__/within.test.js
```diff
@@ -0,0 +1,72 @@
...
+        <TextInput value="Same Value" placeholder="Same Placeholder" />
+      </View>
+    </View>
+  );
+
+  expect(rootQueries.getAllByText('Same Text')).toHaveLength(2);
+  expect(rootQueries.getAllByDisplayValue('Same Value')).toHaveLength(2);
+  expect(rootQueries.getAllByPlaceholder('Same Placeholder')).toHaveLength(2);
+
+  const firstQueries = within(rootQueries.getAllByTestId('first')[0]);
```
thymikee, line 25: Can we avoid using ByTestId in our tests?

thymikee:
Missing the docs :)

typings/index.d.ts
mdjastrzebski: Done

src/__tests__/within.test.js
mdjastrzebski: I've added proper index.js level export and import here.

src/__tests__/within.test.js
mdjastrzebski: Replaced byTestId with byA11yHint

mdjastrzebski:
Added docs for `within` operator

src/__tests__/within.test.js
```diff
@@ -0,0 +1,71 @@
...
+  const secondQueries = within(rootQueries.getByA11yHint('second'));
+  expect(secondQueries.getAllByText('Same Text')).toHaveLength(1);
+  expect(secondQueries.getByText('Same Text')).toBeTruthy();
+  expect(secondQueries.queryAllByText('Same Text')).toHaveLength(1);
+  expect(secondQueries.queryByText('Same Text')).toBeTruthy();
+  expect(secondQueries.getByDisplayValue('Same Value')).toBeTruthy();
+  expect(secondQueries.getByPlaceholder('Same Placeholder')).toBeTruthy();
+});
+
+test('within() exposes a11y queries', () => {
```
thymikee, line 41: I think we could merge these 2 tests, but no biggie

website/docs/API.md
```diff
@@ -341,6 +341,30 @@ test('waiting for an Banana to be ready', async () => {
...
+
+- [`Example code`](https://github.com/callstack/react-native-testing-library/blob/master/src/__tests__/within.test.js)
+
+Defined as:
+
+```jsx
+function within(instance: ReactTestInstance): Queries
+```
+
+Perform [queries](./Queries.md) scoped to given element.
```
thymikee, line 354: I think it's worth to add this note, WDYT?
```suggestion
Perform [queries](./Queries.md) scoped to given element. 

:::info
Please note that additional `render` specific operations like `update`, `unmount`, `debug`, `toJSON` are _not_ included.
:::
```

PR #308, opened 2020-05-14
chore: bump eslint config to latest

thymikee:
### Summary

New Prettier and stuff

### Test plan

CI

### DIFF
--- a/src/__tests__/a11yAPI.test.js
+++ b/src/__tests__/a11yAPI.test.js
@@ -27,13 +27,13 @@ class Button extends React.Component<any> {
       <TouchableOpacity
         accessibilityHint={BUTTON_HINT}
         accessibilityLabel={BUTTON_LABEL}
-        accessibilityRole={BUTTON_ROLE}
+        accessibilityRole={BUTTON_ROLE} // eslint-disable-line react-native-a11y/has-valid-accessibility-role
         accessibilityStates={['selected']}
       >
         <Typography
           accessibilityHint={TEXT_HINT}
           accessibilityLabel={TEXT_LABEL}
-          accessibilityRole={TEXT_ROLE}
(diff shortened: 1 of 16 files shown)


src/__tests__/a11yAPI.test.js
```diff
@@ -27,13 +27,13 @@ class Button extends React.Component<any> {
...
         accessibilityLabel={BUTTON_LABEL}
-        accessibilityRole={BUTTON_ROLE}
+        accessibilityRole={BUTTON_ROLE} // eslint-disable-line react-native-a11y/has-valid-accessibility-role
         accessibilityStates={['selected']}
       >
         <Typography
           accessibilityHint={TEXT_HINT}
           accessibilityLabel={TEXT_LABEL}
-          accessibilityRole={TEXT_ROLE}
+          accessibilityRole={TEXT_ROLE} // eslint-disable-line react-native-a11y/has-valid-accessibility-role
```
mdjastrzebski, line 36: TEXT_ROLE is set to link, so maybe it would be better to use it here directly and do not suppress eslint for a11y?

src/__tests__/a11yAPI.test.js
```diff
@@ -27,13 +27,13 @@ class Button extends React.Component<any> {
       <TouchableOpacity
         accessibilityHint={BUTTON_HINT}
         accessibilityLabel={BUTTON_LABEL}
-        accessibilityRole={BUTTON_ROLE}
+        accessibilityRole={BUTTON_ROLE} // eslint-disable-line react-native-a11y/has-valid-accessibility-role
```
mdjastrzebski, line 30: BUTTON_ROLE is set to 'button', so maybe it would be better to use it here directly and do not suppress eslint for a11y?

src/__tests__/a11yAPI.test.js
thymikee: Good point, it was only used in like 1 or 2 places, unnecessary

PR #324, opened 2020-05-18
fix(breaking): `getByTestId` should work only for native elements

mdjastrzebski:
### Summary

Existing version of `getByTestId` had a bug (#270) as it returned also non-native elements. This was in contract with `getAllByTestId` which returned only native elements.

### Test plan
Added tests checking `getByTestId` behavior.

### DIFF
--- a/src/__tests__/getByApi.test.js
+++ b/src/__tests__/getByApi.test.js
@@ -1,0 +1,40 @@
+// @flow
+import React from 'react';
+import { View, Text, TextInput, Button } from 'react-native';
+import { render } from '..';
+
+const options = { timeout: 10 }; // Short timeout so that this test runs quickly
+
+const MyComponent = () => {
+  return <Text>My Component</Text>;
+};
+
+test('byTestId returns only native elements', () => {
+  const {
+    getByTestId,
+    getAllByTestId,
+    queryByTestId,
+    queryAllByTestId,
+  } = render(
+    <View>
+      <Text testID="text">Text</Text>
+      <TextInput testID="textInput" />
+      <View testID="view" />
+      <Button testID="button" title="Button" />
+      <MyComponent testID="myComponent" />
+    </View>
+  );
+
+  expect(getByTestId('text')).toBeTruthy();
+  expect(getByTestId('textInput')).toBeTruthy();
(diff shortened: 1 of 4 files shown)


mdjastrzebski:
Fixes #270

src/__tests__/getByApi.test.js
```diff
@@ -0,0 +1,38 @@
...
+  expect(getByTestId('textInput')).toBeTruthy();
+  expect(getByTestId('view')).toBeTruthy();
+  expect(getByTestId('button')).toBeTruthy();
+
+  expect(getAllByTestId('text')).toHaveLength(1);
+  expect(getAllByTestId('textInput')).toHaveLength(1);
+  expect(getAllByTestId('view')).toHaveLength(1);
+  expect(getAllByTestId('button')).toHaveLength(1);
+
+  expect(queryByTestId('myComponent')).toBeFalsy();
```
thymikee, line 36: can we add `getByTestId` that fails for completeness?

mdjastrzebski: Sure, I've replaced queryBy with getBy for negative cases.

src/__tests__/getByApi.test.js
```diff
@@ -33,6 +28,6 @@ test('byTestId returns only native elements', () => {
   expect(getAllByTestId('view')).toHaveLength(1);
   expect(getAllByTestId('button')).toHaveLength(1);
 
-  expect(queryByTestId('myComponent')).toBeFalsy();
-  expect(queryAllByTestId('myComponent')).toHaveLength(0);
+  expect(() => getByTestId('myComponent')).toThrowError();
+  expect(() => getAllByTestId('myComponent')).toThrowError();
```
thymikee, line 32: Let's add what text we expect to see there like `toThrowError('no instances found')`. The `toThrowError()` matcher without an argument may hide unexpected errors and should generally be avoided

mdjastrzebski: Good idea, fixed! :-)

conor909:
Is there some documentation that needs to be updated regarding this?  My `getByTestId` code works for native but not for web

PR #328, opened 2020-05-18
Feat(breaking): remove legacy degug export

### LINKED ISSUE #317: Remove `debug` export from the library
## Describe the Feature

There are 2 ways to `debug` in RNTL at the moment. 
The legacy way:
```js
import { debug } from 'react-native-testing-library';
debug(<Component />); // renders shallow by default
debug.deep(<Component />); 
```
The correct way:
```js
import { render } from 'react-native-testing-library';
const {debug} = render(<Component />); 
debug(); // renders deep by default
debug.shallow();
```

We'd like to get rid of "the legacy way". It was introduced first and proven to be worse API to the later-introduced "the correct way", which is easier to use – less typing and scoped to the component we're currently testing. It's confusing we have 2 ways to do the same thing, which also have swapped defaults, which is confusing.

mdjastrzebski:
### Summary

Remove global `debug` export and implementation, in favor of `debug` method returned by `render()` functions. 

Resolves #317 

### Test plan
Removed related tests.


thymikee:
Doesn't this leave us with no tests around `debug` returned from `render`?

mdjastrzebski:
There are some `render()`s tests for `debug` which are roughly equivalent to removed global `debug` tests.

PR #333, opened 2020-05-19
docs: Native module example and documentation

### LINKED ISSUE #282: Include an example for mocking native modules
## Describe the Feature
As part of the examples being created, please include an example of mocking native modules.

cross19xx:
### Summary
This PR describes how to test native modules with RNTL. Resolves #282


examples/nativemodules/Storage.test.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { fireEvent, render, act } from 'react-native-testing-library';
+import StorageView from './StorageView';
+
+/* eslint-disable jest/no-test-callback */
+
+// Mock the react-native-testing-library
+// For this test case, we are interested in mocking just 'getItem' and 'setItem'
+jest.mock('@react-native-community/async-storage', () => {
```
mdjastrzebski, line 9: It seems that async-storage offeres [their own mock](https://github.com/react-native-community/async-storage/blob/master/docs/Jest-integration.md#using-async-storage-mock). Not sure if it's relevant here though.

mdjastrzebski:
@cross19xx not sure if I understand PR goal correctly. In my opinion it would be useful to show how to mock some true native modules as well as custom native UI components. Which probably would include mocking `NativeModules` and/or `requireNativeComponent` from `react-native`. What do you think?

examples/nativemodules/Storage.test.js
cross19xx: Actually, it's not. But I wanted to pick an example that shows mocking. I plan on including a link to testing in Async storage inside the docs

cross19xx:
> @cross19xx not sure if I understand PR goal correctly. In my opinion it would be useful to show how to mock some true native modules as well as custom native UI components. Which probably would include mocking `NativeModules` and/or `requireNativeComponent` from `react-native`. What do you think?

Actually, that's a good idea. This PR will also include mocking custom UI native components.

examples/nativemodules/PureNativeModule.test.js
```diff
@@ -0,0 +1,26 @@
...
+  RN.NativeModules.NativeNotification = {
+    show: jest.fn(),
+  };
+
+  return RN;
+});
+
+// Error thrown by import-order. Ignored due to the mock
+// eslint-disable-next-line
+import RN from 'react-native';
```
thymikee, line 16: `jest.mock` is hosited above imports (which are transformed to `require`s). Also, this import is not needed at all, please remove

examples/nativemodules/Storage.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    const populateBtn = getByText(/Populate/i);
+    await act(async () => {
+      fireEvent.press(populateBtn);
+    });
+
+    const text = getByText(/Counter/);
+    expect(text.props.children.join('')).toEqual('Counter 1');
+
+    done();
```
thymikee, line 43: Oh nope nope nope. Never ever use `done`. You have `async` already, all you need to do is await async stuff and that's it. Please remove and never use `done` again. There's a lint rule for that, maybe it's not enabled by default

examples/nativemodules/Storage.test.js
```diff
@@ -0,0 +1,45 @@
...
+describe('Native module test', () => {
+  test('increments the counter the Increment is pressed', async (done) => {
+    const { getByText } = render(<StorageView />);
+    const button = getByText(/Increment/i);
+    fireEvent.press(button);
+
+    const populateBtn = getByText(/Populate/i);
+    await act(async () => {
+      fireEvent.press(populateBtn);
+    });
```
thymikee, line 38: Please remove `act`. `fireEvent` is already wrapped in act. If you get errors, it's because the test finishes before async actions. You should then use `await findByText()` instead of `getBy`

examples/nativemodules/Storage.test.js
thymikee: @mdjastrzebski is right, we shouldn't promote writing custom mocks, if there's an official one provided. Please remove the Storage example entirely

examples/nativemodules/StorageView.js
```diff
@@ -0,0 +1,40 @@
+import React, { useState } from 'react';
+import { Text, View, Button } from 'react-native';
+import AsyncStorage from '@react-native-community/async-storage';
+
+/* eslint-disable no-console */
+const KEY = '@storage_key';
+
+export default function StorageView() {
+  const [count, setCount] = useState('0');
```
thymikee, line 9: counters are usually numbers, so we should use numbers

examples/nativemodules/PureNativeModule.test.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { fireEvent, render } from 'react-native-testing-library';
+import PureNativeModule from './PureNativeModule';
+
+jest.mock('react-native', () => {
+  const RN = jest.requireActual('react-native');
+  RN.NativeModules.NativeNotification = {
+    show: jest.fn(),
+  };
```
thymikee, line 9: Some native components, like `react-native-safe-area-context`, register through `UIManager.getViewManagerConfig`. We need to add that as well. Check their tests:
https://github.com/th3rdwave/react-native-safe-area-context/blob/24018ffa38fec85fb5dac1f34a93d23319f7df7b/src/__tests__/initialWindowMetrics-test.tsx

examples/nativemodules/PureNativeModule.test.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { fireEvent, render } from 'react-native-testing-library';
+import PureNativeModule from './PureNativeModule';
+
+jest.mock('react-native', () => {
```
thymikee, line 5: Also, I think we should be able to mock like this:
```js
import { NativeModules, UIManager } from 'react-native';

jest.spyOn(NativeModules, 'NativeNotification').mockImplementation(() => ({show: jest.fn()}));
jest.spyOn(UIManager, 'RNCSafeAreaView').mockImplementation(() => ({}));
```

I think that would be cleaner than using `jest-mock`. The downside is that `jest.spyOn` is not hoisted, so if a module has side-effects and calls NativeModules/UIManager at import time, it probably won't work

examples/nativemodules/PureButton.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+describe('Testing Pure button', () => {
+  test('expect RN to be truthy', async () => {
+    const { getByAccessibilityHint } = render(<App />);
+    const sample = getByAccessibilityHint('sample');
+
+    fireEvent.press(sample);
+
+    expect(
+      UIManager.getViewManagerConfig('PureButton').onPress
```
thymikee, line 26: Can we instead import PureButton and assert on that? That's usually what users want to test

PR #337, opened 2020-05-20
maintenance: cleanup TS typedefs tests

mdjastrzebski:
### Summary
I have compacted our TS typedefs tests, so that queries with same return type are grouped together as an array, which makes the code more compact and easier to spot missing/wrong tests.

### DIFF
--- a/typings/__tests__/index.test.tsx
+++ b/typings/__tests__/index.test.tsx
@@ -4,7 +4,6 @@ import { ReactTestInstance } from 'react-test-renderer';
 import {
   render,
   fireEvent,
-  shallow,
   flushMicrotasksQueue,
   waitFor,
   act,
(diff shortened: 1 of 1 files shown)


typings/__tests__/index.test.tsx
```diff
@@ -161,105 +165,50 @@ const findAllBy: Promise<ReactTestInstance[]>[] = [
...
+  }),
+  waitFor<ReactTestInstance>(() => tree.getByA11yLabel('label'), {
+    timeout: 100,
+    interval: 10,
+  }),
+];
 
-const waitForFlush: Promise<any> = flushMicrotasksQueue();
+const waitGetAllBy: Promise<ReactTestInstance[]>[] = [
+  waitFor<ReactTestInstance[]>(() => tree.getAllByA11yLabel('label')),
```
thymikee, line 199: `waitFor` can work with any throwable function, maybe we can showcase that and get rid of `tree.getAllBy` examples?

PR #340, opened 2020-05-20
chore(breaking): deprecate flushMicrotasksQueue

### LINKED ISSUE #339: Deprecate `flushMicrotaskQueue`
## Summary

Writing async tests using `flushMicrotasksQueue` is quite not intuitive, as there are better alternatives available like `waitFor` or `findBy` queries. Similar function while contained in RTL is not exported from it.

Let's deprecate it in 2.x release and the remove it in 3.x. Users still relying on it, will be able to replace it with copy-paste implementation.

mdjastrzebski:
### Summary

Making `flushMicrotasksQueue` deprecated in favor of `findBy` and `waitFor`.

Closes #339

### DIFF
--- a/src/flushMicrotasksQueue.js
+++ b/src/flushMicrotasksQueue.js
@@ -1,7 +1,15 @@
 // @flow
+
+import { printDeprecationWarning } from './helpers/errors';
+
 /**
  * Wait for microtasks queue to flush
  */
 export default function flushMicrotasksQueue(): Promise<any> {
+  printDeprecationWarning('flushMicrotasksQueue');
+  return flushMicrotasksQueueInternal();
+}
+
+export function flushMicrotasksQueueInternal(): Promise<any> {
   return new Promise((resolve) => setImmediate(resolve));
 }
--- a/website/docs/Migration20.md
+++ b/website/docs/Migration20.md
@@ -116,7 +116,7 @@ In general, you should avoid `byTestId` queries when possible and rather use que
 
 ## Deprecated `flushMicrotasksQueue`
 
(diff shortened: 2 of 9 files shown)


src/flushMicrotasksQueue.js
```diff
@@ -1,7 +1,15 @@
...
+
 /**
  * Wait for microtasks queue to flush
  */
 export default function flushMicrotasksQueue(): Promise<any> {
+  printDeprecationWarning('flushMicrotasksQueue');
+  return flushMicrotasksQueueInternal();
+}
+
+export function flushMicrotasksQueueInternal(): Promise<any> {
```
thymikee, line 13: Can we call it just `flushMicrotasks`?

website/docs/Migration20.md
```diff
@@ -113,3 +113,15 @@ If you relied on setting `testID` for your custom components, you should probabl
 :::caution
 In general, you should avoid `byTestId` queries when possible and rather use queries that check things that can been seen by the user (e.g. `byText`, `byPlaceholder`, etc) or accessability queries (e.g. `byA11yHint`, `byA11yLabel`, etc).
 :::
+
+## Deprecated `flushMicrotasksQueue`
+
+We have deprecated `flushMicrotasksQueue` and plan to remove it in the next major version, as currently there are better available alternatives for helping you write async tests: `findBy` async queries and `waitFor` helper.
```
thymikee, line 119:
```suggestion
We have deprecated `flushMicrotasksQueue` and plan to remove it in the next major version, as currently there are better alternatives available for helping you write async tests: `findBy` async queries and `waitFor` helper.
```

src/flushMicrotasksQueue.js
mdjastrzebski: Renamed to `flushMicroTasks` as this is how it's spelled in RTL

PR #344, opened 2020-05-21
feat: wrap `waitFor` in (async) act to support async queries without explicit `act` call

### LINKED ISSUE #343: Wrap `waitFor` with async `act`
## Describe the Feature

React batches updates to the underlying native platform using `act` helper. We already use that for sync calls to `render`, `update` and `fireEvent`. However, it's common to put side effects updating React component state inside `useEffect`. Doing that will produce the dreaded "updates should be wrapped into act(...)" warning. To account for that, React introduced async version of `act` in v16.9.

So, to avoid our users wrap their `waitFor` and `findBy` queries with `async act()`, we can wrap `waitFor` callback with an async act, so it's done automatically, removing churn when testing React.

## Possible Implementations

If we bump the minimum required React version to 16.9, it should be as easy as wrapping wait for with async `act`. However, if we want to provide compatibility with older React versions, we'll need to come up with a workaround, because IIRC `act` in React 16.8 yells if you try to use it in non-sync way

mdjastrzebski:
## Summary

Wrap `waitFor` in (async) act to support async queries without explicit `act` call. Otherwise there was a problem when async state changes caused by e.g. async `useEffect` uses (fetching data, ...) generated dreded `act` warnings in the console during test runs.

Closes #343

### DIFF
--- a/src/waitFor.js
+++ b/src/waitFor.js
@@ -50,7 +50,7 @@ export default async function waitFor<T>(
     result = await waitForInternal(expectation, options);
   });
 
-  //$FlowFixMe: either we have result or `waitFor` there error
+  //$FlowFixMe: either we have result or `waitFor` threw error
   return result;
 }
 
--- a/src/waitFor.js
+++ b/src/waitFor.js
@@ -45,7 +45,7 @@ export default async function waitFor<T>(
 ): Promise<T> {
   let result: T;
 
-  //$FlowFixMe: this is just too complicated for flow
+  //$FlowFixMe: `act` has incorrect typing
   await act(async () => {
     result = await waitForInternal(expectation, options);
   });
--- a/src/waitFor.js
+++ b/src/waitFor.js
@@ -9,9 +9,7 @@ const DEFAULT_INTERVAL = 50;
 
 function checkReactVersionAtLeast(major: number, minor: number): boolean {
   if (React.version === undefined) return false;
-  const [actualMajor, actualMinor] = React.version
-    .split('.')
-    .map(Number);
+  const [actualMajor, actualMinor] = React.version.split('.').map(Number);
 
   return actualMajor > major || (actualMajor === major && actualMinor >= minor);
 }
--- a/src/waitFor.js
+++ b/src/waitFor.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React from 'react';
+import * as React from 'react';
 import act from './act';
 import { throwRemovedFunctionError } from './helpers/errors';
 
--- a/src/waitFor.js
+++ b/src/waitFor.js
@@ -1,5 +1,6 @@
 // @flow
 
+import act from './act';
 import { throwRemovedFunctionError } from './helpers/errors';
 
 const DEFAULT_TIMEOUT = 4500;
@@ -10,7 +11,7 @@ export type WaitForOptions = {
   interval?: number,
 };
 
-export default function waitFor<T>(
+function waitForInternal<T>(
   expectation: () => T,
   options?: WaitForOptions
 ): Promise<T> {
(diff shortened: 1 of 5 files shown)


src/waitFor.js
```diff
@@ -38,6 +39,21 @@ export default function waitFor<T>(
...
+): Promise<T> {
+  let result: T;
+
+  //$FlowFixMe: this is just too complicated for flow
+  await act(async () => {
+    result = await waitForInternal(expectation, options);
+  });
+
+  //$FlowFixMe: either we have result or `waitFor` there error
+  return result;
```
thymikee, line 54: How about:
```suggestion
  return act(() => waitForInternal(expectation, options));
```

mdjastrzebski: Unfortunately does not work, as act is not forwarding the return value from the callback.

mdjastrzebski: We need to capture it to intermediate variable so that it can be returned by the `waitFor` async wrapper

website/docs/API.md
```diff
@@ -336,6 +336,10 @@ function waitFor<T>(
 
 Waits for non-deterministic periods of time until your element appears or times out. `waitFor` periodically calls `expectation` every `interval` milliseconds to determine whether the element appeared or not.
 
+:::info
+In order to properly use `waitFor` you need at least React >=16.9.0 (featuing async `act`) or React Native >=0.60 (which comes with React >=16.9.0).
```
thymikee, line 340: I'm not sure if "In order to properly use" is good-enough wording. Maybe: "To avoid `act` warnings"?
```suggestion
To avoid `act` warnings when using `waitFor` you need React >=16.9.0 (featuring async `act`) or React Native >=0.60 (which comes with React >=16.9.0).
```
Not sure which is better, feel free to pick.

mdjastrzebski: It's not about warning only, async tests will actually not work properly if you don't have R >= 16.9.

src/waitFor.js
thymikee: Ok, I see it now. The error suppressed by "//$FlowFixMe: this is just too complicated for flow" comes from lacking lib defs of react-test-renderer, not Flow being dumb. Worth updating the comment

src/waitFor.js
```diff
@@ -38,6 +49,25 @@ export default function waitFor<T>(
   });
 }
 
+export default async function waitFor<T>(
+  expectation: () => T,
+  options?: WaitForOptions
+): Promise<T> {
+  if (!checkReactVersionAtLeast(16, 9)) {
+    return await waitForInternal(expectation, options);
```
thymikee, line 57: `return await` is redundant, you can omit the `await`

src/waitFor.js
```diff
@@ -1,16 +1,27 @@
...
 import { throwRemovedFunctionError } from './helpers/errors';
 
 const DEFAULT_TIMEOUT = 4500;
 const DEFAULT_INTERVAL = 50;
 
+function checkReactVersionAtLeast(major: number, minor: number): boolean {
+  if (React.version === undefined) return false;
+  const [actualMajor, actualMinor] = React.version
+    .split('.')
+    .map((n) => parseInt(n, 10));
```
thymikee, line 14:
```suggestion
    .map(Number);
```

src/waitFor.js
```diff
@@ -1,16 +1,27 @@
...
+
+  return actualMajor > major || (actualMajor == major && actualMinor >= minor);
```
thymikee, line 16: We don't do `==` here, it's dangerous in JS :D 
```suggestion
  return actualMajor > major || (actualMajor === major && actualMinor >= minor);
```

src/waitFor.js
mdjastrzebski: Updated comment

src/waitFor.js
mdjastrzebski: Removed await

src/waitFor.js
```diff
@@ -1,16 +1,25 @@
 // @flow
 
+import React from 'react';
```
thymikee, line 3: Can we do this?
```suggestion
import * as React from 'react';
```
It's a recommended way to import stuff from React recently

PR #364, opened 2020-05-30
docs: update React Navigation tests to use find queries

wwdrew:
### Summary

This change updates the documentation around React Navigation integration to use the find* queries instead of get* queries.

It solves the tests displaying errors regarding ForwardRef(NavigationContainer).

### Test plan

Only a documentation change.


thymikee:
This is great! Could you also update the example as well so it stays in sync? https://github.com/callstack/react-native-testing-library/blob/master/examples/reactnavigation/src/__tests__/AppNavigator.test.js

wwdrew:
Absolutely, apologies I didn't actually realise that was there :)

PR #376, opened 2020-06-06
feature: wait for element to be removed

### LINKED ISSUE #375: Add `waitForElementToBeRemoved` API
## Describe the Feature
RTL contains a `waitForElementToBeRemoved` method for detecting if some element has been removed asynchronously (e.g. with animation). This type of query is not very straigthforward using existing `waitFor` method. 

## Possible Implementations
Implement `waitForElementToBeRemoved` based on RTL's implementation. Note it seems that you can use `getBy`, `getAllBy` as well as `queryBy` and `queryAllBy` with it.

## Related Issues
N/A

mdjastrzebski:
### Summary

`waitForElementToBeRemoved` implementation with tests & typescript types.

- [x] implementation
- [x] tests
- [x] typescript types
- [x] docs

Resolves #375 

### Test plan

Added tests based on `waitFor` & veryfing `getBy`, `getAllBy`, `queryBy` & `queryAllBy` queries.

### DIFF
--- a/src/waitForElementToBeRemoved.js
+++ b/src/waitForElementToBeRemoved.js
@@ -12,6 +12,14 @@ export default async function waitForElementToBeRemoved<T>(
   // Created here so we get a nice stacktrace
   const timeoutError = new Error('Timed out in waitForElementToBeRemoved.');
 
+  // Elements have to be present initally and then removed.
+  const initialElements = expectation();
+  if (isRemoved(initialElements)) {
+    throw new Error(
+      'The element(s) given to waitForElementToBeRemoved are already removed. waitForElementToBeRemoved requires that the element(s) exist(s) before waiting for removal.'
+    );
+  }
+
   return waitFor(() => {
     let result;
     try {
--- a/src/waitForElementToBeRemoved.js
+++ b/src/waitForElementToBeRemoved.js
@@ -1,0 +1,29 @@
+// @flow
+import waitFor from './waitFor';
+import type { WaitForOptions } from './waitFor';
+
+const isRemoved = (result) =>
+  !result || (Array.isArray(result) && !result.length);
+
+export default async function waitForElementToBeRemoved<T>(
+  expectation: () => T,
+  options?: WaitForOptions
+): Promise<boolean> {
+  // Created here so we get a nice stacktrace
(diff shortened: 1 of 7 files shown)


src/waitForElementToBeRemoved.js
```diff
@@ -0,0 +1,29 @@
...
+import waitFor from './waitFor';
+import type { WaitForOptions } from './waitFor';
+
+const isRemoved = (result) =>
+  !result || (Array.isArray(result) && !result.length);
+
+export default async function waitForElementToBeRemoved<T>(
+  expectation: () => T,
+  options?: WaitForOptions
+): Promise<null> {
```
mdjastrzebski, line 11: Not sure if `null` is a proper type here. But we do not have anything meaningful to return.
RTL declares its version in TS as `Promise<T>` but uses `return true` in their impl.

mdjastrzebski: IMO I would keep the `Promise<T>` return type and return the initial result of `expectation` call. Since in code we require that initially `expectation` is successful (returns something) and only then starts throwing/returning nothing.

thymikee: yup, let's do it

mdjastrzebski: Done

mdjastrzebski: Also logged issue in Dom TL that they have similar discrepancy.

thymikee:
Used `ErrorWithStack` to get a nicer stack trace:

|before|after|
|--|--|
|Screen Shot 2020-06-09 at 09 30 14|Screen Shot 2020-06-09 at 09 30 08|

website/docs/API.md
```diff
@@ -336,20 +336,55 @@ function waitFor<T>(
...
+
+  await waitFor(() => getByText('Banana ready'));
+});
+```
+
 :::info
 In order to properly use `waitFor` you need at least React >=16.9.0 (featuring async `act`) or React Native >=0.60 (which comes with React >=16.9.0).
 :::
 
+If you're using Jest's [Timer Mocks](https://jestjs.io/docs/en/timer-mocks#docsNav), remember not to use `async/await` syntax as it will stall your tests.
```
thymikee, line 353: for later: we need to figure out why this happens.

PR #407, opened 2020-06-27
Add support for react-native-web in getAllByTestId

abmantis:
### Summary

Because react-native-web uses "data-testid" instead of "testID", getAllByTestId was failing to get elements. This PR fixes it by looking for "data-testid" too.

### Test plan

Run tests with react-native-web. I'm using jest-expo/web preset for that.


abmantis:
@thymikee Hey, can you please take a look?

thymikee:
I'd like @Esemesek and @mdjastrzebski and @satya164 to have a look first. This widens the implementation area we use and makes assumptions about the environments we run in, so I'd like give it a proper thought. 

Personally, I'd use `react-testing-library` to test logic specific to web, because it works in an emulated JSDOM environment and is likely to produce better results and cover more APIs without mocking yourself.

mdjastrzebski:
@abmantis I am curious how did you configure your test environment? If I understand correctly, by default RNTL renders output using React Test Rendered, which is platform-agnostic, and renders things like `<View>`s, `<Text>`s, etc, but not `<div>`s, and other HTML-specific elements.

abmantis:
@mdjastrzebski I'm using Expo, with the [jest-expo](https://www.npmjs.com/package/jest-expo), with the `jest-expo/ios`, `jest-expo/android` and `jest-expo/web` jest presets.

mdjastrzebski:
@abmantis Ok, that's interesting. So, when you run jest-expo with jest-expo/web preset, does it actually render things as `<View>`s, etc with this `data-testid` attribute or it goes to `<div>`s, etc with `data-testid`?

abmantis:
@mdjastrzebski yes, I think it renders JSDom. Here's a snapshot:

```
exports[`<ColumnCard /> renders correctly 1`] = `
<div
  className="css-view-1dbjc4n"
  data-testid="container"
  onMoveShouldSetResponder={[Function]}
  onMoveShouldSetResponderCapture={[Function]}
  onResponderEnd={[Function]}
  onResponderGrant={[Function]}
  onResponderMove={[Function]}
  onResponderReject={[Function]}
  onResponderRelease={[Function]}
  onResponderStart={[Function]}
  onResponderTerminate={[Function]}
  onResponderTerminationRequest={[Function]}
  onStartShouldSetResponder={[Function]}
  onStartShouldSetResponderCapture={[Function]}
>
  <div
    aria-label="Card touchable area"
    className="css-view-1dbjc4n"
    data-focusable={true}
    onKeyDown={[Function]}
    onKeyUp={[Function]}
    onResponderGrant={[Function]}
    onResponderMove={[Function]}
    onResponderRelease={[Function]}
    onResponderTerminate={[Function]}
    onResponderTerminationRequest={[Function]}
    onStartShouldSetResponder={[Function]}
    style={
      Object {
        "MozTransitionProperty": "opacity",
        "MozUserSelect": "none",
        "WebkitTransitionDuration": "0.15s",
        "WebkitTransitionProperty": "opacity",
        "WebkitUserSelect": "none",
        "cursor": "pointer",
        "msTouchAction": "manipulation",
        "msUserSelect": "none",
        "touchAction": "manipulation",
        "transitionDuration": "0.15s",
        "transitionProperty": "opacity",
        "userSelect": "none",
      }
    }
    tabIndex="0"
  >
    <div
      className="css-view-1dbjc4n"
    >
      <CardBase
        title="What a great card I am"
      />
    </div>
  </div>
</div>
`;

I was using the getAllByTestId in a test that is common to all platforms. I'm running it in ios+android+web presets (but not node).
```

thymikee:
Why not use `react-testing-library` for these web tests then?

mdjastrzebski:
@thymikee: It seems that @abmantis is getting his error with `getByTestId` when running his generic tests on `jest-expo/web` preset. He does not write web-specific tests, but rather uses the same tests on ios/android/web.

mdjastrzebski:
Lets recap the facts and findings:
* RNTL was written with "generic" jest setup in mind
* jest generic preset seems to actually be Platform.ios === true + React Test Renderer.
* jest-expo introduced presets that run tests under different Platform values: ios, android, web, node.
* when running Jest Expo in web preset, it not only sets Platform.web === true, but also performs full web rendering up to divs, and runs it JSDom env.
* when you write generic tests using RNTL and run them using `jest-expo/universal` preset, it runs them in ios, android, web. But because the web output env is different then ios & android, these tests fail.

thymikee:
I see, but the same test will run 1 time for each platform anyway, essentially having the same test ran 3 times. It may be a good idea to write web-specific tests separately, e.g. with `.web` extension (or however you configure Jest to do so).

Querying by testID is only a tip of the ice berg. We have a whole range of accessibility queries that are not compatible with web. There's a second PR for accessibilityLabel, but we also have accessibilityStates, accessibilityHint, etc. I don't want us to monkey patch the queries like that, because it's not scalable and hard to maintain.

mdjastrzebski:
So we seems to have following options:
1) Do not support `jest-expo/web` at all. Which means users of Jest Expo + RNTL will not able to test their web specific code using generic tests. Our suggestion would be for them to write web-specific RTL tests, which doubles the work for them.
2) Full support of 'jest-expo/web'. Which means that we have to verify & potentially fix all differences connected to different test env (RN Web renderer, JSDom), which is a major effort for us. If, achieved however, this will allow users to write generic tests once, and be sure that their web-specific code runs correctly under these tests.
3) Limited support of 'jest-expo/web'. Whichm means we actknowledge that the only correct way to test web-specific code is to write web-specific tests in RTL. But we allow running generic tests on web-specific code on your own risk.

satya164:
Maybe it'll be useful for someone to create a wrapper around this and `@testing-library/react` which uses relevant library based on platform for people using expo universal preset with testing library.

thymikee:
I dig @satya164 idea.

Expo (or regular users) preset could include `moduleNameMapper` Jest config and replace `react-native-testing-library` and `@testing-library/react-native` with `@testing-library/react` in web environment. Or it could point to a file, that would require the library specific for the environment – this would allow to handle API differences (like `screen` in or lack of `ByA11y*` queries in RTL).

mdjastrzebski:
As an alternative solution to the problem, Jest Expo could provide a separatr preset that sets `Platform.web === true`, but uses the same env, as for iOS and Android. In fact, it might be good idea for them to provide two web presets: for React Test Renderer and JSDom.

abmantis:
> I see, but the same test will run 1 time for each platform anyway, essentially having the same test ran 3 times. It may be a good idea to write web-specific tests separately, e.g. with `.web` extension (or however you configure Jest to do so).

@thymikee I have some web-specific tests too. But some are common since the behavior of the app should be the same on all platforms.

> Querying by testID is only a tip of the ice berg. We have a whole range of accessibility queries that are not compatible with web. There's a second PR for accessibilityLabel, but we also have accessibilityStates, accessibilityHint, etc. I don't want us to monkey patch the queries like that, because it's not scalable and hard to maintain.

Yeah, I added only accessibilityLabel but my plan was to add the others if you guys agreed to having this in RNTL. 


Since expo is used (and recommended as an environment) in the official RN docs, and since it seems that the plan is to push jest-expo/universal forward, my reasoning was that it made sense to support RN web on RNTL. But, I'm newb to this whole RN thing :)

abmantis:
@mdjastrzebski @thymikee Any further comments on this? I still think we should add web support to RNTL :)

mdjastrzebski:
After considering the above arguments, I am against supporting React Native Web rendered to JS Dom, as this is not our use case. We probably could support React Native Web rendered by React Test Renderer, but to do so, Jest Expo should provide a separate preset for such case.

byCedric:
@mdjastrzebski isn't `jest-expo/web` the separate Jest preset for web rendering?

mdjastrzebski:
@byCedric AFAIK `jest-expo/web` operates in a quite different manner then `jest-expo/ios` and `jest-expo/android`:

* Native tests (`jest-expo/ios', `jest-expo/android`) are rendered with React Test Rendered to `<View>`, `<Text>`, etc. This is the environment we expect and support in React Native Testing Library. RNTL is providing some runtime environment features that simulate React Native env (like, event firing, filtering out composite components, handling for `disabled` prop, etc).

* Web tests however are rendered to `<div>`, `<span>`, etc with React Native Web renderer, and run in JS DOM environment. RNTL will still try provide it's runtime env features, with unexpected consequences.

abmantis:
Ok, so should I close this then?

Also, any recommendation on how can I proceed? I want to continue using this library but also jest-expo/web+android

joaogn:
I'm using this workaround. 
```  
    let loginButton: ReactTestInstance;
    if (Platform.OS === 'web') {
      const props = { 'data-testid': 'loginButton' };
      loginButton = UNSAFE_getByProps(props);
    } else {
      loginButton = getByTestId('loginButton');
    }
```

satya164:
@byCedric have you considered doing the same thing `jest-expo/web` as done in `jest-expo/android`/`jest-expo/ios`, i.e. set appropriate `Platform`, import correct file extension etc.? And make it optional to render to JSDOM?

I'd guess it's a common issue to run into problems when writing multi-platform tests with current way as you can't really assert same things when the output format is very different. e.g. `testId` isn't the only thing, a lot of attributes would be translated to different things when rendering to a JSDOM environment.

PR #408, opened 2020-06-27
Support react-native-web in accessibilityLabel queries

abmantis:
### Summary

Because react-native-web uses "aria-label" instead of "accessibilityLabel", accessibilityLabel queries were failing to get elements. **NOTE:** If this approach is ok, I can check the remaining a11y queries and add support for them too.

### Test plan

Run tests with react-native-web. I'm using jest-expo/web preset for that.


mdjastrzebski:
Related to #407. Waiting for info about @abmantis test setup.

mdjastrzebski:
See discussion in #407

Issue #442, opened 2020-07-16
Migration guide?


kentcdodds:
We're in the process of completely replacing https://github.com/testing-library/native-testing-library with this repo.

Would anyone be interested in writing a migration guide?

balavishnuvj:
I could help.

thymikee:
Awesome @balavishnuvj! Let me know if you need any help. So far, in terms of differences between the libs, I'm only aware of the "disabled" property being specially handled by the second library. Is there anything else?

balavishnuvj:
@thymikee I don't understand in and out of the library as of now. Would need your help. Also, this would one of my first open-source contributions.

ryanjwessel:
I am working on migrating a project from `@testing-library/react-native` to this one now.

The first difference I noticed was that `'@testing-library/react-native/cleanup-after-each',` in the `jest.config.js` was no longer required because it happens by default in this library ([see docs here](https://callstack.github.io/react-native-testing-library/docs/migration-v2/#auto-cleanup)).

Second, there is a difference in querying by accessibility labels between the two libraries. `@testing-library/react-native` uses `queryByLabelText` ([see here](https://www.native-testing-library.com/docs/api-queries#bylabeltext)), whereas this library uses `queryByAccessibilityLabel`, or `queryByA11yLabel` ([see here](https://callstack.github.io/react-native-testing-library/docs/api-queries/#bya11ylabel-byaccessibilitylabel)).

Hope this helps!

thymikee:
I'm happy to accept a PR that aliases `queryByA11yLabel` to `queryByLabelText` to ease the migration :)

ryanjwessel:
I would be happy to take on this PR, but it would be my first time contributing to the project.

To clarify, we would alias any queries ending in `ByLabelText` to use `ByA11yLabel` instead? Should we display some sort of warning that this is temporary to support migration from the other library?

thymikee:
Great! No warning is needed, `queryByAccessibilityLabel` is already an alias of `queryByA11yLabel`, so it's a matter of extending that to another value. Should be fairly straight-forward I hope :)

ryanjwessel:
@thymikee Just put a PR up for it now. I think I see what you mean, hopefully this was what you were thinking. 😄

vanGalilea:
Hi all, 

Just migrated my real world examples repo, was pretty straightforward here are the diffs:

[...]

PR #445, opened 2020-07-17
feat: introduce ByLabelText, ByRole and ByHintText a11y aliases

ryanjwessel:
# Summary

Users migrating from `@testing-library/react-native` will have to update their accessibility queries due to a slightly different naming convention. This PR adds aliases for the `@testing-library/react-native` queries to be usable in this library.

This is motivated by the thread on Issue #442 

### Test plan

I am not sure about testing this. I can make additions to this PR if needed. The test suite passes, and I see `a11yAPI.test.js` uses only the `a11y` version of the queries. 

### To Note

The queries for placeholder text are not one-to-one between the two libraries, but I don't think those are set up for aliasing yet. All of the `@testing-library/react-native` placeholder queries are appended with `Text`, whereas this library does not do that.


ryanjwessel:
Thanks @thymikee! I duplicated the tests and updated the types. I'm not quite sure what `typings/__tests__/index.test.tsx` is doing, but I updated that as well. I can revert that change if needed.

ryanjwessel:
Fair point @mdjastrzebski. I think they should fall under the same test blocks since they are covering the same functionality. I'll update this now.

ryanjwessel:
I have consolidated the tests, but there were some issues with this. I added some `FIXME` tags where the tests were failing due to the consolidation. Two out of three seem to have trouble with async during `...AllBy` queries. I haven't had time to dig into why though.

Also, I updated the docs to include these new queries.

src/__tests__/a11yAPI.test.js
```diff
@@ -62,133 +62,255 @@ function Section() {
...
 
   expect(getByA11yLabel(BUTTON_LABEL).props.accessibilityLabel).toEqual(
     BUTTON_LABEL
   );
+  expect(getByLabelText(BUTTON_LABEL).props.accessibilityLabel).toEqual(
+    BUTTON_LABEL
+  );
   const button = queryByA11yLabel(/button/g);
   expect(button && button.props.accessibilityLabel).toEqual(BUTTON_LABEL);
+  const buttonAlias = queryByA11yLabel(/button/g);
```
mdjastrzebski, line 83:
```suggestion
  const buttonAlias = queryByLabelText(/button/g);
```

src/__tests__/a11yAPI.test.js
```diff
@@ -62,133 +62,255 @@ function Section() {
...
+  expect(queryByHintText(NO_MATCHES_TEXT)).toBeNull();
 
   expect(() => getByA11yHint(TEXT_HINT)).toThrow(FOUND_TWO_INSTANCES);
+  expect(() => getByHintText(TEXT_HINT)).toThrow(FOUND_TWO_INSTANCES);
   expect(() => queryByA11yHint(TEXT_HINT)).toThrow(FOUND_TWO_INSTANCES);
+  expect(() => queryByHintText(TEXT_HINT)).toThrow(FOUND_TWO_INSTANCES);
 
   const asyncButton = await findByA11yHint(BUTTON_HINT);
   expect(asyncButton.props.accessibilityHint).toEqual(BUTTON_HINT);
+  const asyncButtonAlias = await findByA11yHint(BUTTON_HINT);
```
mdjastrzebski, line 188:
```suggestion
  const asyncButtonAlias = await findByHintText(BUTTON_HINT);
```

src/__tests__/a11yAPI.test.js
```diff
@@ -62,133 +62,255 @@ function Section() {
...
+  expect(queryByRole(NO_MATCHES_TEXT)).toBeNull();
 
   expect(() => getByA11yRole('link')).toThrow(FOUND_TWO_INSTANCES);
+  expect(() => getByRole('link')).toThrow(FOUND_TWO_INSTANCES);
   expect(() => queryByA11yRole('link')).toThrow(FOUND_TWO_INSTANCES);
+  expect(() => queryByRole('link')).toThrow(FOUND_TWO_INSTANCES);
 
   const asyncButton = await findByA11yRole('button');
   expect(asyncButton.props.accessibilityRole).toEqual('button');
+  const asyncButtonAlias = await findByA11yRole('button');
```
mdjastrzebski, line 274:
```suggestion
  const asyncButtonAlias = await findByRole('button');
```

src/__tests__/a11yAPI.test.js
```diff
@@ -62,133 +62,255 @@ function Section() {
...
   );
+  /**
+   * FIXME:
+   * Additional await expect yields the following:
+   * Expected substring: "No instances found"
+   * Received message:   "Unable to find node on an unmounted component."
+   */
+  // await expect(findAllByHintText(NO_MATCHES_TEXT)).rejects.toThrow(
+  //   NO_INSTANCES_FOUND
+  // );
```
thymikee, line 237: How about not touching the existing tests and only checking if `findByA11yHint` et. al. reference the same function as `finByHintText`? These are aliases, so we only if it's possible to use them (with typings tests) and they do exactly the same thing (by checking if they're referentially equal).

src/__tests__/a11yAPI.test.js
ryanjwessel: Thanks for catching these. I rushed this last commit and forgot to actually change the queries to their aliases 🥴

I'll have a better edit up soon.

src/__tests__/a11yAPI.test.js
ryanjwessel: This would likely be a lot cleaner than the current approach. Plus, it gives us the confidence that the aliases are referring to the same thing. I'll push an update for this soon, thanks again!

ryanjwessel:
Thanks for your review @thymikee and @mdjastrzebski! I just pushed an updated a11yAPI test file that asserts referential equality for the query aliases. The test suite is passing, and the diff is a lot simpler now. 😄

Please let me know if there are any other changes either of you feel are necessary.

CONTRIBUTING.md
```diff
@@ -32,7 +32,7 @@ Our pre-commit hooks verify that your commit message matches this format when co
 
 ### Linting and tests
 
-We use `flow` for type checking, `eslint` with `prettier` for linting and formatting the code, and `jest` for testing. Our pre-commit hooks verify that the linter and tests pass when commiting. You can also run the following commands manually:
+We use `flow` for type checking, `eslint` with `prettier` for linting and formatting the code, and `jest` for testing. Our pre-commit hooks verify that the linter and tests pass when committing. You can also run the following commands manually:
```
thymikee, line 35: thanks for theses :)

src/__tests__/a11yAPI.test.js
```diff
@@ -62,10 +62,20 @@ function Section() {
...
+    queryByA11yLabel,
+    queryByLabelText,
+    findByA11yLabel,
+    findByLabelText,
+  } = render(<Section />);
+
+  // Assert that query aliases are referencing the same function
+  expect(getByA11yLabel).toBe(getByLabelText);
+  expect(queryByA11yLabel).toBe(queryByLabelText);
+  expect(findByA11yLabel).toBe(findByLabelText);
```
thymikee, line 78: This looks way better now. Can we extract all reference equality checks to a separate test? Something like:
```js
test('a11y queries have aliases', () => {
const {
     getByA11yLabel,
     getByLabelText,
     queryByA11yLabel,
     queryByLabelText,
     findByA11yLabel,
     findByLabelText,
     ... // "query all by" here
   } = render(<Section />);

    // Assert that query aliases are referencing the same function
   expect(getByA11yLabel).toBe(getByLabelText);
   expect(queryByA11yLabel).toBe(queryByLabelText);
   expect(findByA11yLabel).toBe(findByLabelText);
})
```
this way we don't need to alter existing tests

ryanjwessel: Even better! I like this. I'll push an update shortly 👍

thymikee:
Thank you very much for you contribution @ryanjwessel, appreciated!

ryanjwessel:
Thank you both for your help! This was my first ever contribution to OSS, so I am really grateful for your guidance in contributing to this project. Glad I was able to contribute!

PR #460, opened 2020-07-27
feat: prevent events on disabled elements

### LINKED ISSUE #28: fireEvent seems to always trigger event (on disabled TouchableOpacity)

mdjastrzebski:
### Summary
Prevent calling event handlers for disabled components.

Fixes #28 

### Test plan
Added tests for `TouchableOpacity` and `Pressable` in two configuration: lone element (expected throw no handler found) and nested elements: inner disabled, outer handling event.

### DIFF
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -217,7 +217,7 @@ test('should pass event up on disabled TouchableOpacity', () => {
   expect(handleOuterPress).toHaveBeenCalledTimes(1);
 });
 
-test.only('should pass event up on disabled Pressable', () => {
+test('should pass event up on disabled Pressable', () => {
   const handleInnerPress = jest.fn();
   const handleOuterPress = jest.fn();
   const screen = render(
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -224,3 +224,21 @@ test('should pass event up on disabled Pressable', () => {
   expect(handleInnerPress).not.toHaveBeenCalled();
   expect(handleOuterPress).toHaveBeenCalledTimes(1);
 });
+
+const TestComponent = ({ onPress }) => {
+  return (
+    <TouchableOpacity onPress={onPress}>
+      <Text>Trigger Test</Text>
+    </TouchableOpacity>
+  );
+};
+
+test('is not fooled by non-native disabled prop', () => {
+  const handlePress = jest.fn();
+  const screen = render(
+    <TestComponent onPress={handlePress} disabled={true} />
+  );
+
+  fireEvent.press(screen.getByText('Trigger Test'));
+  expect(handlePress).toHaveBeenCalledTimes(1);
+});
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -3,6 +3,7 @@ import React from 'react';
 import {
   View,
   TouchableOpacity,
+  Pressable,
   Text,
   ScrollView,
   TextInput,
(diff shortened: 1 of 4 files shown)


src/fireEvent.js
```diff
@@ -9,9 +9,15 @@ const findEventHandler = (
 ) => {
   const eventHandler = toEventHandlerName(eventName);
 
-  if (typeof element.props[eventHandler] === 'function') {
+  if (
+    typeof element.props[eventHandler] === 'function' &&
+    element.props.disabled !== true
```
thymikee, line 14: This will work fine in most use cases, but since we check not only host components, it may prevent firing events on custom components that happen to have prop named `disabled` that doesn't prevent events.

src/fireEvent.js
```diff
@@ -5,13 +5,25 @@ import { ErrorWithStack } from './helpers/errors';
...
+  const nearestHostComponent = isHostComponent
+    ? element
+    : nearestHostDescendent;
+
   const eventHandler = toEventHandlerName(eventName);
 
-  if (typeof element.props[eventHandler] === 'function') {
+  if (
+    typeof element.props[eventHandler] === 'function' &&
+    nearestHostComponent?.props.onStartShouldSetResponder?.() !== false
```
thymikee, line 20: Can we calculate this check once and reuse?

src/fireEvent.js
```diff
@@ -5,13 +5,25 @@ import { ErrorWithStack } from './helpers/errors';
 const findEventHandler = (
   element: ReactTestInstance,
   eventName: string,
-  callsite?: any
+  callsite?: any,
+  nearestHostDescendent?: ReactTestInstance
```
thymikee, line 9: can we flip the order so `callsite` stays at the end?

mdjastrzebski:
I've managed to implement this using `props.onStartShouldSetResponder()`, the same check that React Native uses to verify whether it should start events. The problematic issues has been that, when using React Test Renderer output, event handlers like `onPress` are usually props of composite components while `props.onStartShouldSetResponder()` is available on a descedant host component.

This implementation captures the last host component when traversing up the element tree when looking for event handler.

src/fireEvent.js
mdjastrzebski: I've tried it, but the drawback is that in the root call-site it changes

```js
const handler = findEventHandler(element, eventName, callsite);
```

to
```js
const handler = findEventHandler(element, eventName, undefined, callsite);
```

With that undefined argument actually being implementation details for recursive function.

thymikee: 👍

src/__tests__/fireEvent.test.js
```diff
@@ -163,3 +164,63 @@ test('event with multiple handler parameters', () => {
...
+      </TouchableOpacity>
+    </TouchableOpacity>
+  );
+
+  fireEvent.press(screen.getByText('Inner Trigger'));
+  expect(handleInnerPress).not.toHaveBeenCalled();
+  expect(handleOuterPress).toHaveBeenCalledTimes(1);
+});
+
+test('should pass event up on disabled Pressable', () => {
```
thymikee, line 212: Can we add one more test with a component that has "disabled" prop but doesn't pass it to the native View?

thymikee:
Please update the migration guide accordingly as well :)

mdjastrzebski:
Updated docs

PR #461, opened 2020-07-28
feat: add `getQueriesForElement` alias to `within`

### LINKED ISSUE #457: feat: introduce `getQueriesForElement` as alias for `within`.
## Describe the Feature
TL/RN v6 has provided `getQueriesForElement` for scoping queries to given element & it's children. RNTL v2 had a similar `within` function. Both of these were modeled on similar RTLs helper. In [RTL both names are available (as alias)](https://testing-library.com/docs/dom-testing-library/api-helpers#within-and-getqueriesforelement-apis) to provide the same function.

## Possible Implementations
Provide `getQueriesForElement` as alias to `within`.

## Benefits

* compatibility with RTL
* easier migration path for TL/RN v6 users.

## Related Issues
Migration to v7, #456

mdjastrzebski:
### Summary

TL/RN v6 has provided `getQueriesForElement` for scoping queries to given element & it's children. RNTL v2 had a similar `within` function. Both of these were modeled on similar RTLs helper. In RTL both names are available (as alias) to provide the same function. 

I've added `getQueriesForElement` as an alias to `within`.

Resolves #457 

### Test plan
* tests for checking that functions are the same
* typescript type checks

### DIFF
--- a/typings/__tests__/index.test.tsx
+++ b/typings/__tests__/index.test.tsx
@@ -9,6 +9,7 @@ import {
   waitForElementToBeRemoved,
   act,
   within,
+  getQueriesForElement,
 } from '../..';
 
 interface HasRequiredProp {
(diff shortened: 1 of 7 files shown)


typings/__tests__/index.test.tsx
```diff
@@ -306,6 +307,18 @@ const withinGet: Array<ReactTestInstance> = [
   within(instance).getByRole('button'),
   within(instance).getByA11yState({ busy: true }),
   within(instance).getByA11yValue({ min: 10 }),
+  getQueriesForElement(instance).getByText('Test'),
```
thymikee, line 310: I wish we could do something like `within === getQueriesForElement` instead, but TS is structural and this will likely pass anyway. And I wouldn't like to resort to [hacks like described here](https://isthisit.nz/posts/2020/typescript-ii-nominal-and-structural-typing/)

PR #470, opened 2020-07-31
fix: fire event should not throw for disabled handler

### LINKED ISSUE #469: No handler function found with fireEvent.press on disabled TouchableOpacity
## Describe the bug

When running `fireEvent.press` on a disabled TouchableOpacity, the following error appears: ` No handler function found for event: "press"`

## Expected behavior

Should not raise an error

## Steps to Reproduce

Run the following

```
import { TouchableOpacity } from 'react-native'

describe.only('Test TouchableOpacity', () => {
    const Button = ({ onPress }) => {
        return <TouchableOpacity testID='testBtn' disabled={true} onPress={onPress}></TouchableOpacity>
    }

    const onPressSpy = sinon.spy()

    it('onPress does not trigger', () => {
        const { getByTestId } = render( <Button onPress={onPressSpy}/>)
        fireEvent.press(getByTestId('testBtn'))
        expect(props.onPress.callCount).toBe(0)
    })
})
```

## Versions

```
  npmPackages:
    @testing-library/react-native: ^7.0.0 => 7.0.0 
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.0 => 0.63.0 
    react-test-renderer: ^16.13.0 => 16.13.1 
```

mdjastrzebski:
### Summary

Fixes #469 

RNTL has a check that throws error when event handler missing. This is intended to inform developer that there is a mismatch between test assertions and components under test. This PR fixes a cases when this check throws error while component has a disabled event handler.

### Test plan
* updated tests for present & disabled handler not to throw
* we already have test for missing handler (not disabled), and for success cases

### DIFF
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -173,7 +173,7 @@ test('should not fire on disabled TouchableOpacity', () => {
     </TouchableOpacity>
   );
 
-  expect(() => fireEvent.press(screen.getByText('Trigger'))).not.toThrow();
+  fireEvent.press(screen.getByText('Trigger'));
   expect(handlePress).not.toHaveBeenCalled();
 });
 
@@ -185,7 +185,7 @@ test('should not fire on disabled Pressable', () => {
     </Pressable>
   );
 
-  expect(() => fireEvent.press(screen.getByText('Trigger'))).not.toThrow();
+  fireEvent.press(screen.getByText('Trigger'));
   expect(handlePress).not.toHaveBeenCalled();
 });
 
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -168,9 +168,11 @@ test('event with multiple handler parameters', () => {
 test('should not fire on disabled TouchableOpacity', () => {
   const handlePress = jest.fn();
   const screen = render(
-    <TouchableOpacity onPress={handlePress} disabled={true}>
-      <Text>Trigger</Text>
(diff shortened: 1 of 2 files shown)


src/__tests__/fireEvent.test.js
```diff
@@ -173,9 +173,7 @@ test('should not fire on disabled TouchableOpacity', () => {
     </TouchableOpacity>
   );
 
-  expect(() => fireEvent.press(screen.getByText('Trigger'))).toThrow(
-    'No handler function found for event: "press"'
-  );
+  expect(() => fireEvent.press(screen.getByText('Trigger'))).not.toThrow();
```
thymikee, line 176: we could get rid of the expect altogether here and there
```suggestion
fireEvent.press(screen.getByText('Trigger'));
```

src/fireEvent.js
```diff
@@ -6,30 +6,48 @@ const findEventHandler = (
...
   const hostElement = isHostComponent ? element : nearestHostDescendent;
   const isEventEnabled =
     hostElement?.props.onStartShouldSetResponder?.() !== false;
+  if (handler && isEventEnabled) return handler;
 
-  const eventHandlerName = toEventHandlerName(eventName);
+  // Do not bubble event to the root element
+  if (element.parent === null || element.parent.parent === null) {
+    if (hasHandler) return null;
+    else
```
thymikee, line 24: please use curly braces in if/else statements. I find it easier to scan through the code, especially if expressions inside are multiline

PR #489, opened 2020-08-07
fix: Return closest Text matching a query

AugustinLF:
### Summary
When using `getByText()` on a component using nested `Text`s, it should match the closest one.
```jsx
<Text id={1}>
  <Text id={2}>My text</Text>
</Text>
```

I've found this bug when trying to upgrade from `@5.X` to `@7.X`, this breaks one of my tests where an `onPress` handler is attached to the lower Text, and since `getByText` returns  its parent, using `fireEvent.press(getByText('My text'))` doesn't work.

### Test plan

The added tests should show the covered case. I added an another test to be sure I would not introduce a regression.

### DIFF
--- a/src/__tests__/queryByApi.test.js
+++ b/src/__tests__/queryByApi.test.js
@@ -74,6 +74,21 @@ test('queryByText with nested Text components return the closest Text', () => {
   expect(getByText('My text').props.nativeID).toBe('2');
 });
 
+test('queryByText with nested Text components each with text return the lowest one', () => {
+  const NestedTexts = () => (
+    <Text nativeID="1">
+      bob
+      <Text nativeID="2">My text</Text>
+    </Text>
+  );
+
+  const { getByText } = render(<NestedTexts />);
+
+  expect(getByText('My text').props.nativeID).toBe('2');
+});
+
+
+
 test('queryByText nested <CustomText> in <Text>', () => {
   const CustomText = ({ children }) => {
     return <Text>{children}</Text>;
--- a/src/__tests__/queryByApi.test.js
+++ b/src/__tests__/queryByApi.test.js
@@ -69,9 +69,9 @@ test('queryByText with nested Text components return the closest Text', () => {
     </Text>
   );
 
-  const { getByText } = render(<NestedTexts />);
+  const { queryByText } = render(<NestedTexts />);
 
-  expect(getByText('My text').props.nativeID).toBe('2');
+  expect(queryByText('My text')?.props.nativeID).toBe('2');
 });
 
 test('queryByText with nested Text components each with text return the lowest one', () => {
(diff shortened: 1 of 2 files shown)


src/helpers/getByAPI.js
```diff
@@ -44,6 +44,18 @@ const getChildrenAsText = (children, TextComponent, textContent = []) => {
...
     if (child?.props?.children) {
+      // This means that down in the tree of a <TextComponent /> there's another TextComponent that
+      // contains a string, but the parent itself doesn't contain any other string
+      // We should then match against this child, since querying aims to always return the element
+      // the closest to the target text. For instance in this example we don't want
+      // getChildrenAsText to match on the Text of nativeID "1", but the one of nativeID "2"
+      // <Text nativeID="1">
+      //   <Text nativeID="2">My text</Text>
+      // </Text>
+      if (filterNodeByType(child, TextComponent) && textContent.length === 0) {
```
thymikee, line 55: What if the case is like this:
```jsx
<Text nativeID="1">
  bob
  <Text nativeID="2">My text</Text>
</Text>
```
Will this check still work?

AugustinLF: It seems to work! c36e5345394c967afbcb657fd34195a976dd12b9

AugustinLF:
@thymikee Is there anything I could do to help this PR to move forward? It prevents us from upgrading to the latest version :/

src/helpers/getByAPI.js
thymikee: Adjusted the comment so I can understand it better. Hope you don't mind :)

src/__tests__/queryByApi.test.js
```diff
@@ -48,7 +48,7 @@ test('queryByText not found', () => {
   ).toBeFalsy();
 });
 
-test('queryByText nested multiple <Text> in <Text>', () => {
+test('queryByText nested text across multiple <Text> in <Text>', () => {
   expect(
     render(
       <Text>
```
mdjastrzebski, line 54: Could we add similar `nativeID` check for this test too? IMO it should return top level `Text`.

src/helpers/getByAPI.js
```diff
@@ -44,6 +44,14 @@ const getChildrenAsText = (children, TextComponent, textContent = []) => {
     }
 
     if (child?.props?.children) {
+      // Bail on traversing text children down the tree if current node (child)
+      // has no text. In such situations, react-test-renderer will traverse down
+      // this tree in a separate call and run this query again. As a result, the
+      // query will match a deeper text node that may have text some content.
```
mdjastrzebski, line 50:
```suggestion
      // query will match a deeper text node that has the same text content.
```

thymikee: My intention was "my have some text content" because it may happen it's still empty. Make sense to update the text this way?

mdjastrzebski: "some" or "same" text content?

thymikee: "some". Like:
```jsx
<Text> 
  // no text
  <Text>
    // no text here as well
    <Text>some text content</Text>
  </Text>
</Text>
```

mdjastrzebski: Maybe we can go with: "As a result, ByText queries will match a deepest text node that matches requested text.", wdyt?

thymikee: Sounds good 👍 "a deepest" -> "the deepest", rest is fine

thymikee:
Thank you @AugustinLF !

AugustinLF:
Thanks a lot, was a pleasure to help get that in, thanks for the final fixes!

PR #507, opened 2020-08-15
Switch to Yarn workspaces

### LINKED ISSUE #493: Convert repository to monorepository
## Describe the Feature
The idea behind this PR is to convert the repository into a monorepo, allowing us to have multiple packages and allowing inter-dependence between them.

## Possible Implementations
The proposed folder structure are as follows:
```sh
/ # Project root. Contains all configurations
/ examples   # Contains all examples. Will be using the `react-native-testing-library` package
/ website    # Contains the RNTL website
/ packages
    / react-native-testing-library
```

cross19xx:
### Summary
Refactoring the codebase to use yarn workspaces. Resolved #493


cross19xx:
### Steps leading to this change:
- Moved `eslint`, `jest` and  its dependencies to the project root

- Moved `.release-it.json`, `babel.config.js`, `dont-cleanup-after-each.js`, `pure.js`, `package.json`, `src` and `typings` into `packages/react-native-testing-library`

- Due to issues moving flow, I kept `flow-typed` and `.flowconfig` at the project root and installed [`flow-mono-cli`](https://github.com/ImmoweltGroup/flow-mono-cli). The symlinks are created after post install on the project root

- Fixed docusaurus theme configuration by changing `themeConfig.links` to `themeConfig.items` in the `docusaurus.config.js`

- All libraries that were installed in the project root had their versions changed to `"*"`

- Switched all examples to use `expo` to reduce project size since expo did not require the `android` and `ios` directory

thymikee:
I'm not convinced we need `flow-mono-cli`. I'd like to have a single Flow instance looking over whole project from the root.

cross19xx:
> I'm not convinced we need `flow-mono-cli`. I'd like to have a single Flow instance looking over whole project from the root.
> Also, the PR needs a rebase

It was more for future cases. In case another package was introduced which had different flow configurations from the global one

thymikee:
> It was more for future cases

Such future may (most likely) never come. There's no reason to add more dependencies to the project "just in case". Only when necessary – e.g. we have long-term plan of supporting such use case. Which we don't.

.circleci/config.yml
```diff
@@ -36,21 +36,21 @@ jobs:
       - attach_workspace:
           at: ~/react-native-testing-library
       - run: |
+          cd packages/react-native-testing-library
```
thymikee, line 39: linting and type checks should go from the root, please remove and adjust configuration if needed

package.json
```diff
@@ -1,79 +1,23 @@
...
+    "react-native": "^0.63.2"
   },
-  "jest": {
-    "preset": "react-native",
-    "moduleFileExtensions": [
-      "js",
-      "json"
-    ],
-    "rootDir": "./src"
+  "devDependencies": {
```
thymikee, line 13: FYI, there's no need to differentiate between deps and devDeps in workspace root, I don't mind it though, your call :)

packages/react-native-testing-library/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "babel-jest": "*",
+    "conventional-changelog-cli": "^2.0.11",
+    "dedent": "^0.7.0",
+    "eslint": "*",
+    "flow-bin": "^0.122.0",
+    "flow-copy-source": "^2.0.9",
+    "jest": "*",
+    "react": "*",
+    "react-native": "*",
+    "react-test-renderer": "*",
```
thymikee, line 49: can we get rid of `*` in our dev deps and bring back the versions we had in original package.json? I don't mind having `*` in examples though

PR #516, opened 2020-08-20
Fix getByTestId nested custom text edge case

### LINKED ISSUE #515: Nested custom Text componentns not working correctly
## Describe the bug

If custom `Text` components are nested and find with text, then the wrong other element is retrieved.

## Steps to Reproduce

```tsx
const CustomText = ({ children, ...props }: PropsWithChildren<TextProps>) => {
  return <Text {...props}>{children}</Text>;
};
CustomText.displayName = 'CustomText';

const NestedText = () => {
  return (
    <View>
      <CustomText testID="outer">
        <CustomText testID="inner1"></CustomText>
        <CustomText testID="inner2">Hello</CustomText>
      </CustomText>
    </View>
  );
};

it('wrong result', () => {
  const { getByText, debug } = render(<NestedText />);

  const element = getByText(/Hello/);
  debug();

  expect(element.props.testID).toBe('inner2');
});
```

```
  console.log
    <View>
      <Text
        testID="outer"
      >
        <Text
          testID="inner1"
        />
        <Text
          testID="inner2"
        >
          Hello
        </Text>
      </Text>
    </View>

      at debugDeep (node_modules/@testing-library/react-native/build/helpers/debugDeep.js:19:13)


Error: expect(received).toBe(expected) // Object.is equality

Expected: "inner2"
Received: "outer"
<Click to see difference>


    at Object.<anonymous> (/Users/mj/Desktop/Turing/MathKing/src/components/shared/AiTeacherCommentText.test.tsx:79:32)
    at Object.asyncJestTest (/Users/mj/Desktop/Turing/MathKing/node_modules/jest-jasmine2/build/jasmineAsyncInstall.js:100:37)
    at /Users/mj/Desktop/Turing/MathKing/node_modules/jest-jasmine2/build/queueRunner.js:47:12
    at new Promise (<anonymous>)
    at mapper (/Users/mj/Desktop/Turing/MathKing/node_modules/jest-jasmine2/build/queueRunner.js:30:19)
    at /Users/mj/Desktop/Turing/MathKing/node_modules/jest-jasmine2/build/queueRunner.js:77:41
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

```

## Versions

```
  npmPackages:
    @testing-library/react-native: 7.0.2 => 7.0.2 
    react: 16.13.1 => 16.13.1 
    react-native: 0.63.2 => 0.63.2 
    react-test-renderer: 16.13.1 => 16.13.1 

```

mym0404:
### This PR will close #515 

- Refactor `getChildrenAsText` function in `getByAPI.js`
- Add `getByApi.test.js`, `queryByApi.test.js` test cases about nested `<Text>` component

#489 fixed case like the following.

```js
test('getByText with nested native Text component get closest Text', () => {
  const NestedText = () => {
    return (
      <Text testID="outer">
        <Text testID="inner">Test</Text>
      </Text>
    );
  };

  const { getByText } = render(<NestedText />);

  expect(getByText('Test').props.testID).toBe('inner');
});
```

But not 

```js
test('getByText with nested custom Text component get closest Text', () => {
  const CustomText = ({ children, ...rest }) => (
    <Text {...rest}>{children}</Text>
  );

  const NestedText = () => {
    return (
      <CustomText testID="outer">
        <CustomText testID="inner">Test</CustomText>
      </CustomText>
    );
  };

  const { getByText } = render(<NestedText />);

  expect(getByText('Test').props.testID).toBe('inner');
});
```

This is because `filterNodeByType(child, ...)` is returned as `false` for custom text component.

So I handling the edge case for the above.
I can't sure this is the correct way, but this feature doesn't break before test cases.

Also, I refactored type checking code whether `child` is `string` or `number` for adding to `textContent`.

### DIFF
--- a/src/helpers/getByAPI.js
+++ b/src/helpers/getByAPI.js
@@ -31,16 +31,21 @@ const getNodeByText = (node, text) => {
   }
 };
 
-const getChildrenAsText = (children, TextComponent, textContent = []) => {
-  React.Children.forEach(children, (child) => {
-    if (typeof child === 'string') {
-      textContent.push(child);
-      return;
-    }
+const isTextContentType = (candidate) =>
+  ['string', 'number'].includes(typeof candidate);
 
-    if (typeof child === 'number') {
-      textContent.push(child.toString());
-      return;
+const getChildrenAsText = (children, TextComponent, textContent = []) => {
+  if (
+    typeof children === 'object' &&
+    isTextContentType(children.props?.children)
+  ) {
+    return textContent;
+  }
+
+  React.Children.forEach(children, (child) => {
+    if (isTextContentType(child)) {
+      textContent.push(child + '');
+      return textContent;
     }
 
     if (child?.props?.children) {
@@ -49,7 +54,7 @@ const getChildrenAsText = (children, TextComponent, textContent = []) => {
       // this tree in a separate call and run this query again. As a result, the
       // query will match the deepest text node that matches requested text.
       if (filterNodeByType(child, TextComponent) && textContent.length === 0) {
-        return;
+        return textContent;
       }
 
       getChildrenAsText(child.props.children, TextComponent, textContent);
--- a/src/__tests__/getByApi.test.js
+++ b/src/__tests__/getByApi.test.js
@@ -35,3 +35,54 @@ test('getByTestId returns only native elements', () => {
     'No instances found with testID: myComponent'
   );
 });
+
+test('getByText with nested native Text component get closest Text', () => {
+  const NestedText = () => {
+    return (
+      <Text testID="outer">
+        <Text testID="inner">Test</Text>
+      </Text>
+    );
+  };
+
+  const { getByText } = render(<NestedText />);
+
+  expect(getByText('Test').props.testID).toBe('inner');
+});
+
+test('getByText with nested multiple custom Text component get closest Text', () => {
+  const CustomText = ({ children, ...rest }) => (
+    <Text {...rest}>{children}</Text>
+  );
+
+  const NestedText = () => {
+    return (
+      <CustomText testID="outer">
+        <CustomText testID="inner1">Test1</CustomText>
+        <CustomText testID="inner2">Test2</CustomText>
+      </CustomText>
+    );
+  };
+
+  const { getByText } = render(<NestedText />);
+
+  expect(getByText('Test1').props.testID).toBe('inner1');
+});
+
+test('getByText with nested custom Text component get closest Text', () => {
+  const CustomText = ({ children, ...rest }) => (
+    <Text {...rest}>{children}</Text>
+  );
+
+  const NestedText = () => {
+    return (
+      <CustomText testID="outer">
+        <CustomText testID="inner">Test</CustomText>
+      </CustomText>
+    );
+  };
+
+  const { getByText } = render(<NestedText />);
+
+  expect(getByText('Test').props.testID).toBe('inner');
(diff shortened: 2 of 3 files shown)


mym0404:
@thymikee I tried :)

src/helpers/getByAPI.js
```diff
@@ -31,16 +31,21 @@ const getNodeByText = (node, text) => {
...
+const isTextContentType = (candidate) =>
+  ['string', 'number'].includes(typeof candidate);
+
 const getChildrenAsText = (children, TextComponent, textContent = []) => {
-  React.Children.forEach(children, (child) => {
-    if (typeof child === 'string') {
-      textContent.push(child);
-      return;
-    }
+  if (
```
mym0404, line 38: This part helps edge case handling.

src/helpers/getByAPI.js
```diff
@@ -31,16 +31,21 @@ const getNodeByText = (node, text) => {
...
+    typeof children === 'object' &&
+    isTextContentType(children.props?.children)
+  ) {
+    return textContent;
+  }
```
mym0404, line 43: This PR will close #515

mym0404:
I found some fail cases more... This PR should be pending now

thymikee:
The #515 mentions `getByText` with RegExp as an argument

mym0404:
Yes, I realized the cause of this problem. 

Case 1. Test failed only with `RegExp`

For example, the components exist like 
```js
    const NestedText = () => {
      return (
        <CustomText testID="outer">
          <CustomText testID="inner1">Test</CustomText>
          <CustomText testID="inner2">Test2</CustomText>
        </CustomText>
      );
    };
```

And, in `getByAPI.js`
```js
    if (isTextComponent) {
      const textChildren = getChildrenAsText(node.props.children, Text);
      if (textChildren) {
        return typeof text === 'string'
          ? text === textToTest
          : text.test(textToTest);
      }
    }
```

The `textToTest` value of `outer` Text is `TestTest2` and `inner1` is `Test`.
So with `string` expectation, then `inner1` is returned correctly. But in `RegExp`, `outer` is also matched with `/Test/`

Case 2. Test failed both
```js
    const NestedText = () => {
      return (
        <CustomText testID="outer">
          <CustomText testID="inner1"></CustomText>
          <CustomText testID="inner2">Test</CustomText>
        </CustomText>
      );
    };
```

The `textToTest` value of `outer` Text is also `Test`... so `getByText('Text')` and `getByText(/Text/)` both matched to `outer`

mym0404:
I think this is not about the bug and about the principle of the text matching test of this library.

In my case, I don't want to test like #489 with a recursive text collection way.
I just want to test the text in any `<Text>` component directly.

IMAO, the tests like the following should be failed.
```js
  expect(
    render(
      <Text>
        <CustomText>Hello</CustomText> <CustomText>World!</CustomText>
      </Text>
    ).queryByText('Hello World!')
  ).toBeTruthy();
```

But, this is just my opinion thank you for your comment. It helped me what is the main cause of the problem.
I will test my components like this.

```js
    const matches = testingLib.getAllByText(text);
    const element = matches[matches.length - 1];
```

thymikee:
> IMAO, the tests like the following should be failed

I can't agree to that. The user would actually see the "Hello World!" text and this is what we try to achieve here. We can probably come up with a better way of doing so, which works well with both string and regex matches.

mym0404:
Thank you for your concern

PR #519, opened 2020-08-30
docs: update scroll event api for flat list

verden11:
### Summary

update API docs to no longer use `getByType` as it is deprecated and will be removed. Replaces the example to use `getByTestId`


thymikee (approved):
Thanks!

Issue #513, opened 2020-08-17
baseElement is missing in v7


TAGraves:
In v6, the result of calling `render` included a `baseElement` property (see https://www.native-testing-library.com/docs/next/api-main#baseelement). `baseElement` was the root element of the render result.

In v7, `baseElement` is no longer returned from `render`.

I do wonder if it would be better to just deprecate this and include a notice in the migration guide. Looking over our tests at Root, most of them seemed to be misunderstand what `baseElement` did anyway, and so our tests were written in a way that could never fail. For example I see a lot of cases like:
```
expect(baseElement).toBeEnabled()
```

but with v6's implementation, that would always return true -- since baseElement was a View rendered by the AppContainer and couldn't be disabled. In our internal PR migrating to v7 we replaced baseElement with the results of more specific queries and didn't have any problem.

thymikee:
I'm good with not adding that (no one actually requested the feature for years) and adding a note to the migration guide.

mdjastrzebski:
@TAGraves is `baseElement` something similar to `container` from React Testing Library?

> container reference to the DOM node where the component is mounted
https://testing-library.com/docs/react-testing-library/cheatsheet

PR #533, opened 2020-09-03
fix: add note about baseElement removed from render

thymikee:
### Summary

Fixes https://github.com/callstack/react-native-testing-library/issues/513 by mentioning removal of `baseElement` in the migration guide. We _could_ add it to the library, but there's little to no push on that feature. Happy to introduce it later, if requested.

### Test plan

None


Swor71 (commented):
- typo in `because`
- I would also change `although we advise not doing so.` to `although we advise against doing so.` as it sounds more natural. (just an opinion, tho)

thymikee:
@Swor71 thanks!

Swor71 (approved):

PR #535, opened 2020-09-07
feat: RegExp support for *ByTestId queries

acatalina:
### Summary

All *AllByTestId queries support a RegExp matcher.

We are currently preparing to migrate to the the v7 of the testing library. In our tests, the lack of this feature is a massive disadvantage for our migration as we relay in this functionality quite often.

### Test plan

Tests added following preexisting tests.


src/helpers/getByAPI.js
```diff
@@ -89,6 +89,16 @@ const getTextInputNodeByDisplayValue = (node, value) => {
   }
 };
 
+const getNodeByTestId = (node, testID) => {
+  try {
+    return typeof testID === 'string'
+      ? testID === node.props.testID
+      : testID.test(node.props.testID);
+  } catch (error) {
+    throw createLibraryNotSupportedError(error);
```
thymikee, line 98: We only use `createLibraryNotSupportedError ` when importing something from `react-native`. No need for this try/catch here

thymikee (changes requested):
Awesome! Can we add this for `getByTestId` as well? :)

src/helpers/getByAPI.js
acatalina: oh, fair enough. Sorted

thymikee (approved):
Looks good 👍

acatalina:
> Awesome! Can we add this for `getByTestId` as well? :)

ha, you got me while I was updating the other *ByTestId and the docs

src/__tests__/getByApi.test.js
```diff
@@ -35,3 +35,17 @@ test('getByTestId returns only native elements', () => {
...
+    <View>
+      <Text testID="text">Text</Text>
+      <TextInput testID="textInput" />
+      <View testID="view" />
+      <Button testID="button" title="Button" onPress={jest.fn()} />
+      <MyComponent testID="myComponent" />
+    </View>
+  );
+
+  expect(getAllByTestId(/text/)).toHaveLength(2);
```
thymikee, line 50: let's add a test for `getByTestId ` as well, for completeness

acatalina: sure, sorted

mdjastrzebski (approved):
LGTM

PR #538, opened 2020-09-08
Clarify waitFor documentation in regards to timer mocks

dcalhoun:
Update documentation to explicitly state that one should not await the return or resolution of a `waitFor` call when relying upon timer mocks.


mdjastrzebski (approved):
LGTM

Issue #514, opened 2020-08-17
The string precision API is missing in v7


TAGraves:
In v6, queries could be passed string precision options so that they performed fuzzier text matching or normalized text differently (see https://www.native-testing-library.com/docs/next/api-queries#precision). These options are not available in v7. v6's behavior matches that of [dom-testing-library](https://testing-library.com/docs/dom-testing-library/api-queries#textmatch) so it is probably worth adding them back in -- those docs do a good job of describing the expected behavior of them, too.

RafikiTiki:
I'll try to tackle this

PR #541, opened 2020-09-09
feat: add text match options a.k.a string precision API

RafikiTiki:
### Summary
As mentioned in #514 current implementation lacked options for text matching customisation. This PR bridges this gap and provides smoother migration experience for users coming from native-testing-library.

### Things that need further discussion
This is a basic implementation with just one option allowing to opt-in for a partial and/or case-insensitive string match. NTL had another option for text normalisation customisation which this PR currently lacks. It's up for debate if we need this feature. 

Another thing to consider is API consistency: to avoid pushing breaking change to existing RNTL users I've put TextSearchOptions param as the last one (even in findBy* queries, coming after waitFor options; in NTL findBy* queries this param was coming second to last, before waitFor options).

### DIFF
--- a/src/__tests__/getByApi.test.js
+++ b/src/__tests__/getByApi.test.js
@@ -1,7 +1,7 @@
 // @flow
 import React from 'react';
 import { View, Text, TextInput, Button } from 'react-native';
-import { render } from '..';
+import { render, getDefaultNormalizer } from '..';
 
 const MyComponent = () => {
   return <Text>My Component</Text>;
@@ -123,7 +123,7 @@ describe('Supports normalization', () => {
     expect(getByText('Text and whitespace')).toBeTruthy();
   });
 
-  test('trim and collapseWhitespace is customizable by queryOptions param', () => {
+  test('trim and collapseWhitespace is customizable by getDefaultNormalizer param', () => {
     const testTextWithWhitespace = `  Text     and
         
         
@@ -136,8 +136,10 @@ describe('Supports normalization', () => {
 
     expect(
       getByText(testTextWithWhitespace, {
-        trim: false,
-        collapseWhitespace: false,
+        normalizer: getDefaultNormalizer({
+          trim: false,
+          collapseWhitespace: false,
+        }),
       })
     ).toBeTruthy();
   });
--- a/src/__tests__/getByApi.test.js
+++ b/src/__tests__/getByApi.test.js
@@ -108,3 +108,54 @@ describe('Supports a TextMatch options', () => {
     expect(getAllByText('detail', { exact: false })).toHaveLength(2);
   });
 });
+
+describe('Supports normalization', () => {
+  test('trims and collapses whitespace by default', () => {
+    const { getByText } = render(
+      <View>
+        <Text>{`  Text     and
+        
+        
+        whitespace`}</Text>
+      </View>
+    );
+
+    expect(getByText('Text and whitespace')).toBeTruthy();
+  });
+
+  test('trim and collapseWhitespace is customizable by queryOptions param', () => {
+    const testTextWithWhitespace = `  Text     and
+        
+        
+        whitespace`;
+    const { getByText } = render(
+      <View>
+        <Text>{testTextWithWhitespace}</Text>
+      </View>
+    );
+
+    expect(
+      getByText(testTextWithWhitespace, {
+        trim: false,
+        collapseWhitespace: false,
+      })
+    ).toBeTruthy();
+  });
+
+  test('normalizer function is customisable', () => {
+    const testText = 'A TO REMOVE text';
+    const normalizerFn = (textToNormalize: string) =>
+      textToNormalize.replace('TO REMOVE ', '');
+    const { getByText } = render(
+      <View>
+        <Text>{testText}</Text>
+      </View>
+    );
+
+    expect(
+      getByText('A text', {
+        normalizer: normalizerFn,
+      })
+    ).toBeTruthy();
+  });
+});
--- a/src/__tests__/getByApi.test.js
+++ b/src/__tests__/getByApi.test.js
@@ -50,3 +50,61 @@ test('supports a regex matcher', () => {
   expect(getByTestId(/view/)).toBeTruthy();
   expect(getAllByTestId(/text/)).toHaveLength(2);
 });
+
+describe('Supports a TextMatch options', () => {
+  test('getByText, getAllByText', () => {
+    const { getByText, getAllByText } = render(
+      <View>
+        <Text testID="text">Text and details</Text>
+        <Button
+          testID="button"
+          title="Button and a detail"
+          onPress={jest.fn()}
+        />
+      </View>
+    );
+
+    expect(getByText('details', { exact: false })).toBeTruthy();
+    expect(getAllByText('detail', { exact: false })).toHaveLength(2);
+  });
+
+  test('getByPlaceholderText, getAllByPlaceholderText', () => {
+    const { getByPlaceholderText, getAllByPlaceholderText } = render(
+      <View>
+        <TextInput placeholder={'Placeholder with details'} />
+        <TextInput placeholder={'Placeholder with a DETAIL'} />
+      </View>
+    );
+
+    expect(getByPlaceholderText('details', { exact: false })).toBeTruthy();
+    expect(getAllByPlaceholderText('detail', { exact: false })).toHaveLength(2);
+  });
+
+  test('getByDisplayValue, getAllByDisplayValue', () => {
+    const { getByDisplayValue, getAllByDisplayValue } = render(
+      <View>
+        <TextInput value={'Value with details'} />
+        <TextInput value={'Value with a detail'} />
+      </View>
+    );
+
+    expect(getByDisplayValue('details', { exact: false })).toBeTruthy();
+    expect(getAllByDisplayValue('detail', { exact: false })).toHaveLength(2);
+  });
+
+  test('with TextMatch option exact === false text search is NOT case sensitive', () => {
+    const { getByText, getAllByText } = render(
+      <View>
+        <Text testID="text">Text and details</Text>
+        <Button
+          testID="button"
+          title="Button and a DeTAil"
+          onPress={jest.fn()}
+        />
+      </View>
+    );
+
+    expect(getByText('DeTaIlS', { exact: false })).toBeTruthy();
+    expect(getAllByText('detail', { exact: false })).toHaveLength(2);
+  });
+});
--- a/src/__tests__/queryByApi.test.js
+++ b/src/__tests__/queryByApi.test.js
@@ -89,17 +89,28 @@ test('queryByText with nested Text components each with text return the lowest o
   expect(queryByText('My text')?.props.nativeID).toBe('2');
 });
 
-test('queryByText with nested Text components each with text return the top one when using not-exact text match', () => {
-  const NestedTexts = () => (
+test('queryByText with nested Text components: not-exact text match returns the most deeply nested common component', () => {
+  const { queryByText: queryByTextFirstCase } = render(
     <Text nativeID="1">
       bob
-      <Text nativeID="2">My text</Text>
+      <Text nativeID="2">My </Text>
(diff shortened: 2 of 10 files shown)


thymikee:
cc @TAGraves

website/docs/Queries.md
```diff
@@ -53,6 +53,9 @@ type ReactTestInstance = {
 };
 ```
 
+### Options
+Query first argument can be a **string** or a **regex**. Some queries accept optional argument which change string matching behaviour. See [TextMatch](api-queries#textmatch) for more info.
```
thymikee, line 57: Can you run Prettier over this file?

website/docs/Queries.md
```diff
@@ -199,6 +202,53 @@ const { getByA11yValue } = render(<Component />);
 const element = getByA11yValue({ min: 40 });
 ```
 
+## TextMatch
+Some APIs accept an object containing options affecting precision of string matching as the last argument:
```
thymikee, line 206: Let's use the wording from official docs:
```suggestion

Several APIs accept a `TextMatch` which can be a `string`, `regex` or a `function` which returns `true` for a match and `false` for a mismatch.
```

thymikee: For the reset of the text I would suggest getting as close to the official docs as possible as well. See: https://testing-library.com/docs/dom-testing-library/api-queries#precision

src/helpers/getByAPI.js
```diff
@@ -9,9 +9,32 @@ import {
...
+};
+
+const DEFAULT_TEXT_MATCH_OPTIONS: TextMatchOptions = {
+  exact: true,
+};
+
 const filterNodeByType = (node, type) => node.type === type;
 
-const getNodeByText = (node, text) => {
+const matchText = (
```
thymikee, line 22: Let's use the original implementation, which can be found here: https://github.com/testing-library/native-testing-library/blob/master/src/lib/matches.js. This will also solve the missing normalization, which I believe we can implement, even just to be more compatible with testing library in general, it's not too much code really

thymikee (changes requested):
This is grat, thank you! Left a few notes to address, but this is pretty close

typings/index.d.ts
```diff
@@ -127,28 +131,34 @@ interface QueryByAPI {
...
-    waitForOptions?: WaitForOptions
+    waitForOptions?: WaitForOptions,
+    options?: TextMatchOptions,
   ) => FindReturn;
   findByTestId: (testID: string | RegExp, waitForOptions?: WaitForOptions) => FindReturn;
   findAllByText: (
     text: string | RegExp,
-    waitForOptions?: WaitForOptions
+    waitForOptions?: WaitForOptions,
+    options?: TextMatchOptions,
```
mdjastrzebski, line 151: It seems that RTL uses different args order for all findBy queries: text, queryOptions, waitForOptions. I think we should use the same order.

typings/index.d.ts
```diff
@@ -64,19 +65,22 @@ interface GetByAPI {
 }
 
 interface QueryByAPI {
-  queryByText: (name: string | RegExp) => ReactTestInstance | null;
+  queryByText: (name: string | RegExp, options?: TextMatchOptions) => ReactTestInstance | null;
```
mdjastrzebski, line 68: RTL in their docs (but curiously not their ts typesing code) have a type TextMatch = string | RegExp, maybe we could use the same.

src/helpers/findByAPI.js
```diff
@@ -32,34 +37,68 @@ export const findAllByTestId = (instance: ReactTestInstance) => (
...
+    placeholder,
+    waitForOptions,
+    textMatchOptions
+  );
 
 export const findAllByPlaceholderText = (instance: ReactTestInstance) => (
   placeholder: string | RegExp,
-  waitForOptions: WaitForOptions = {}
+  waitForOptions: WaitForOptions = {},
+  textMatchOptions?: TextMatchOptions
```
mdjastrzebski, line 67:
```suggestion
  queryOptions?: TextMatchOptions
```
Consider naming textMatchOptions as queryOptions it match RTL docs (please change in other places as well).

website/docs/Queries.md
mdjastrzebski: Do we really have 3rd options of "predicate function" at the moment? RTL does have it, but it seems that we support only string and regex args ATM.

typings/index.d.ts
```diff
@@ -325,6 +335,10 @@ export declare const render: (
 export declare const cleanup: () => void;
 export declare const fireEvent: FireEventAPI;
 
+type TextMatchOptions = {
+  exact: boolean,
```
mdjastrzebski, line 339:
```suggestion
  exact?: boolean,
```

mdjastrzebski (commented):
Overall looks good, I've added some style & RTL-compat comments

typings/index.d.ts
thymikee: This was intentional decision, because switching arguments would be a breaking change. Maybe worth to have this feature in the next major? I doubt many use waitForOptions at all so maybe we could get away with this, but maybe not

website/docs/Queries.md
RafikiTiki: ATM in RNTL there are only `string` and `regex` options implemented

src/helpers/getByAPI.js
RafikiTiki: OK, I'll go with that implementation but in slightly modified manner: we don't have a matcher defined as predicate function option implemented

src/matches.js
```diff
@@ -0,0 +1,67 @@
...
+
+function makeNormalizer({
+  trim,
+  collapseWhitespace,
+  normalizer,
+}: NormalizerInput = {}): NormalizerFn {
+  if (normalizer) {
+    if (
+      typeof trim !== 'undefined' ||
+      typeof collapseWhitespace !== 'undefined'
```
thymikee, line 52: It's cheaper to not typeof 
```suggestion
      trim !== undefined || collapseWhitespace !== undefined
```

website/docs/Queries.md
```diff
@@ -199,6 +203,90 @@ const { getByA11yValue } = render(<Component />);
...
+`exact` option defaults to `true` but if you want to search for a text slice or make text matching case-insensitive you can override it. That being said we advise you to use regex in more complex scenarios.
+
+### Normalization
+
+Before running any matching logic against text, it is automatically normalized. By default, normalization consists of trimming whitespace from the start and end of text, and collapsing multiple adjacent whitespace characters into a single space.
+
+If you want to prevent that normalization, or provide alternative normalization (e.g. to remove Unicode control characters), you can provide a `normalizer` function in the options object. This function will be given a string and is expected to return a normalized version of that string.
+
+:::info
+Specifying a value for `normalizer` replaces the built-in normalization, but you can call `getDefaultNormalizer` to obtain a built-in normalizer, either to adjust that normalization or to call it from your own normalizer.
```
thymikee, line 228: you can't call `getDefaultNormalizer ` because it's not exported from the library entrypoint

typings/index.d.ts
mdjastrzebski: I generally thing that 1) findBy is least used query type of [getBy, queryBy, findBy] and 2) I think that matchOptions are more used than waitForOptions, as some users seem to prefer { exact: false} over /regex/ query.

Overall, I am for breaking change with minor version bump, and docs note. 

As a help for the users, we could emit error/warning if matchOptions contain fields from waitForOptions (timeout, interval) to ease the transition.

thymikee: It's kinda missing feature we should have ship in v7 anyway...

src/__tests__/getByApi.test.js
```diff
@@ -50,3 +50,114 @@ test('supports a regex matcher', () => {
...
+      </View>
+    );
+
+    expect(
+      getByText('A text', {
+        normalizer: normalizerFn,
+      })
+    ).toBeTruthy();
+  });
+});
```
AugustinLF, line 163: I would expect this test to pass. This is would be consistent with the web behaviour. It currently fails in this PR.
```
test('queryAllByText does not match several times the same text', () => {
  const allMatched = render(
    <Text nativeID="1">
      Start
      <Text nativeID="2">This is a long text</Text>
    </Text>
  ).queryAllByText('long text', { exact: false });
  expect(allMatched.length).toBe(1);
  expect(allMatched[0].props.nativeID).toBe('2');
});
```

src/__tests__/queryByApi.test.js
```diff
@@ -87,6 +89,19 @@ test('queryByText with nested Text components each with text return the lowest o
...
+  const NestedTexts = () => (
+    <Text nativeID="1">
+      bob
+      <Text nativeID="2">My text</Text>
+    </Text>
+  );
+
+  const { queryByText } = render(<NestedTexts />);
+
+  expect(queryByText('My text', { exact: false })?.props.nativeID).toBe('1');
```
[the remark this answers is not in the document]
AugustinLF: On web this test returns the element with `nativeId=2`

AugustinLF (changes requested):
I was working on a parallel PR #546, to enable partial matching (somehow I thought that `{exact: false}` was the default behaviour. As suggested by @thymikee I closed my PR but identified two tests that I would expect to behave differently. If you need any help, feel free to ask or look at my implementation, were I got those to pass.

website/docs/Queries.md
mdjastrzebski:
```suggestion

Several APIs accept a `TextMatch` which can be a `string` or `regex`.
```

typings/index.d.ts
mdjastrzebski: I think I've got it, as a compat feature till next major release we would inspect `queryOptions` for `timeout` and `interval` props and use them if they exist, but issue a console warning asking user to migrate their code.

This won't break the older code, but will notify the user about API change. For us it means a relatively small change in next major release (vs alternative of re-ordering the `queryOptions` and `waitForOptions` args for all potential users).

WDYT?

RafikiTiki:
@AugustinLF could you try to run your tests against most current code from this PR? I think I resolved the problem you've mentioned in your review.

src/helpers/getByAPI.js
```diff
@@ -117,9 +117,32 @@ const getNodeByTestId = (node, testID) => {
...
 
 export const getByText = (instance: ReactTestInstance) =>
-  function getByTextFn(text: string | RegExp, options?: TextMatchOptions) {
+  function getByTextFn(
+    text: string | RegExp,
+    options: TextMatchOptions = { exact: true }
+  ) {
     try {
-      return instance.find((node) => getNodeByText(node, text, options));
+      const matches = instance.findAll((node) =>
```
thymikee, line 125: Hm, seems like we could include this logic in `getAllByText` and reuse here? It's what we do with `ByTestId` queries. Otherwise we get different behavior between `getByText` and `getAllByText`.

RafikiTiki: Yup good point, definitely gonna implement this right away

AugustinLF:
@RafikiTiki If you need a hand on this PR, that would be a pleasure to help to get this one in 😄

RafikiTiki:
@AugustinLF Yup, I'd love to get some help on this PR as I started working in a full time project again this week so I won't have time to quickly finish this up.

I've already talked with @thymikee last week about issues we're facing with current implementation and the main problem is the way we are getting text content of nested `<Text />` components (which is different from both RTL's & RNTL's way of getting text content - they're not build it recursively), especially when using `exact: false` option. I believe that's a good starting point to resolve all problems this PR is currently facing.

AugustinLF:
@RafikiTiki I will do so. I will cherry-pick your commits, look about removing the recursive look-up, and see how the tests fare!

janithl:
**BUMP** for this PR, would really help a lot of us stuck in v6.0 to migrate to the latest version.

AugustinLF:
@janithl, yeah I totally dropped the ball on this one, other priorities. I still have that in the back of my mind, but I see that the code changed quite a lot since, so I'll see if I can squeeze that. But if you want to pick it up, don't hesitate!

PR #546, opened 2020-09-14
Enable partial matching

AugustinLF:
### Summary
Follow up of #489, I'm trying to make the migration from `@5.X` to `@7.X` in a fairly large project. Another missed breaking change is that the current implementation doesn't permit "partial matches", as is `render(<Text>Hello World</Text>).getByText('Hello')` will throw. We have a lot of tests relying on this pattern (which is also the way web behaves). In this PR I fixed that.

*This is technically a breaking change* since some people might rely on `getByText` throwing in this case. But I do believe that we should seek for an API as close as possible from `@testing-library/react`.

### Test plan
I added several tests that should cover the different edge cases related to this change.

### DIFF
--- a/src/helpers/getByAPI.js
+++ b/src/helpers/getByAPI.js
@@ -20,12 +20,29 @@ const getNodeByText = (node, text) => {
       const textChildren = getChildrenAsText(node.props.children, Text);
       if (textChildren) {
         const textToTest = textChildren.join('');
-        return typeof text === 'string'
-          ? text === textToTest
-          : text.test(textToTest);
+        if (typeof text === 'string') {
+          if (text === textToTest) {
+            return { exact: true };
+          } else {
+            if (textToTest.includes(text)) {
+              return {
+                exact: false,
+                text,
+              };
+            }
+
+            return null;
+          }
+        } else {
+          return text.test(textToTest)
+            ? {
+                exact: true,
+              }
+            : null;
+        }
       }
     }
-    return false;
+    return null;
   } catch (error) {
     throw createLibraryNotSupportedError(error);
   }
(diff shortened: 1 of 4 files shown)


src/helpers/getByAPI.js
```diff
@@ -20,12 +20,29 @@ const getNodeByText = (node, text) => {
...
       if (textChildren) {
         const textToTest = textChildren.join('');
-        return typeof text === 'string'
-          ? text === textToTest
-          : text.test(textToTest);
+        if (typeof text === 'string') {
+          if (text === textToTest) {
+            return { exact: true };
+          } else {
+            if (textToTest.includes(text)) {
```
AugustinLF, line 27: Using includes here posed a lot of problem since it means that the parent nodes will also match. Hence the _significant_ update to the logic.

src/helpers/getByAPI.js
```diff
@@ -95,13 +112,46 @@ const getNodeByTestId = (node, testID) => {
     : testID.test(node.props.testID);
 };
 
+export const nonErroringGetByText = (instance: ReactTestInstance) =>
+  function nonErroringGetByTextFn(text: string | RegExp) {
+    const matchingInstances = instance.findAll((node) =>
```
AugustinLF, line 117: Since we switch from `find` to `findAll`, the implementation doesn't throw anymore if no match is found. So I updated that so we can still use this implementation in `queryByText` (which shouldn't throw).

thymikee:
How does it relate to #541?

AugustinLF:
I believe that actually #541 is a more correct approach. Now that I'm digging in the textMatch option I realise that some of my assumptions seem to be incorrect. I'm going to close my PR for now and will follow #541. If needed I'll come back to it. Sorry for the trouble.

thymikee:
Don't be sorry! Too bad you didn't see the PR earlier. However, if you feel like it, I'd appreciate reviewing that PR to make it right :)

AugustinLF:
I'm on it, will run my tests against that implementation to see if it behaves as I'd expect it!

Issue #549, opened 2020-09-21
getBy error message should include the value that was not found


sregg:
## Describe the Feature
Currently, when a `getBy` (or `findBy`) fails, the error is only `No instance found`.
It would be great to have what value was not found.
e.g. `getByText('text that doesn't exist')` would throw `No instance found with text 'text that doesn't exist'`

## Possible Implementations
If it's easier to implement you could even omit the type like `No instance found: 'text that doesn't exist'`

## Related Issues
https://github.com/callstack/react-native-testing-library/issues/547
Currently `findBy` doesn't show the stack-trace so it's hard to know which one failed.
Having a better error message would make it easier to debug, even without the stack-trace.

thymikee:
Isn't the code frame enough to determine where the error happens? It should point to the faulty code. If it's not there, it's a bug and should be fixed

sregg:
The stack-trace is not working for `findBy` and `waitFor` at the moment (see #547).
Adding the value in the error message seemed to be an easier thing to prioritize.

mdjastrzebski:
This is implemented, as all queries now report their predicate: e.g. 
`Unable to find an element with text: ${String(text)}`

PR #550, opened 2020-09-21
feat: add the value expected in getBy error messages 

sregg:
Implements https://github.com/callstack/react-native-testing-library/issues/549

### Summary

Add the value that was expected in error message for the following methods:

- `getByText('bla')` => No instances found with text 'bla'
- `getByPlaceholderText('bla')` => No instances found with placeholder text 'bla'
- `getByDisplayValue('bla')` => No instances found with display value 'bla'
- `getByA11yLabel('bla')` => No instances found with accessibilityLabel 'bla' 
- `getByA11yHint('bla')` => No instances found with accessibilityHint 'bla' 
- `getByA11yRole('bla')` => No instances found with accessibilityRole 'bla' 
- `getByA11yStates('bla')` => No instances found with accessibilityStates 'bla' 
- `getByA11yState({ disabled: true})` => No instances found with accessibilityState '{"disabled":true}' 
- `getByA11yValue({ min: 50 })` => No instances found with accessibilityValue  '{"min":50}'

### Test plan

I updated some tests to expect the new error message

### DIFF
--- a/src/__tests__/a11yAPI.test.js
+++ b/src/__tests__/a11yAPI.test.js
@@ -11,7 +11,7 @@ const TEXT_HINT = 'static text';
 const NO_MATCHES_TEXT: any = 'not-existent-element';
 const FOUND_TWO_INSTANCES = 'Expected 1 but found 2 instances';
 
-const fullNoInstanceError = (name: string, value: string = NO_MATCHES_TEXT) => {
+const getNoInstancesFoundMessage = (prop: string, value: string = NO_MATCHES_TEXT) => {
   return `No instances found with ${name} '${value}'`;
 };
 
--- a/src/__tests__/a11yAPI.test.js
+++ b/src/__tests__/a11yAPI.test.js
@@ -9,8 +9,11 @@ const TEXT_LABEL = 'cool text';
 const TEXT_HINT = 'static text';
 // Little hack to make all the methods happy with type
 const NO_MATCHES_TEXT: any = 'not-existent-element';
-const NO_INSTANCES_FOUND = 'No instances found';
 const FOUND_TWO_INSTANCES = 'Expected 1 but found 2 instances';
+
+const fullNoInstanceError = (name: string, value: string = NO_MATCHES_TEXT) => {
+  return `No instances found with ${name} '${value}'`;
+};
 
 const Typography = ({ children, ...rest }: any) => {
   return <Text {...rest}>{children}</Text>;
@@ -73,7 +76,9 @@ test('getByA11yLabel, queryByA11yLabel, findByA11yLabel', async () => {
   const button = queryByA11yLabel(/button/g);
   expect(button && button.props.accessibilityLabel).toEqual(BUTTON_LABEL);
 
-  expect(() => getByA11yLabel(NO_MATCHES_TEXT)).toThrow(NO_INSTANCES_FOUND);
+  expect(() => getByA11yLabel(NO_MATCHES_TEXT)).toThrow(
+    fullNoInstanceError('accessibilityLabel')
+  );
   expect(queryByA11yLabel(NO_MATCHES_TEXT)).toBeNull();
 
   expect(() => getByA11yLabel(TEXT_LABEL)).toThrow(FOUND_TWO_INSTANCES);
@@ -83,7 +88,7 @@ test('getByA11yLabel, queryByA11yLabel, findByA11yLabel', async () => {
   expect(asyncButton.props.accessibilityLabel).toEqual(BUTTON_LABEL);
   await expect(
     findByA11yLabel(NO_MATCHES_TEXT, waitForOptions)
-  ).rejects.toThrow(NO_INSTANCES_FOUND);
+  ).rejects.toThrow(fullNoInstanceError('accessibilityLabel'));
 
   await expect(findByA11yLabel(TEXT_LABEL, waitForOptions)).rejects.toThrow(
     FOUND_TWO_INSTANCES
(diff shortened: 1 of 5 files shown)


src/helpers/getByAPI.js
```diff
@@ -100,7 +100,12 @@ export const getByText = (instance: ReactTestInstance) =>
     try {
       return instance.find((node) => getNodeByText(node, text));
     } catch (error) {
-      throw new ErrorWithStack(prepareErrorMessage(error), getByTextFn);
+      throw new ErrorWithStack(
+        `${prepareErrorMessage(error)} with text '${
+          typeof text === 'string' ? text : text.toString()
+        }'`,
```
mdjastrzebski, line 106:
```suggestion
        `${prepareErrorMessage(error)} with text '${text}'`,
```

We should probably be able to the if condition, sa template string should do the checks & toString() call for us.

src/helpers/getByAPI.js
```diff
@@ -112,20 +117,29 @@ export const getByPlaceholderText = (instance: ReactTestInstance) =>
       );
     } catch (error) {
       throw new ErrorWithStack(
-        prepareErrorMessage(error),
+        `${prepareErrorMessage(error)} with placeholder text '${
+          typeof placeholder === 'string' ? placeholder : placeholder.toString()
+        }'`,
```
mdjastrzebski, line 122:
```suggestion
        `${prepareErrorMessage(error)} with placeholder '${placeholder}'`,
```

We should probably be able to the if condition, sa template string should do the checks & toString() call for us.

src/helpers/getByAPI.js
```diff
@@ -112,20 +117,29 @@ export const getByPlaceholderText = (instance: ReactTestInstance) =>
...
+        getTextInputNodeByDisplayValue(node, displayValue)
       );
     } catch (error) {
-      throw new ErrorWithStack(prepareErrorMessage(error), getByDisplayValueFn);
+      throw new ErrorWithStack(
+        `${prepareErrorMessage(error)} with display value '${
+          typeof displayValue === 'string'
+            ? displayValue
+            : displayValue.toString()
+        }'`,
```
mdjastrzebski, line 140:
```suggestion
        `${prepareErrorMessage(error)} with display value '${displayValue}'`,
```

We should probably be able to the if condition, sa template string should do the checks & toString() call for us.

src/helpers/makeQuery.js
```diff
@@ -37,7 +37,12 @@ const makeQuery = <P: mixed, M: mixed>(
         (node) => isNodeValid(node) && matcherFn(node.props[name], matcher)
       );
     } catch (error) {
-      throw new ErrorWithStack(prepareErrorMessage(error), getBy);
+      throw new ErrorWithStack(
+        `${prepareErrorMessage(error)} with ${name} '${
+          typeof matcher === 'string' ? matcher : JSON.stringify(matcher) || ''
+        }'`,
```
mdjastrzebski, line 43: What is the reason for using `JSON.stringify` instead of `toString` like in other places?

src/__tests__/a11yAPI.test.js
```diff
@@ -9,9 +9,12 @@ const TEXT_LABEL = 'cool text';
 const TEXT_HINT = 'static text';
 // Little hack to make all the methods happy with type
 const NO_MATCHES_TEXT: any = 'not-existent-element';
-const NO_INSTANCES_FOUND = 'No instances found';
 const FOUND_TWO_INSTANCES = 'Expected 1 but found 2 instances';
 
+const fullNoInstanceError = (name: string, value: string = NO_MATCHES_TEXT) => {
```
mdjastrzebski, line 14:
```suggestion
const getNoInstancesFoundMessage = (prop: string, value: string = NO_MATCHES_TEXT) => {
```

mdjastrzebski (dismissed):
Nice work @sregg! 👍 

I've added some suggestions for improving code readability.

src/helpers/getByAPI.js
sregg: That was my first idea but I was getting this Flow error:

image

```
Cannot coerce `text` to string because  `RegExp` [1] should not be coerced.
```

sregg:
```javascript
${text.toString}
```
seems to work
image

Would you be happy with that?

src/helpers/makeQuery.js
sregg: I'm getting this error if I use `toString()`

image

```
Cannot call `matcher.toString` because property `toString` is missing in  mixed
```

src/__tests__/a11yAPI.test.js
```diff
@@ -9,9 +9,12 @@ const TEXT_LABEL = 'cool text';
 const TEXT_HINT = 'static text';
 // Little hack to make all the methods happy with type
 const NO_MATCHES_TEXT: any = 'not-existent-element';
-const NO_INSTANCES_FOUND = 'No instances found';
 const FOUND_TWO_INSTANCES = 'Expected 1 but found 2 instances';
 
+const getNoInstancesFoundMessage = (prop: string, value: string = NO_MATCHES_TEXT) => {
+  return `No instances found with ${name} '${value}'`;
```
thymikee, line 15: Let's use double quote:
```suggestion
  return `No instances found with ${name} "${value}"`;
```

src/helpers/getByAPI.js
thymikee: @sregg doing `text.toString()` is the way to go here, yea 👍

src/helpers/makeQuery.js
```diff
@@ -47,7 +52,12 @@ const makeQuery = <P: mixed, M: mixed>(
     );
 
     if (results.length === 0) {
-      throw new ErrorWithStack('No instances found', getAllBy);
+      throw new ErrorWithStack(
+        `No instances found with ${name} '${
+          typeof matcher === 'string' ? matcher : JSON.stringify(matcher) || ''
```
thymikee, line 57: we have `pretty-format` in our deps, so let's use that with `{min: true}` option set: https://github.com/facebook/jest/tree/master/packages/pretty-format#usage-with-options

src/helpers/getByAPI.js
thymikee: Btw, I think that `prepareErrorMessage` function is a perfect place for handling this extra text. Similar to what you have in a test here: https://github.com/callstack/react-native-testing-library/pull/550/files#diff-4246008ab708750218ff9e41118401a1R14

sregg:
@thymikee I addressed your comments in my latest commit

thymikee (approved):
👍

PR #554, opened 2020-09-28
feat: add text match options a.k.a string precision API

### LINKED ISSUE #514: The string precision API is missing in v7

AugustinLF:
Continuation of @RafikiTiki's work of #541

### Summary

This does two things, adding some options to the queries (see initial PR), and remove matching strings spread in several levels of text. Namely the following example will not be valid anymore.

```
render(
      <Text>
        Hello <CustomText>World!</CustomText>	
      </Text>	
    ).queryByText('Hello World!')	
  ).toBeTruthy();
```

This is a breaking change, but it aligns the APIs and behaviour with `@testing-library/react`.

Closes #541 
Closes #514


AugustinLF:
@thymikee is there anything I can do to help this PR to move forward?

src/__tests__/queryByApi.test.js
```diff
@@ -48,7 +48,7 @@ test('queryByText not found', () => {
   ).toBeFalsy();
 });
 
-test('queryByText nested text across multiple <Text> in <Text>', () => {
+test('queryByText does not match nested text across multiple <Text> in <Text>', () => {
```
thymikee, line 51: So we're regressing this functionality to be compatible with Testing Library, which also doesn't support this?

src/__tests__/queryByApi.test.js
```diff
@@ -72,6 +72,7 @@ test('queryByText with nested Text components return the closest Text', () => {
   const { queryByText } = render(<NestedTexts />);
 
   expect(queryByText('My text')?.props.nativeID).toBe('2');
+  expect(queryByText('My text', { exact: false })?.props.nativeID).toBe('2');
```
thymikee, line 75: but it's supported with `exact: false`, I think that's fine

thymikee (approved):
Before merging, I need to make sure that @RafikiTiki gets attribution for his work. Since there are 2 breaking changes here now, I'm considering releasing it in a major update

src/__tests__/queryByApi.test.js
AugustinLF: Yes. I checked, testing library doesn't match in this case.

src/__tests__/queryByApi.test.js
AugustinLF: I'm not sure what you're talking about here. exact doesn't have anything to do with how we match text spread across multiple node (if this comment is linked to https://github.com/callstack/react-native-testing-library/pull/554/files/4087aff6f3cc2200762308911929a3de295902b4#r500085148).

AugustinLF:
Yes, I think this should be a major release. I only see the recursive matching as breaking (don't know which other one you're talking about) though. I can help with the release notes if needed.

src/__tests__/queryByApi.test.js
thymikee: Ugh, I read it the wrong way 🙈

src/__tests__/findByApi.test.js
```diff
@@ -18,19 +18,19 @@ test('findBy queries work asynchronously', async () => {
   } = render(<View />);
   await expect(findByTestId('aTestId', options)).rejects.toBeTruthy();
   await expect(findAllByTestId('aTestId', options)).rejects.toBeTruthy();
-  await expect(findByText('Some Text', options)).rejects.toBeTruthy();
-  await expect(findAllByText('Some Text', options)).rejects.toBeTruthy();
+  await expect(findByText('Some Text', {}, options)).rejects.toBeTruthy();
+  await expect(findAllByText('Some Text', {}, options)).rejects.toBeTruthy();
   await expect(
-    findByPlaceholderText('Placeholder Text', options)
+    findByPlaceholderText('Placeholder Text', {}, options)
```
thymikee, line 24: this is the second breaking change I mentioned

src/helpers/findByAPI.js
```diff
@@ -12,54 +12,92 @@ import {
...
   testId: string | RegExp,
   waitForOptions: WaitForOptions = {}
-) => makeFindQuery(instance, getByTestId, testId, waitForOptions);
+) => makeFindQuery(instance, getByTestId, testId, {}, waitForOptions);
 
 export const findAllByTestId = (instance: ReactTestInstance) => (
   testId: string | RegExp,
   waitForOptions: WaitForOptions = {}
-) => makeFindQuery(instance, getAllByTestId, testId, waitForOptions);
+) => makeFindQuery(instance, getAllByTestId, testId, {}, waitForOptions);
```
thymikee, line 36: ByTestId also accepts `TextMatch` https://testing-library.com/docs/dom-testing-library/api-queries#bytestid, can you add it?

AugustinLF: Done!

thymikee:
FYI, there's an ongoing discussion in DOM Testing Library regarding matching nested text: https://github.com/testing-library/dom-testing-library/issues/473#issuecomment-598432817. Also, seems like `getByLabelText` already behaves similar to what we have today. 

Hope you understand why I'm so hesitant merging this. We're likely regressing a use case that we'd eventually had to add anyway. I'd like to think of a way to not break matching nested nodes using `exact: false`

src/__tests__/findByApi.test.js
mdjastrzebski: We could provide an easy migration path by looking for `waitFor` options (`timeout`, `interval`) also in the `queryOptions` only for `findBy` queries and only till next major release. While displaying a console warning to inform the developer. Wdyt?

src/__tests__/getByApi.test.js
```diff
@@ -50,3 +50,125 @@ test('supports a regex matcher', () => {
...
+    const { getByText } = render(
+      <View>
+        <Text>{`  Text     and
+        
+        
+        whitespace`}</Text>
+      </View>
+    );
+
+    expect(getByText('Text and whitespace')).toBeTruthy();
```
mdjastrzebski, line 134: I like that one 👍

src/__tests__/queryByApi.test.js
```diff
@@ -87,18 +88,26 @@ test('queryByText with nested Text components each with text return the lowest o
...
+  expect(queryByTextFirstCase('My text')).toBe(null);
   expect(
-    render(
-      <Text>
-        Hello <CustomText>World!</CustomText>
-      </Text>
-    ).queryByText('Hello World!')
-  ).toBeTruthy();
+    queryByTextSecondCase('My text', { exact: false })?.props.nativeID
+  ).toBe('2');
```
mdjastrzebski, line 110: Does seem strange a bit, that we use one component structure with the default `exact: true` and other component structure with `exact: false`. If we were to contast exact true/false IMO we should use the same component structure.

Looks like we should do one of the following (in order of preference):
1. merge queryByTextFirstCase + queryByTextSecondCase
1. add queryByTextFirstCase + exact: false & queryByTextSecondCase + exact: true
1. split the test into two unreleated tests

src/__tests__/queryByApi.test.js
```diff
@@ -112,5 +121,28 @@ test('queryByText nested deep <CustomText> in <Text>', () => {
         <CustomText>Hello</CustomText> <CustomText>World!</CustomText>
       </Text>
     ).queryByText('Hello World!')
-  ).toBeTruthy();
+  ).toBe(null);
+});
+
+test('queryAllByText does not match several times the same text', () => {
```
mdjastrzebski, line 127: 👍

src/matches.js
```diff
@@ -0,0 +1,51 @@
+// @flow
+export type NormalizerFn = (textToNormalize: string) => string;
+
+function matches(
+  textToMatch: string,
+  matcher: string | RegExp,
+  normalizer: NormalizerFn,
+  exact?: boolean = true
+) {
```
mdjastrzebski, line 9:
```suggestion
function matches(
  matcher: string | RegExp,
  text: string,
  normalizer: NormalizerFn,
  exact?: boolean = true
) {
```

readability, nit: more natural order would be "matcher matches text", hence the suggested argument order. 

Also textToMatch felt confusing as I thought that this is the text we are looking for. Suggested rename: `text` or `textToMatchAgainst`, with `text` being simplier.

src/helpers/getByAPI.js
```diff
@@ -20,9 +27,11 @@ const getNodeByText = (node, text) => {
       const textChildren = getChildrenAsText(node.props.children, Text);
       if (textChildren) {
         const textToTest = textChildren.join('');
-        return typeof text === 'string'
-          ? text === textToTest
-          : text.test(textToTest);
+        const { exact, normalizer } = options;
+        const normalizerFn = makeNormalizer({
+          normalizer,
+        });
```
mdjastrzebski, line 33: Looks like const normalizeFn = `normalizer ?? getDefaultNormalizer()` would be more straightforward.

src/helpers/getByAPI.js
```diff
@@ -52,63 +62,93 @@ const getChildrenAsText = (children, TextComponent, textContent = []) => {
...
 
 export const getByText = (instance: ReactTestInstance) =>
-  function getByTextFn(text: string | RegExp) {
+  function getByTextFn(
+    text: string | RegExp,
+    options: TextMatchOptions = { exact: true }
+  ) {
     try {
-      return instance.find((node) => getNodeByText(node, text));
+      const results = getAllByText(instance)(text, options);
```
mdjastrzebski, line 128: 👍

typings/index.d.ts
```diff
@@ -126,32 +132,38 @@ interface QueryByAPI {
...
-    waitForOptions?: WaitForOptions
+    value: TextMatch,
+    waitForOptions?: WaitForOptions,
+    options?: TextMatchOptions,
   ) => FindAllReturn;
   findAllByTestId: (
-    testID: string | RegExp,
+    testID: TextMatch,
     waitForOptions?: WaitForOptions
   ) => FindAllReturn;
```
mdjastrzebski, line 168: Looks like typescript types for findBy have incorrect options args order.

typings/index.d.ts
```diff
@@ -350,6 +372,8 @@ export declare const getQueriesForElement: (
   instance: ReactTestInstance
 ) => Queries;
 
+export declare const getDefaultNormalizer: (normalizerConfig: NormalizerConfig) => NormalizerFn;
```
mdjastrzebski, line 375:
```suggestion
export declare const getDefaultNormalizer: (normalizerConfig?: NormalizerConfig) => NormalizerFn;
```

mdjastrzebski (commented):
Wow a lot of good code! 👍 

I've added some comments as there seems to be some minor errors (e.g. typescript types) and opportunities to improve readability/code structure.

typings/index.d.ts
AugustinLF: Oops forgot to update, thanks!

RafikiTiki:
@AugustinLF I'll be grateful if you could fix issues pointed by @mdjastrzebski as I'm still like 120% engaged in another project

src/__tests__/queryByApi.test.js
AugustinLF: I do overall believe that tests should be restructured, a lot of cases are tested only because the internal APIs are the same (a lot of edge cases for `findByText` are only tested in `getByText` for instance), so I'd like to have more tests, and make them more verbose. But I believe it's out of the scope of this PR.

But this specific case makes sense, I'll follow your suggestion, I got a bit lazy and monkey patched some tests without really reconsidering them.

AugustinLF:
@thymikee I do understand your concerns. It sadly seems that this issue has not made much progress since March. I do however perfectly understand your worries about double breaking changes.

Firstly I'd like to reiterate the reason why I see this breaking change necessary: where I work we run platform agnostic tests, i.e. the same tests are ran both on web with `@testing-library/react` and native with `@testing-library/react-native`. That means that differences in API are tricky but can be handled (we have a wrapper so we can some form of mapping, for instance between `onPress` and `onClick`). But differences in behaviour (what is matched) are a blocker. Currently the _old_ implementation of native means that we can run the same code for both platform. We are currently prevented from upgrading.

But I see your point regarding the `exact: false` behaviour. I could implement that (which is closer from what was my previous implementation in #546), but I'd like to be sure that we're all on the same page and want to move forward with this implementation so no one wastes more time.

### Proposal

1. `getByText('myText', {exact: true})` (default case) will not match text across several nodes
2. `getByText('My text', {exact: false})` will match text across several nodes `<Text>My <Text>text<Text/></Text>`

In the current implementation the first point is true, the second false, but that corresponds to the current `@testing-library/react` implementation. You suggest to do the second point, which is doable but would make both implementation diverge. If you confirm that's what you want and that you feel is the best for the project, I'm fine with that, because it's still an improvement compared to the implementation of the library.

michaelduminy:
I'm trying to upgrade to RN 0.63 and include this library but due to the difference in APIs between this and `@testing-library/react` I'm having trouble with shared code, so really rooting for this PR to land. Please let me know if I can do anything :)

src/__tests__/queryByApi.test.js
michaelduminy: How does this PR change this functionality? Finding the text in nodes still recurses through children

AugustinLF: Yes https://github.com/callstack/react-native-testing-library/pull/554/files#diff-28f6e2b77eef7ef8b7583339cfdf593fR43

src/matches.js
```diff
@@ -0,0 +1,51 @@
...
+
+type NormalizerConfig = {
+  trim?: boolean,
+  collapseWhitespace?: boolean,
+};
+
+function getDefaultNormalizer({
+  trim = true,
+  collapseWhitespace = true,
+}: NormalizerConfig = {}): NormalizerFn {
```
michaelduminy, line 32: **thought** if `collapseWhitespace` is true then the value of `trim` can be ignored. Maybe the logic should more clearly reflect this:

```javascript 
if (collapseWhitespace) return text.replace(/\s+/g, ' ')
if (trim) return text.trim()

return text
```

src/__tests__/queryByApi.test.js
michaelduminy: I must be missing something but **how** does the logic added to `getNodeByText` end up not matching in this case? As I see it the the logic in `getChildrenAsText ` will return `['Hello', '  ', 'World', '!']` which when joined will be `"Hello World!"`, which is an exact match for what you are looking for.

michaelduminy:
**suggestion** maybe we can add to this PR (or after it lands) the ability to not search child text nodes recursively, sometimes you don't want to match across Text elements

src/__tests__/queryByApi.test.js
AugustinLF: Yes, the change is to keep matching component with multiple children, but not `<Text>Hello <Text>World</Text></Text`, because now `getChildrenAsText` don't read its children's children

michaelduminy: Ah, so if the test render looked like this:
```javascript
<Text>
  <Text>Hello</Text>
  World!
</Text>
```
Would this then match the outer text?

AugustinLF: This currently match is the current released version, but wouldn't match in my current implementation. It's the topic of discussion in https://github.com/callstack/react-native-testing-library/pull/554#issuecomment-705437898

src/matches.js
mdjastrzebski: If I understand correctly collapeseWhitespace does not imply trim. Am I missing something?

E.g. for input: `"     a    a     "`
- collapseWhitespace: `" a a "`
- trim only: `"a    a"`
- collapseWhitespace + trim: `"a a"`

src/__tests__/queryByApi.test.js
dsznajder, line 110: Those could contain additional `describe` there as it contains 2 different scenarios.

src/__tests__/findByApi.test.js
AugustinLF: That's a good idea that I missed, I fixed it in the latest commit d2f9f337faa45f945c82bcaf41fc8145e00403b6

AugustinLF:
Hey folks, it's never too late to pick up a stale PR! I've addressed one of the comments to reduce the breaking changes, but there's still the main question, which is the issue with matching across multiple nodes and the difference in the API with `@testing-library/react`. @thymikee you were unsure about this change and raised some concerns. I've written higher in this PR what my opinion is (i.e. same behaviour on both platform), but also discussed some other alternatives, what do you think? Do you see a solution to move forward?

thymikee:
Hey, I'm gonna look at this once again soon. We already release v8, so we're clear to introduce other unrelated breaking changes 💪🏼

AugustinLF:
Hey hey @thymikee, anything I can do to help? I admit that current status is a bit messy since the PR went over several months and building on top of another contribution. If that helps, I could try to make a recap of the whole discussion/status.

thymikee:
@AugustinLF that won't hurt for sure! I'm pretty sure this is the way we want to move forward. It would help to gather all the breaking changes this can cause and ways to migrate. 

I'm a bit swamped lately with work and personal stuff and having a really hard time to fit this into my pipeline. But maybe I could only have a brief look and release this as an alpha version for testing. Appreciate you stick with us all this time 🙌🏼

AugustinLF:
This PR includes two main changes: it adds `TextMatch` options, and changes the way nested text is matched:

Support to [TextMatch](https://testing-library.com/docs/queries/about#textmatch) in all text based queries. This moves the current `waitForOptions` parameter of the `findBy*` queries from the 2nd place to the 3rd place. For backward compatibility, we still read it from the 2nd parameter but log a deprecation warning.

Nested text matching changes, it will not match text broken across several nodes:
```tsx
// This examples matches before this PR and will not after
render(
      <Text>
        Hello <CustomText>World!</CustomText>	
      </Text>	
    ).getByText('Hello World!')	
```

Extra care has been put to make sure that the relevant node gets matched, when several nodes could match (in the previous example, when matching `"World"`, it'll match `CustomText` and not `Text`.

These two changes have the benefit of making the API closer to the older implementation of `testing-library/react-native` (before being transferred to callstack), and to make it closer to the web implementation.

thymikee:
@AugustinLF I've pushed a commit with a few updates, please check it. One thing that we're still missing is matching a11y queries, but I'm thinking about introducing it in a separate PR

AugustinLF:
Great, that looks good. I've added a commit to restore a mock, a TODO I forgot in a test.

The a11y queries are very high on my list (to be honest we've reimplemented them in user-land on our side). But I'd love to take a step back and try to get a more holistic view and figure out what we're missing, if anything needs to be improved, I'm planning to try to get some feedback from our devs here at Klarna. I hope to be able to free more time to improve our test infra and `@testing-library/react-native` is a very important part of it, so looking forward helping.

Any plan to switch to TS? 👀

thymikee:
> The a11y queries are very high on my list (to be honest we've reimplemented them in user-land on our side). But I'd love to take a step back and try to get a more holistic view and figure out what we're missing

Yes, let's take some time to make it right. We may even want to revisit the API and maybe somehow mimic the `getByRole` from the web, as accessibility matures on RN? Not sure if that's even viable, so treating it as an idea / open question for now.

> Any plan to switch to TS? 👀

Yes please!

thymikee:
Published a pre-release: https://github.com/callstack/react-native-testing-library/releases/tag/v9.0.0-alpha.0. Thank you once again @AugustinLF, much appreciated!

janpe:
@thymikee any estimate on when 9.0.0 could be released?

AugustinLF:
We need to publish some announcement around those change, I hope to pick that work today or tomorrow https://github.com/callstack/react-native-testing-library/discussions/862#discussioncomment-1705927

PR #557, opened 2020-09-30
chore: Update CircleCI to run examples

cross19xx:
### Summary
CircleCI configuration runs all examples currently in the example directory

### Test plan
Currently, the implementation runs only in master


.circleci/config.yml
```diff
@@ -65,6 +65,15 @@ jobs:
...
           cd website && yarn install && yarn build
+  test-examples:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: ~/react-native-testing-library
+      - run: |
+          cd examples/reactnavigation && yarn install && yarn test
+      - run: |
+          cd examples/redux && yarn install && yarn test
```
thymikee, line 76: can we have a "for-each" that would go through all examples?

cross19xx: `ls -d ./examples/* | xargs -I {} bash -c "cd '{}' && npm install && npm run test"`

This lists all directory inside the examples directory and for each of them, installs dependencies and runs a test

.circleci/config.yml
```diff
@@ -65,6 +65,13 @@ jobs:
           at: ~/react-native-testing-library
       - run: |
           cd website && yarn install && yarn build
+  test-examples:
+    <<: *defaults
+    steps:
+      - attach_workspace:
+          at: ~/react-native-testing-library
+      - run: |
+          ls -d ./examples/* | xargs -I {} bash -c "cd '{}' && yarn install && npm run test"
```
thymikee, line 74:
```suggestion
          ls -d ./examples/* | xargs -I {} bash -c "cd '{}' && yarn install && yarn test"
```

thymikee (approved):

Issue #566, opened 2020-10-01
A way to get at top level rendered element


dephiros:
## Describe the Feature
At my company, we have a few uses case where we want to make assertion on the top level rendered element. Pre v7, we were able to do `const element = view.container.children?.[0] || null`. Is there a way to get the same result for post v7 without adding additional `testID`?

Trivial use cases example:
```
const Component = (render=true) => (render ? <View></View> : null);

// assert that component render when render is true
// assert that component render nothing when render is false
```

## Possible Implementations
I see that we have [renderer.root here](https://github.com/callstack/react-native-testing-library/blob/master/src/render.js#L37). Maybe just have to expose that element

## Related Issues
https://github.com/callstack/react-native-testing-library/issues/477#issuecomment-699096451

thymikee:
> we want to make assertion on the top level rendered element

What kind of assertions? Do they test something meaningful? You can always add a testID to this component if you really need access to that instance. But from the user perspective – does it make sense? Maybe this test can be deleted? It's worth to ask such questions, as there may be better ways to test. If there aren't I'm open to discuss a specific use case, explained in detail, so we can figure something out.

> const element = view.container.children?.[0] || null

This code is testing implementation details unfortunately. How would the user know about children or parent elements? They don't. In the spirit of Testing Library, we shouldn't allow to easily grab it.

https://github.com/romgain/react-select-event/blob/master/src/index.ts#L13 is a counter-example of how one should test. It's not much different than using `dive()` from Enzyme. It makes tests fragile. Even though it's defined in one place, there may be multiple places that rely on it, and thus could break.

> Basic example of custom query is instead of having in multiple tests getByTestID('my-component') you can create your custom queries: getMyComponent, getAllMyComponents etc. (which can be generated).

Like this? 
```js
function getMyComponent(component) {
  const {getByTestId} = render(component);
  return getByTestId("my-component");
}
```
I'm not sure why access to the container would be necessary in such case. Would love to have more details and see if we really can't solve it with existing tools.

> Another example would be when we want to narrow the the place where we are looking for

That's what `within` helper is doing. Have you tried it?

klekowskim:
In general you have right. I'll try different arguments:
1. Why we should not export root? This returns the same object as every query, and I can use test data id on a wrapper query, and using getByTestId I can have my root component. I'm saying that there is a workaround, but this should not be necessary.
[...]

PR #567, opened 2020-10-06
feat: export `container` from render

klekowskim:
### Summary

Exports `root` test instance from render method. It is a change mentioned in this feature request: #566

After this change it is possible to get test instance from render method:
```
const { container } = render(<View />);
```

### DIFF
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -62,9 +62,7 @@ describe('fireEvent', () => {
   });
 
   test('should throw an Error when event handler was not found', () => {
-    const { getByText } = render(
-      <WithoutEventComponent onPress={() => 'this is not passed to children'} />
-    );
+    const { getByText } = render(<WithoutEventComponent />);
 
     expect(() => fireEvent(getByText('Without event'), 'press')).toThrow(
       'No handler function found for event: "press"'
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -62,7 +62,9 @@ describe('fireEvent', () => {
   });
 
   test('should throw an Error when event handler was not found', () => {
-    const { getByText } = render(<WithoutEventComponent />);
+    const { getByText } = render(
+      <WithoutEventComponent onPress={() => 'this is not passed to children'} />
+    );
 
     expect(() => fireEvent(getByText('Without event'), 'press')).toThrow(
       'No handler function found for event: "press"'
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -382,3 +382,28 @@ test('renders options.wrapper around updated node', () => {
     </RCTSafeAreaView>
   `);
 });
+
+it('returns root test instance', () => {
+  // when
+  const renderResult = render(<View testID="inner" />);
+
+  // then
+  expect(renderResult.root).toBeDefined();
+  expect(renderResult.root.type).toBe(View);
+});
+
+it('returns wrapped component as root test instance', () => {
+  // given
+  const WrapperComponent = ({ children }) => (
+    <SafeAreaView testID="wrapper">{children}</SafeAreaView>
+  );
+
+  // when
+  const renderResult = render(<View testID="inner" />, {
+    wrapper: WrapperComponent,
+  });
+
+  // then
+  expect(renderResult.root).toBeDefined();
+  expect(renderResult.root.type).toBe(WrapperComponent);
+});
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -388,6 +388,7 @@ test('returns container', () => {
 
   expect(container).toBeDefined();
   expect(container.type).toBe(View);
+  expect(container.props.testID).toBe('inner');
 });
 
 test('returns wrapped component as container', () => {
@@ -401,4 +402,5 @@ test('returns wrapped component as container', () => {
 
   expect(container).toBeDefined();
   expect(container.type).toBe(WrapperComponent);
+  expect(container.props.testID).not.toBeDefined();
 });
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -387,6 +387,8 @@ test('returns container', () => {
   const { container } = render(<View testID="inner" />);
 
   expect(container).toBeDefined();
+  // `View` composite component is returned. This behavior will break if we
+  // start returning only host components.
   expect(container.type).toBe(View);
   expect(container.props.testID).toBe('inner');
 });
@@ -401,6 +403,8 @@ test('returns wrapped component as container', () => {
   });
 
   expect(container).toBeDefined();
+  // `WrapperComponent` composite component is returned with no testID passed to
+  // it. This behavior will break if we start returning only host components.
   expect(container.type).toBe(WrapperComponent);
   expect(container.props.testID).not.toBeDefined();
 });
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -383,16 +383,16 @@ test('renders options.wrapper around updated node', () => {
   `);
 });
 
-it('returns root test instance', () => {
+test('returns container', () => {
   // when
   const renderResult = render(<View testID="inner" />);
 
   // then
-  expect(renderResult.root).toBeDefined();
-  expect(renderResult.root.type).toBe(View);
+  expect(renderResult.container).toBeDefined();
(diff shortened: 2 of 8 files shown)


klekowskim:
I can also change it to export `container`, instead of the `root`. To do that probably I would need to add internal wrapper, as it was in the second library [here](https://github.com/testing-library/native-testing-library/blob/c864cd573cd3ac9f40da8540d6ffbc6b8dc6aa52/src/index.js#L26).

In that case the API would be more like in the react-testing-library.

michaelduminy:
+1 to this change, and +1 on adding `container` and `baseElement`

mdjastrzebski:
I guess I am in favor of exposing `container` instead of `root`. Getting root element from contains is very easy and we have maintain api between RTL and RNTL with all benefits of such (code and knowledge sharing, etc).

Regarding wrapping of root element, I think that wrapping it in plain `View` should do the cases [TL/RN v6](https://github.com/testing-library/native-testing-library/blob/c864cd573cd3ac9f40da8540d6ffbc6b8dc6aa52/src/index.js#L26) did wrap it in `AppContainer` from `''react-native/Libraries/ReactNative/AppContainer'` but reading the code of `AppContainer` this seems not useful as it deals with low-level RN stuff like profiling, logbox, inspector, etc.

klekowskim:
@thymikee What do you think? I can adjust it but I'd like to know if it is going to be merged 😄

thymikee:
Let's do it. Like @mdjastrzebski said

src/__tests__/fireEvent.test.js
```diff
@@ -62,9 +62,7 @@ describe('fireEvent', () => {
   });
 
   test('should throw an Error when event handler was not found', () => {
-    const { getByText } = render(
-      <WithoutEventComponent onPress={() => 'this is not passed to children'} />
```
thymikee, line 66: please bring this back

src/__tests__/render.test.js
```diff
@@ -378,3 +378,28 @@ test('renders options.wrapper around updated node', () => {
     </RCTSafeAreaView>
   `);
 });
+
+test('returns container', () => {
+  // when
```
thymikee, line 383: can we get rid of these comments?

src/render.js
```diff
@@ -1,5 +1,7 @@
 // @flow
 import * as React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { View } from 'react-native';
```
thymikee, line 4: I guess it's time to make "react-native" a peer dependency. Making it ">= 0.59" should be fine

src/__tests__/fireEvent.test.js
klekowskim: I was writing the massage for you but you were much faster 😄 

I cannot. The test in that form is wrong or the code has a bug. I believe this cannot be fixed in the code so I've changed the test. The fire event is looking for onPress prop and it was finding that.

After adding wrapper (View in the render method) the test was failing. Previously it worked, because there is a weird "if" [here](https://github.com/callstack/react-native-testing-library/blob/85bccf2f6a6e36d5bf5ef73dcb3c1e6552f84649/src/fireEvent.js#L31) which handles this case.

This test is also failing when wrapping it directly in test, e.g:
```
render(
<View>
      <WithoutEventComponent onPress={() => 'this is not passed to children'} />
</View>
)
```

Previously the test was suppose to check that we throw an error when event handler was not found. But looking at the test case it was there. Do you agree?

src/render.js
klekowskim: done

src/__tests__/render.test.js
klekowskim: done

src/__tests__/fireEvent.test.js
thymikee: Seems like it wouldn't really work with `wrapper` option as well. We'd need to check for the wrapper to make it work:
```diff
diff --git a/src/fireEvent.js b/src/fireEvent.js
index 06fa0d3..859b118 100644
--- a/src/fireEvent.js
+++ b/src/fireEvent.js
@@ -27,7 +27,12 @@ const findEventHandler = (
   if (handler && isEventEnabled) return handler;

   // Do not bubble event to the root element
-  if (element.parent === null || element.parent.parent === null) {
+  if (
+    element.parent === null ||
+    element.parent.parent === null ||
+    element.parent.props.nativeID === 'rntl_wrapper' ||
+    element.parent.parent.props.nativeID === 'rntl_wrapper'
+  ) {
     if (hasHandler) {
       return null;
     } else {
diff --git a/src/render.js b/src/render.js
index c304e21..eb499c1 100644
--- a/src/render.js
+++ b/src/render.js
@@ -28,7 +28,11 @@ export default function render<T>(
   { wrapper: Wrapper, createNodeMock }: Options = {}
 ) {
   const wrap = (innerElement: React.Element<any>) =>
-    Wrapper ? <Wrapper>{innerElement}</Wrapper> : <View>{innerElement}</View>;
+    Wrapper ? (
+      <Wrapper>{innerElement}</Wrapper>
+    ) : (
+      <View nativeID="rntl_wrapper">{innerElement}</View>
+    );
```

One thing I didn't realize would be breaking snapshots when adding this wrapper.

thymikee: Maybe we don't need to throw this error at all? Seems like dom testing library doesn't do it https://github.com/testing-library/dom-testing-library/blob/master/src/events.js

klekowskim: Maybe not. But still this [if](https://github.com/callstack/react-native-testing-library/blob/85bccf2f6a6e36d5bf5ef73dcb3c1e6552f84649/src/fireEvent.js#L31) will not work, when somebody adds theirs own wrapper.

thymikee: Yea, I'm aware. Nobody complained so far

thymikee (approved):
Removed the default wrapper, because it's orthogonal to returning the root instance and would require us to release this feature in a major release.
I left the peer dep on react-native, because it should be there anyway.

Thanks for the contribution!

src/__tests__/fireEvent.test.js
thymikee: Removed the wrapper for now, we can add it separately along with `baseElement`

pvinis (approved):

src/__tests__/render.test.js
```diff
@@ -378,3 +378,23 @@ test('renders options.wrapper around updated node', () => {
     </RCTSafeAreaView>
   `);
 });
+
+test('returns container', () => {
+  const { container } = render(<View testID="inner" />);
+
+  expect(container).toBeDefined();
+  expect(container.type).toBe(View);
```
mdjastrzebski, line 386: Consider adding nativeID/testID check

src/__tests__/render.test.js
```diff
@@ -378,3 +378,23 @@ test('renders options.wrapper around updated node', () => {
...
+  const WrapperComponent = ({ children }) => (
+    <SafeAreaView testID="wrapper">{children}</SafeAreaView>
+  );
+
+  const { container } = render(<View testID="inner" />, {
+    wrapper: WrapperComponent,
+  });
+
+  expect(container).toBeDefined();
+  expect(container.type).toBe(WrapperComponent);
```
mdjastrzebski, line 399: Consider adding testID check. Currently there is no reason for testID props to be present.

mdjastrzebski (approved):
Overall looks good, consider adding testID/nativeID checks in tests

src/__tests__/render.test.js
```diff
@@ -382,3 +382,25 @@ test('renders options.wrapper around updated node', () => {
...
+  expect(container.props.testID).not.toBeDefined();
```
mdjastrzebski, line 405: I though that it should be `"wrapper"` here.

thymikee: We're not currently normalizing container to return a host component, so the actual instance of `WrapperComponent` is returned, hence no testID

pvinis:
I just tried to use this and saw a couple of issues:
- `typings/index.d.ts` is not updated to contain the `container`.
- `ReactTestInstance` is not exported in `typings/index.d.ts` so I can't add this as a type to my helper test function
- I need access to `ErrorWithStack` and `getNodeByText`. Is there a way to use these on `container`?

pvinis:
@thymikee are these 👆 done or being worked on, or would it be better if I make an issue for these?

thymikee:
I don't think these are addressed. Please create an issue :)

PR #568, opened 2020-10-08
fix(breaking): use real timers internally to fix awaiting with fake timers

### LINKED ISSUE #506: Using Jest mock timers and waitFor together causes tests to timeout
## Describe the bug

Related to #391. Using `jest.useFakeTimers()` in combination with `waitFor`, causes the tests using `waitFor` to fail due to timeout error:

>Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.

Relying upon `jest.useFakeTimers("modern")` instead causes the above failure for _all tests_ if the file merely imports `waitFor` at all, regardless if the given test uses `waitFor` or not.

## Expected behavior

All tests in the [reproduction test case](https://github.com/HealthTeacher/rntl-jest-timer-case) should pass.

## Steps to Reproduce

1. Clone the [reproduction test case](https://github.com/HealthTeacher/rntl-jest-timer-case).
1. `yarn install`
1. `yarn test`

## Versions

```
npmPackages:
    @testing-library/react-native: ^7.0.1 => 7.0.1 
    react: ^16.13.1 => 16.13.1 
    react-native: ^0.63.2 => 0.63.2 
    react-test-renderer: ^16.13.1 => 16.13.1
```

michaelduminy:
### Summary

If you use Timer Mocks in your tests and `waitFor` then you cannot `await` this as the timers internal to `waitFor` would be replaced by the mock timers and thus never resolve. This PR borrows techniques from [wait-for-expect](https://github.com/TheBrainFamily/wait-for-expect/blob/master/src/helpers.ts#L6-L30) and [@testing-library/dom](https://github.com/testing-library/dom-testing-library/blob/master/src/helpers.js#L1-L65) to detect if fake timers are in-use, switch back to real timers for a specific call, then re-instate the fake timers using the timer implementation that was present before.

This implementation also ensures that any mocks of `Date` or `Date.now` will not affect `waitFor`.

Outstanding work: update the docs if this lands.

fixes #506 

### Test plan

Added multiple tests to ensure this works. Unfortunately there are [problems](https://github.com/facebook/react-native/issues/29303) when awaiting a promise that uses 'modern' timers (the default in jest 27) and the react-native preset polyfilling Promise.

NOTE: `jest.useFakeTimers('modern')` and awaiting a promise in the context of react-native will timeout

There is a way to fix this, but it requires patching the `@jest/fake-timers` library as described [here](https://github.com/facebook/jest/issues/10221#issuecomment-654687396).

### DIFF
--- a/jest/preset.js
+++ b/jest/preset.js
@@ -1,9 +1,10 @@
 const reactNativePreset = require('react-native/jest-preset');
 
-module.exports = Object.assign({}, reactNativePreset, {
+module.exports = {
+  ...reactNativePreset,
   // this is needed to make modern fake timers work
   // because the react-native preset overrides global.Promise
   setupFiles: [require.resolve('./save-promise.js')]
     .concat(reactNativePreset.setupFiles)
     .concat([require.resolve('./restore-promise.js')]),
-});
+};
--- a/jest/preset.js
+++ b/jest/preset.js
@@ -1,0 +1,9 @@
+const reactNativePreset = require('react-native/jest-preset');
+
+module.exports = Object.assign({}, reactNativePreset, {
+  // this is needed to make modern fake timers work
+  // because the react-native preset overrides global.Promise
+  setupFiles: [require.resolve('./save-promise.js')]
+    .concat(reactNativePreset.setupFiles)
+    .concat([require.resolve('./restore-promise.js')]),
+});
--- a/src/__tests__/timerUtils.js
+++ b/src/__tests__/timerUtils.js
@@ -1,27 +1,14 @@
-import { setTimeout } from '../helpers/getTimerFuncs';
+// @flow
+
+import { setTimeout } from '../helpers/timers';
 
 const TimerMode = {
-  Default: 'default',
   Legacy: 'legacy',
   Modern: 'modern', // broken for now
 };
 
-function setupFakeTimers(fakeTimerType) {
-  switch (fakeTimerType) {
-    case TimerMode.Legacy:
-    case TimerMode.Modern: {
-      jest.useFakeTimers(fakeTimerType);
-      break;
-    }
-    case TimerMode.Default:
-    default: {
-      jest.useFakeTimers();
-    }
-  }
-}
-
-async function sleep(ms) {
+async function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-export { TimerMode, setupFakeTimers, sleep };
+export { TimerMode, sleep };
--- a/src/__tests__/timerUtils.js
+++ b/src/__tests__/timerUtils.js
@@ -1,0 +1,21 @@
+const FakeTimerTypes = [
+  'default',
+  'legacy',
+  // 'modern', // broken for now
+];
+
+function setupFakeTimers(fakeTimerType) {
+  switch (fakeTimerType) {
+    case 'legacy':
+    case 'modern': {
+      jest.useFakeTimers(fakeTimerType);
+      break;
+    }
+    case 'default':
+    default: {
+      jest.useFakeTimers();
+    }
+  }
+}
+
+export { FakeTimerTypes, setupFakeTimers };
--- a/src/__tests__/timerUtils.js
+++ b/src/__tests__/timerUtils.js
@@ -1,3 +1,5 @@
+import { setTimeout } from '../helpers/getTimerFuncs';
+
 const FakeTimerTypes = [
   'default',
   'legacy',
@@ -18,4 +20,8 @@ function setupFakeTimers(fakeTimerType) {
   }
 }
 
-export { FakeTimerTypes, setupFakeTimers };
+async function sleep(ms) {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+export { FakeTimerTypes, setupFakeTimers, sleep };
--- a/src/__tests__/timerUtils.js
+++ b/src/__tests__/timerUtils.js
@@ -1,19 +1,19 @@
 import { setTimeout } from '../helpers/getTimerFuncs';
 
-const FakeTimerTypes = [
-  'default',
-  'legacy',
-  // 'modern', // broken for now
-];
+const TimerMode = {
+  Default: 'default',
+  Legacy: 'legacy',
+  Modern: 'modern', // broken for now
+};
 
 function setupFakeTimers(fakeTimerType) {
   switch (fakeTimerType) {
-    case 'legacy':
-    case 'modern': {
+    case TimerMode.Legacy:
+    case TimerMode.Modern: {
       jest.useFakeTimers(fakeTimerType);
       break;
     }
-    case 'default':
+    case TimerMode.Default:
     default: {
       jest.useFakeTimers();
     }
@@ -24,4 +24,4 @@ async function sleep(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
-export { FakeTimerTypes, setupFakeTimers, sleep };
+export { TimerMode, setupFakeTimers, sleep };
--- a/src/__tests__/timers.test.js
+++ b/src/__tests__/timers.test.js
@@ -1,0 +1,23 @@
+import waitFor from '../waitFor';
+import { FakeTimerTypes, setupFakeTimers } from './timerUtils';
+
+describe.each(FakeTimerTypes)('%s fake timers tests', (fakeTimerType) => {
+  beforeEach(() => setupFakeTimers(fakeTimerType));
+
+  test('it successfully runs tests', () => {
+    expect(true).toBeTruthy();
+  });
+
+  test('it successfully uses waitFor', async () => {
+    await waitFor(() => {
+      expect(true).toBeTruthy();
+    });
+  });
+
+  test('it successfully uses waitFor with real timers', async () => {
+    jest.useRealTimers();
+    await waitFor(() => {
+      expect(true).toBeTruthy();
+    });
+  });
+});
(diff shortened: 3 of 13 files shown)


src/flushMicroTasks.js
```diff
@@ -1,5 +1,6 @@
 // @flow
 import { printDeprecationWarning } from './helpers/errors';
+import { setImmediate } from './helpers/getTimerFuncs';
```
michaelduminy, line 3: Not sure about this change. Should promises always be flushed through real timers? Where does this flush even get called?

michaelduminy: @thymikee do you know more about where this flush is used and if it should always use real timers?

michaelduminy:
@thymikee and other maintainers, what do you think of this PR?

thymikee:
Still haven't found enough time to review this properly. Please bare with us a while longer 🙈

michaelduminy:
No problem! I appreciate the demands of reviewing code like this :)

src/__tests__/timerUtils.js
```diff
@@ -0,0 +1,27 @@
+import { setTimeout } from '../helpers/getTimerFuncs';
+
+const FakeTimerTypes = [
+  'default',
+  'legacy',
+  // 'modern', // broken for now
+];
```
mdjastrzebski, line 7: Would it make more sens to expose it as an enum-like object: `{ Default 'default', Legacy: 'legacy', ...}`

src/waitFor.js
```diff
@@ -24,20 +25,24 @@ function waitForInternal<T>(
   options?: WaitForOptions
 ): Promise<T> {
   const timeout = options?.timeout ?? DEFAULT_TIMEOUT;
-  const interval = options?.interval ?? DEFAULT_INTERVAL;
-  const startTime = Date.now();
+  let interval = options?.interval ?? DEFAULT_INTERVAL;
+
+  if (interval < 1) interval = 1;
+  const maxTries = Math.ceil(timeout / interval);
```
mdjastrzebski, line 31: maxTries maybe wrong, since `interval` is not guaranted to be exact, for small intervals the delay might be larger due to evenc cycle time. 

> delay Optional
>     The time, in milliseconds (thousandths of a second), the timer should wait before the specified function or code is executed. If this parameter is omitted, a value of 0 is used, meaning execute "immediately", or more accurately, the next event cycle. Note that in either case, the actual delay may be longer than intended; see Reasons for delays longer than specified below.

https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout

Not sure if that affects your code logic.

mdjastrzebski (commented):
I'm not a specialist on intricacies of jest timers so I just toss my two cents.

src/__tests__/timerUtils.js
michaelduminy: Sounds good, I've made the changes

src/waitFor.js
michaelduminy: It's an interesting point but I don't think it affects things. There is a slight delay when scheduling timeouts and this may lead to a case where a timeout is reached later than it should have been, but this should only happen if the `interval` is 0 or otherwise very low. We could set a minimum `interval`?

Other notes:
- it's a good idea to not rely on `Date` since this can also be mocked
- this approach is what RNTL v6 used
- in the future this can be improved by borrowing an approach from `dom-testing-library`

pke:
@michaelduminy thanks for this patch, I really hope this solves the rest of the problems the modern fake timers introduced. Could you please solve the potential merge conflicts so this PR could be further reviewed and merged?

GoMino:
UP 🙄

flextremedev:
Any updates on when this gets reviewed?

thymikee:
Hey, it's a bit of a pain to get through it. Bare with us please. I've got @jaworek and @suezzo working on it, hopefully they can provide some findings here :)

src/waitFor.js
thymikee: Maybe we should adopt wait-for from DOM Testing Library, what do you think? https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js
Interesting thing is that inside they advance fake timers, which we currently miss. There may be some other small edge-cases they handle already. Sounds like a sane thing to do

src/__tests__/waitForElementToBeRemoved.test.js
```diff
@@ -130,30 +131,21 @@ test('waits with custom interval', async () => {
   expect(mockFn).toHaveBeenCalledTimes(4);
 });
 
-test('works with legacy fake timers', async () => {
-  jest.useFakeTimers('legacy');
+test.each([TimerMode.Default, TimerMode.Legacy])(
```
thymikee, line 134: we should now test all timers legacy and modern, the same, as they're supposed to work the same

src/__tests__/timerUtils.js
```diff
@@ -0,0 +1,27 @@
+import { setTimeout } from '../helpers/getTimerFuncs';
+
+const TimerMode = {
+  Default: 'default',
```
thymikee, line 4: Let's remove default. There are only 2 modes and default will change between v26 and v27.

src/__tests__/timers.test.js
```diff
@@ -0,0 +1,26 @@
+import waitFor from '../waitFor';
+import { TimerMode, setupFakeTimers } from './timerUtils';
+
+describe.each([TimerMode.Default, TimerMode.Legacy])(
```
thymikee, line 4:
```suggestion
describe.each([TimerMode.Modern, TimerMode.Legacy])(
```

src/__tests__/waitFor.test.js
```diff
@@ -79,28 +80,76 @@ test('waits for element with custom interval', async () => {
   expect(mockFn).toHaveBeenCalledTimes(3);
 });
 
-test('works with legacy fake timers', async () => {
-  jest.useFakeTimers('legacy');
+test.each([TimerMode.Default, TimerMode.Legacy])(
```
thymikee, line 83:
```suggestion
test.each([TimerMode.Modern, TimerMode.Legacy])(
```

src/__tests__/waitForElementToBeRemoved.test.js
michaelduminy: Unfortunately this does not work right now because the RN jest setup patches `global.Promise` which [breaks modern timers](https://github.com/facebook/jest/issues/10221).

[This PR](https://github.com/facebook/react-native/pull/29754/files) would need to land first.

An alternative approach is to manually capture and restore the original `global.Promise` as seen [here](https://github.com/sbalay/without_await/commit/64a76486f31bdc41f5c240d28263285683755938). WDYT?

src/__tests__/timerUtils.js
michaelduminy: Done

src/waitFor.js
michaelduminy: Changed `waitFor` to be based on DOM Testing Library's implementation

src/__tests__/waitForElementToBeRemoved.test.js
michaelduminy: I went with the alternative approach

michaelduminy:
Update: tests are all passing using legacy and modern fake timers.

Please have another look @thymikee @jaworek @suezzo 🙏

thymikee:
Thanks! Booking some time to have a look at this soon

Trancever:
I can confirm this PR works in my case. I published a fork to the npm, so feel free to try it out - https://www.npmjs.com/package/react-native-testing-library-fork

michaelduminy:
How do we feel about merging this? 😀

thymikee:
@michaelduminy I feel very strongly to do it! :D I'm just swamped with work recently and want to have a proper look at this before merging. Targeting next week for the work on this 🙂

src/__tests__/waitFor.test.js
```diff
@@ -76,39 +77,63 @@ test('waits for element with custom interval', async () => {
     // suppress
   }
 
-  expect(mockFn).toHaveBeenCalledTimes(3);
+  expect(mockFn).toHaveBeenCalledTimes(2);
```
thymikee, line 80: this is a change in behavior

jest/preset.js
```diff
@@ -0,0 +1,10 @@
+const reactNativePreset = require('react-native/jest-preset');
```
sbalay, line 1: Is this being exposed as a jest preset of `@testing-library/react-native`? If so, it should be documented somewhere

thymikee (approved):
This makes sense. I'm good with the changes that make us vendor implementation of wait-for from the dom-testing-library. I think it's safe to assume this change may be breaking for some users, so I'd like to include it in next major. We also need to expose the preset and document usage when Jest flips the switch

thymikee:
Thank you @michaelduminy, great contribution!

jest/preset.js
thymikee: Yup, will do in a followup

thymikee:
This is released under https://github.com/callstack/react-native-testing-library/releases/tag/v8.0.0-rc.0

flextremedev:
Good job! Works like a charm.

michaelduminy:
Awesome that this landed, thanks for reviewing and cleaning up things!

michaelduminy:
@thymikee Are we waiting for other PRs to land before releasing version 8 (non-RC)?

thymikee:
Not really, I should've promote the latest version to stable v8. Will do it once I find some time soon

henrymoulton:
@thymikee anything we can do to help get v8 out as stable? Happy to test the RC but currently test suite isn't that large.

thymikee:
You can treat the RC as a stable version at this point. I will promote it soon, but not sure when exactly.

henrymoulton:
thanks, appreciate you and your work on this awesome library!

Issue #325, opened 2020-05-18
Reorganize internal queries code by predicate typ


mdjastrzebski:
## Describe the Issue

Current queries are organized in files by return type (`getByAPI.js`, `queryByAPI.js`, `findByAPI.js`) with exception for `A11yAPI.js` which is organized by predicate (byA11yLabel, byA11yHint, etc). Generally `getBy` and `getAllBy` queries are used as core query, with `queryBy`/`queryAllBy` and `findBy`/`findAllBy` as manually written wrappers around `getBy`/`getAllBy` quries.

There is a lot of repeated code, especially in duplicate `getBy` and `getAllBy` queries as well as manual `queryBy`/`queryAllBy`/`findBy`/... wrappers. 

This presents an opportunity to refactor queries so that for each predicate type (`ByText`, `byA11yLabel`, etc) there is only function encoding the search algorithm, while all other functions are generated by some generic code.

In Testing Library this is achieved by implementing query predicates as `queryAll` function first, and then using generic [`buildQueries` function](https://github.com/testing-library/dom-testing-library/blob/master/src/query-helpers.js) that returnes `getBy`/`getAllBy`/etc variants.

## Benefits

1. Single definition of each predicate, so avoiding recent `getByTestId`/`getAllByTestId` differences.
1. Easier creation of additional predicates, you just need to create `queryAll` version and all other queries will generated for you by `buildQueries` function.
1. Reduced code side

## Downsides
None found yet.

## Considerations
1. This change should not change the API
1. We should keep the same unit tests, but we might reorganize them by predicate vs by return type to match new queries file structure.
1. Queries performance should be the same, because `react-test-renderer` `find` uses `findAll` under the hood without any query optimization.
1. Typescript typings tests should probably stay the same, i.e. repeated use-case statements, and not be generated for code correctness reasons.

thymikee:
Good call, let's do it.

> 3. Queries performance should be the same, because react-test-renderer find uses findAll under the hood without any query optimization.

This is an implementation detail. I guess that in some far-away future some optimizations may be added, but it's not the case now, and I don't think it has any priority. Once RTR does that, we can think about including these optimizations.

MattAgn:
I'd be glad to lend a hand on this issue @thymikee @mdjastrzebski. I could try to do it for one query before validating with you, what do you think?

thymikee:
@MattAgn lovely!

michaelduminy:
[...]

PR #573, opened 2020-10-09
Adding support for extending with custom queries

pvinis:
### Summary

After the discussion on https://github.com/callstack/react-native-testing-library/pull/569 and https://github.com/callstack/react-native-testing-library/issues/566, I tried to make the prototype for this.

I think it's kind of alright, but I need some help with the flowtyping, as I'm not too familiar with it.

### DIFF
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -382,3 +382,18 @@ test('renders options.wrapper around updated node', () => {
     </RCTSafeAreaView>
   `);
 });
+
+test('returns custom queries added', () => {
+  const _getByCustom = (instance: ReactTestInstance) =>
+    function (someArg: string) {
+      return `You sent: ${someArg}`;
+    };
+
+  const { getByCustom } = render(<View />, {
+    extendWithQueries: (instance) => ({
+      getByCustom: _getByCustom(instance),
+    }),
+  });
+
+  expect(getByCustom('yass!!')).toBe('You sent: yass!!');
+});
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -391,7 +391,7 @@ test('returns custom queries added', () => {
     };
 
   const { getByCustom } = render(<View />, {
-    extendWithQueries: (instance) => ({
+    queries: (instance) => ({
       getByCustom: _getByCustom(instance),
     }),
   });
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -384,16 +384,14 @@ test('renders options.wrapper around updated node', () => {
 });
 
 test('returns custom queries added', () => {
-  const _getByCustom = (instance: ReactTestInstance) =>
-    function (someArg: string) {
-      console.log({ instance });
-      return `You sent: ${someArg}`;
-    };
+  const _getByCustom = (instance: ReactTestInstance, someArg: string) => {
+    return `You sent: ${someArg}`;
+  };
 
   const { getByCustom } = render(<View />, {
-    queries: (instance) => ({
-      getByCustom: _getByCustom(instance),
-    }),
+    queries: {
+      getByCustom: _getByCustom,
+    },
   });
 
   expect(getByCustom('yass!!')).toBe('You sent: yass!!');
--- a/src/__tests__/render.test.js
+++ b/src/__tests__/render.test.js
@@ -383,7 +383,7 @@ test('renders options.wrapper around updated node', () => {
   `);
 });
 
-test('returns custom queries added', () => {
+test('uses custom queries', () => {
   const _getByCustom = (instance: ReactTestInstance, someArg: string) => {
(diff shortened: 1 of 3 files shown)


michaelduminy:
This would be a huge time saver while we wait for other PRs to be merged. +1

klekowskim:
I personally would go with the same API and behaviour as in react-testing-library: https://github.com/testing-library/react-testing-library/blob/c546a6f4927f925bf1187e631ca0444001f067f5/src/pure.js#L35

pvinis:
I can rename it 👍

mdjastrzebski:
I like the idea of allowing custom queries in RNTL.

Big heads up to using the same API as React Testing Library because:
- we re-use RTLs effort that went into creating that API & avoid potential mistakes
- we allow devs to easily transition between web/RN tests
- we allow some code reusage between web/RN

RTL's also exposes (`buildQueries`)[https://github.com/testing-library/dom-testing-library/blob/03cfef8498a5229851984c71c0d6ba8d93ff835d/src/query-helpers.js#L143] so that devs can easily add new quieries in all variants (getBy, getAllBy, queryBy,...) by providing just `queryAllBy` variant. It would be good to provide that as well.

thymikee:
> RTL's also exposes `buildQueries` so that devs can easily add new quieries in all variants (getBy, getAllBy, queryBy,...) by providing just `queryAllBy` variant. It would be good to provide that as well.

Sounds like a good idea for another contribution, that would align with https://github.com/callstack/react-native-testing-library/issues/325

pvinis:
I renamed it. Is this good to go then? And after this is merged we can work on the buildQueries with the refactoring anyway.

thymikee:
Can you make sure lint and flow pass?

pvinis:
Yup yup. I thought the last commit would fix that 😬.

pvinis:
Interestingly it complains even though I have an ignore right above.
https://github.com/callstack/react-native-testing-library/pull/573/files#diff-9c1abad9b45f5a8bc3d32761977018ab6ede546246de46f8c7b1cbf75b320651R384

src/render.js
```diff
@@ -36,13 +36,17 @@ export default function render<T>(
...
+
   addToCleanupQueue(renderer.unmount);
 
   return {
-    ...getByAPI(instance),
-    ...queryByAPI(instance),
-    ...findByAPI(instance),
-    ...a11yAPI(instance),
+    ...(queries: {
+      [key: $Keys<typeof queries>]: (...rest: Array<any>) => any,
```
thymikee, line 47: Struggling a bit on how to type it better, so that custom queries are inferred without forced generics. We can improve it later though

thymikee (approved):
Looks good! I made some final adjustments for the API to be compatible with Testing Library

src/__tests__/render.test.js
```diff
@@ -382,3 +382,17 @@ test('renders options.wrapper around updated node', () => {
     </RCTSafeAreaView>
   `);
 });
+
+test('returns custom queries added', () => {
```
mdjastrzebski, line 386:
```suggestion
test('uses custom queries', () => {
```

typings/index.d.ts
```diff
@@ -293,6 +296,9 @@ export interface Thenable {
 export interface RenderOptions {
```
mdjastrzebski, line 296: I check RTL/DTL and they seem to have quite decent TS typings for custom queries:

```typescript
export type Query = (
    container: HTMLElement,
    ...args: any[]
) => Error | Promise<HTMLElement[]> | Promise<HTMLElement> | HTMLElement[] | HTMLElement | null;

export interface Queries {
    [T: string]: Query;
}
```
Source: https://github.com/testing-library/dom-testing-library/blob/master/types/get-queries-for-element.d.ts

```typescript
export type RenderResult<Q extends Queries = typeof queries> = {
  // ...
} & {[P in keyof Q]: BoundFunction<Q[P]>}

export interface RenderOptions<Q extends Queries = typeof queries> {
  // ...
  queries?: Q
}
```
Source: https://github.com/testing-library/react-testing-library/blob/master/types/index.d.ts

IMO we could have the same at least for TS, possibly for flow as well if possible.

mdjastrzebski: Please note that the `Query` type restricts return value to (Promise)(Array) Element + null + Error. IMO if we provide element queries then we should also consider similar restriction. wdyt?

mdjastrzebski (commented):
Overall

thymikee:
@pvinis we just released [v7.1.0](https://github.com/callstack/react-native-testing-library/releases/tag/v7.1.0) with `container` exposed from render, so `queries` is no longer necessary for your use case. We can wait a bit more with merging this unless we have a good type story around that for Flow and TS.

mdjastrzebski:
Closing the PR as a stale code, however the idea is worth pursuing so I've created an issue #971 based on this PR.

PR #577, opened 2020-10-22
Add onMoveShouldSetResponder to element enabled checking

### LINKED ISSUE #576: fireEvent not working for "responderMove"
## Describe the bug

After updating to v7, calling fireEvent for "responderMove" stopped working. This is due to the new checks to see if a component is "disabled". Since RNTL only checks for onStartShouldSetResponder(), a component that uses onMoveShouldSetResponder() but not onStartShouldSetResponder(), will always be considered to be "disabled" by RNTL.

For testing, I replaced onStartShouldSetResponder with onMoveShouldSetResponder on the following and it solves the issue:
https://github.com/callstack/react-native-testing-library/blob/b9f9ab754e7ac6fa8e704928694332bb57e083bd/src/fireEvent.js#L26

## Expected behavior

The following would work:
```
  fireEvent(container, 'responderMove', {
    touchHistory: { mostRecentTimeStamp: '2', touchBank: [] },
  });

```
## Versions

  npmPackages:
    @testing-library/react-native: ^7.0.2 => 7.0.2 
    react: 16.13.1 => 16.13.1 
    react-native: https://github.com/expo/react-native/archive/sdk-39.0.3.tar.gz => 0.63.2 
    react-test-renderer: ^16.11.0 => 16.13.1

abmantis:
### Summary

This enables elements that have onMoveShouldSetResponder returning true, but not onStartShouldSetResponder.
Fixes #576 .

### DIFF
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -3,6 +3,7 @@ import React from 'react';
 import {
   View,
   TouchableOpacity,
+  PanResponder,
   Pressable,
   Text,
   ScrollView,
@@ -304,4 +305,28 @@ test('is not fooled by non-responder wrapping host elements', () => {
 
   fireEvent.press(screen.getByText('Trigger'));
   expect(handlePress).not.toHaveBeenCalled();
+});
+
+function TestDraggableComponent({ onDrag }) {
+  const responderHandlers = PanResponder.create({
+    onMoveShouldSetPanResponder: (_evt, _gestureState) => true,
+    onPanResponderMove: onDrag,
+  }).panHandlers;
+
+  return (
+    <View {...responderHandlers}>
+      <Text>Trigger</Text>
+    </View>
+  );
+}
+
+test('has only onMove', () => {
+  const handleDrag = jest.fn();
+
+  const screen = render(<TestDraggableComponent onDrag={handleDrag} />);
+
+  fireEvent(screen.getByText('Trigger'), 'responderMove', {
+    touchHistory: { mostRecentTimeStamp: '2', touchBank: [] },
+  });
+  expect(handleDrag).toHaveBeenCalled();
 });
--- a/src/fireEvent.js
+++ b/src/fireEvent.js
@@ -23,7 +23,13 @@ const isEventEnabled = (
 ) => {
   if (isTextInput(element)) return element?.props.editable !== false;
 
-  return touchResponder?.props.onStartShouldSetResponder?.() !== false;
+  const touchStart = touchResponder?.props.onStartShouldSetResponder?.();
+  const touchMove = touchResponder?.props.onMoveShouldSetResponder?.();
+
+  if (touchStart) return true;
+  if (touchMove) return true;
+
+  return touchStart === undefined && touchMove === undefined;
 };
 
 const findEventHandler = (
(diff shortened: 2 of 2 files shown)


thymikee:
Can you add a test? 🙂

src/fireEvent.js
```diff
@@ -23,7 +23,10 @@ const isEventEnabled = (
 ) => {
   if (isTextInput(element)) return element?.props.editable !== false;
 
-  return touchResponder?.props.onStartShouldSetResponder?.() !== false;
+  return !(
+    touchResponder?.props.onStartShouldSetResponder?.() === false ||
+    touchResponder?.props.onMoveShouldSetResponder?.() === false
+  );
```
mdjastrzebski, line 29:
```suggestion
  const isAcceptingTouchStart = touchResponder?.props.onStartShouldSetResponder?.() !== false;
  const isAcceptingTouchMove = touchResponder?.props.onMoveShouldSetResponder?.() !== false;
  
  return isAcceptingTouchStart || isAcceptingTouchMove;
```

To clarify the reason for the original `onStartShouldSetResponder() !== false` contion, it is meant to mean:
* `onStartShouldSetResponder() === true` view explicitly wants to accept touch events
* `onStartShouldSetResponder() === false` view explicitly wants to reject touch events
* `onStartShouldSetResponder == undefined` view did not explicitly declare accepting touch events, which in real RN runtime env would probably prevent accepting touch events BUT in order to simulate events on various components in Jest runtime env we assume that missing value means accepting the touch event.

Having that it mind, I wonder is it possible that one of  `onStartShouldSetResponder` & `onMoveShouldSetResponder` returns explicit `true`, while the other return explicit `false`?

mdjastrzebski (changes requested):
Nice catch @abmantis! 

Please also provide:
- [ ] (some) relevant test(s)

src/fireEvent.js
abmantis: Yeah, Start can return false and Move return true, so I had to add additional checks. Now it seems to work well.

@thymikee yeah, it really needed tests :D

src/__tests__/fireEvent.test.js
```diff
@@ -305,3 +306,27 @@ test('is not fooled by non-responder wrapping host elements', () => {
...
+
+  return (
+    <View {...responderHandlers}>
+      <Text>Trigger</Text>
+    </View>
+  );
+}
+
+test('has only onMove', () => {
+  const handlePress = jest.fn();
```
thymikee, line 324: let's rename it `handleDrag` to avoid confusion

thymikee (approved):

abmantis:
@mdjastrzebski Any input?

src/fireEvent.js
```diff
@@ -23,7 +23,13 @@ const isEventEnabled = (
 ) => {
   if (isTextInput(element)) return element?.props.editable !== false;
 
-  return touchResponder?.props.onStartShouldSetResponder?.() !== false;
+  const touchStart = touchResponder?.props.onStartShouldSetResponder?.();
+  const touchMove = touchResponder?.props.onMoveShouldSetResponder?.();
+
+  if (touchStart) return true;
+  if (touchMove) return true;
```
mdjastrzebski, line 30:
```suggestion
  if (touchStart || touchMove) return true;
```

mdjastrzebski (approved):

thymikee:
Thank you for your contribution @abmantis !

Issue #579, opened 2020-10-26
The links in the "How Should I Query?" page are not pointing to the correct links


ryanjwessel:
## Describe the bug

Clicking any one of the links from the "How Should I Query?" page on the documentation website results in [this screen](https://i.imgur.com/CVemA7n.png)

The current path looks like this: https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/api-queries#bydisplayvalue

So, it seems that the URLs are relative to the "How Should I Query?" page.

## Expected behavior

The link should point to the correct query path on the documentation website (example: https://callstack.github.io/react-native-testing-library/docs/api-queries#byplaceholdertext)

## Steps to Reproduce

- Go to the "How Should I Query" page
- Click on any of the referenced query links.

## Screenshots

[Example screenshot](https://i.imgur.com/CVemA7n.png)

## Versions

Current documentation version?

thymikee:
PRs welcome :)

ryanjwessel:
I would be happy to! I'll aim to have one for today.

ryanjwessel:
The links are working on the production website now, so this can be closed. Thanks, @thymikee!

PR #580, opened 2020-10-26
fix: Update "How Should I Query?" links to use the correct URLs

ryanjwessel:
This addresses #579 

### Summary

This updates the "How Should I Query?" links to use the correct URLs for each query type. 

### Test plan

I verified these URLs are all correct.


thymikee (commented):
Wouldn't prefixing the existing links with `/docs/` help? If not, we can leave absolute paths, but relative would be preferable

ryanjwessel:
I tested while running the website locally, and prefixing with `docs/` results in this URL (note the duplicate `docs/`): http://localhost:3000/react-native-testing-library/docs/docs/api-queries#bytext

Turns out, the relative paths were working locally, but when they are hosted in production they are incorrect.

Local relative path: http://localhost:3000/react-native-testing-library/docs/api-queries#bytext
Production relative path: https://callstack.github.io/react-native-testing-library/docs/how-should-i-query/api-queries#bydisplayvalue

Issue #547, opened 2020-09-17
Show stack-trace for failed findBy or waitFor


sregg:
## Describe the Feature
Currently, if your test has a lot of `await findBy`, if one of them fails, it's hard to know which one from the error message.
It'd be great if the error message would include the line of the `findBy` that fails, the same way it does for `getBy`

with `findBy`:

```
  ● Testing react navigation › clicking on one item takes you to the details screen

    No instances found
```

with `getBy`:

```
 ● Testing react navigation › clicking on one item takes you to the details screen

    No instances found

      24 |     fireEvent(toClick, 'press');
      25 |     const newHeader = await findByText('Showing details for 5');
    > 26 |     const newBody = getByText('the number you have chosen is 6');
```

## Possible Implementations
I'd be happy to look into the code and see if I can provide a PR if nobody is currently working on this.

thymikee:
Happy to accept a PR with this improvement! :)

sregg:
@thymikee has anyone looked into this already?
I don't want to spend hours to find out it's not possible if someone already went through that process 😅

thymikee:
None that I know of

sregg:
I spent almost a full day trying to add the stack-traces but I couldn't make it work properly.
It was actually working find in the internal tests of the lib. Those were already showing some stack-trace, just not the right one. 
By surrounding the code `waitFor` with a `try catch` I was able to show the correct stack-trace:
```javascript

  try {
    let result: T;

    //$FlowFixMe: `act` has incorrect flow typing
    await act(async () => {
      result = await waitForInternal(expectation, options);
    });
    //$FlowFixMe: either we have result or `waitFor` threw error
    return result;
  } catch (error) {
    throw new ErrorWithStack(prepareErrorMessage(error), waitFor);
  }
```
When I run it in my project, I don't get any stack-trace, before or after that change.

thymikee:
So you say that "I was able to show the correct stack-trace" and then "I don't get any stack-trace, before or after that change". I don't quite understand what you mean by that 😅. Have you made this change in node_modules of your project and it still didn't work?

sregg:
I used yalc to get the change in my project.
I validated with some console logs that I was indeed seeing the changes.
What I saw is that in my project, I wouldn't see any stack-trace at all, even without any changes in the lib.
Whereas in the lib's tests, I see a stack-trace (in waitFor) even without any changes.

thymikee:
Gotta fix it then!

[...]

PR #581, opened 2020-10-26
fix: display stack trace and code frame for findBy error

matthieugicquel:
# Summary

This fixes https://github.com/callstack/react-native-testing-library/issues/547 with a solution more or less copied from [this file in dom-testing-library](https://github.com/testing-library/dom-testing-library/blob/master/src/wait-for.js).

A useful stack trace and codeframe are now shown for `findBy` failures.

### Test plan

Capture d’écran 2020-10-26 à 18 04 53

Note that to get the correct codeframe, the repo running the test must be another that uses this repo as a package, this won't work in the internal tests.

I didn't add any test because I'm not sure assertions on the stack trace make sense...

Contrary to the issue above, the stacktrace was already working with `waitFor` for me, so while it should improve the situation for `waitFor` calls too, I can't be sure of it.

### Improving the stack trace more

The stack trace still has 2 useless/internal lines, but I suspect to get rid of these, we must:
- either filter the stack trace string manually
- or rewrite all the `findBy***` helpers to capture the error here

I figured I would open this PR before digging more, since the most important part works :)

### DIFF
--- a/src/helpers/errors.js
+++ b/src/helpers/errors.js
@@ -39,6 +39,13 @@ export const createQueryByError = (error: Error, callsite: Function) => {
   throw new ErrorWithStack(error.message, callsite);
 };
 
+export function copyStackTrace(target: Error, stackTraceSource: Error) {
+  target.stack = stackTraceSource.stack.replace(
+    stackTraceSource.message,
+    target.message
+  );
+}
+
 const warned = {};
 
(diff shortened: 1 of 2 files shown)


thymikee (approved):
Thanks!

suezzo (approved):

PR #582, opened 2020-10-28
fix: add container to TypeScript types

jason0x43:
### Summary

Update the TS types for the new `container` property that was added to the output of `render` in #567.

### Test plan

In a TS project using the current version of @testing-library/react-native (currently 7.1.0), try accessing `container` on the result of a `render` call. A typing error will occur because RenderAPI (the result of `render`) doesn't have a `container` property.

```ts
const bodyText = render(<BodyText>Foo</BodyText>);
expect(bodyText.container).toHaveTextContent('Foo'); // <-- will generate type error
```

With this PR, no error is shown.


thymikee (approved):
Thanks!

PR #583, opened 2020-10-28
Update API.md

MetaMmodern:
Shallow is removed since migration to v2.0. Mentioned [here](https://callstack.github.io/react-native-testing-library/docs/migration-v2#removed-global-shallow-function)
Also no method is available after calling `debug()`. VSCode also shows error when trying to call `shallow()` after `.debug` or `.debug()`

### Summary

Resolved misunderstanding in API. 
### Test plan

image
image
image

### DIFF
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -106,10 +106,6 @@ optional message
 </View>
 ```
 
-#### `debug.shallow`
-
-Pretty prints shallowly rendered component passed to `render` with optional message on top.
-
 ### `toJSON`
 
 ```ts


thymikee:
`debug.shallow()` still works. Sounds like it's not typed correctly. Would you like to send a PR to the `index.d.ts` file?

MetaMmodern:
> 
> 
> `debug.shallow()` still works. Sounds like it's not typed correctly. Would you like to send a PR to the `index.d.ts` file?

If you mean removing last 3 lines from `index.d.ts`  then yes, I can, but I don't understand what do you mean by "is not typed correctly". Typed by me or by maintainers of rntl?

thymikee:
See https://github.com/callstack/react-native-testing-library/pull/584.
Thanks for the contribution anyway!

MetaMmodern:
> See https://github.com/callstack/react-native-testing-library/pull/584.
> Thanks for the contribution anyway!

I'm just new to TS)

PR #608, opened 2020-11-29
chore: restructure ByTestId queries code by predicate type

MattAgn:
### Summary

This PR aims to solve [this issue](https://github.com/callstack/react-native-testing-library/issues/325).

Tasks:

- [ ] group queries by their predicate type (byTestId, byText) instead of grouping them by their return 
type (different for get, find, query).
- [ ] make query builders that use the queryAllBy query as the base to create the getBy, getAllBy, queryBy, findBy and findAllBy queries
- [ ] group the tests by predicate type as well

### Requirements
- [ ] The API should not change

### Test plan


MattAgn:
@thymikee @mdjastrzebski here's what I did so far:
- make new query builders based on queryAll
- use them to generate byTestId queries
- group related tests

Apart from that, I'm not too happy about the typing of the query builders I implemented (there is an any that I don't know how to replace with flow).

Is this kind of refactor what you had in mind?

thymikee:
Looks promising! Nice job

thymikee:
@MattAgn I've pushed to your fork with updated typings and rebase after the upgrade to latest RN with newer Flow version. Can you take a look? :)

MattAgn:
Thanks @thymikee! It's all good to me, thanks for the help on the typing!

For the error messages, I wonder if it would be a good idea to generify them as well to keep them coherent across the different queries. Like "No instances found with <ATTRIBUTE> = "myAttribute". From what I've seen, the message is already identical (except for the attribute) for most queries right?

thymikee:
I've changed the error messages here to reflect what's in the DOM testing library, as I find the wording nicer. We can copy this 1:1, no problem from our side. We can also change it any time and surely make generic where it makes sense

mdjastrzebski (commented):
Overall looks going into right direction. I'm glad you started working on this improvement 🚀 

I would suggest looking into `src/helpers/makeQuery.js` which has similar functionality of building different query types from `queryAll` query. That code has a little more functions, as it's also responsible for building query aliases which is not relevant for this PR.

IMO best API would be to have a single `makeQueries` function taking in `queryAllByXxx` query (and maybe some more args) and returning object with all query types: {query,get,find}[All]ByXxx. With the current code you will have to repeat the query construction (as in byTestId.js lines 27-55) for every query type. Finally we could export the results with proper names and typing: `getBy` => `getByTestId`

MattAgn:
Thanks for the review @mdjastrzebski! I'll take care of the requested changes

src/helpers/makeQueries.js
```diff
@@ -0,0 +1,86 @@
+// @flow
+import waitFor from '../waitFor';
+import type { WaitForOptions } from '../waitFor';
+import { ErrorWithStack } from './errors';
+
```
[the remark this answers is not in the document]
MattAgn: Done :)

src/helpers/makeQueries.js
```diff
@@ -0,0 +1,86 @@
...
+type AllQuery = <T>(
+  instance: ReactTestInstance
+  // $FlowFixMe - Property @@iterator is missing in T [1] but exists in $Iterable [2]
+) => (...args: T) => Array<ReactTestInstance>;
+
+export function makeGetAllQuery<T>(
+  allQuery: AllQuery,
+  instance: ReactTestInstance,
+  getMissingError: (args: T) => string
+): (...args: Array<T>) => Array<ReactTestInstance> {
```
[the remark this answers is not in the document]
MattAgn: What will be the rule for the options? 
Could we have a type like `type Args<OptionType> = [string|RegExp, OptionType]`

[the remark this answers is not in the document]
MattAgn: Because if we use mixed, we will lose the typing of the arg and options right?

src/helpers/makeQueries.js
```diff
@@ -0,0 +1,114 @@
...
+>;
+type FindByQuery<QueryArg> = FindQueryFunction<QueryArg, ReactTestInstance>;
+
+export type Queries<QueryArg> = {
+  getBy: GetByQuery<QueryArg>,
+  getAllBy: GetAllByQuery<QueryArg>,
+  queryBy: QueryByQuery<QueryArg>,
+  findBy: FindByQuery<QueryArg>,
+  findAllBy: FindAllByQuery<QueryArg>,
+};
```
thymikee, line 38: @MattAgn I made Queries parametrized with QueryArg, should solve the problem of queries having different args.

thymikee (approved):
Exciting!

src/helpers/makeQueries.js
MattAgn: Nice! I tried it as well but couldn't make it work with the spread operator if we were to add options for example, or simply to make it work for the find queries, which also have waitForOptions

thymikee:
@mdjastrzebski ping :)

src/helpers/byTestId.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+import { makeQueries } from './makeQueries';
+import type { Queries } from './makeQueries';
+
+const getNodeByTestId = (node, testID) => {
```
mdjastrzebski, line 5: Consider adding explicit types

mdjastrzebski (approved):
Excellent work! This code is well-structured and easy to read. 🥇🥇 

I guess that as a follow-up PR(s) we could work reusing makeQueries to generate all other queries.

PS. Really sorry for long waiting time

src/helpers/byTestId.js
thymikee: We can address that in a followup if necessary (I suspect Flow infers that, as it's internal fn)

MattAgn:
@thymikee @mdjastrzebski thanks for the approval! (it was my first open source contribution and it's been a great experience!)
For the following work on makeQueries, would you rather have one PR with all queries or a bunch of PRs with a few queries addressed in each?

mdjastrzebski:
@MattAgn Now we have the common part of the code already solved, as well as pattern to follow. IMO I would go with one PR, as it boils down to defining `queryAll` for each query type and proper exports. Tests and TS typedefs probably don't need to change.

@thymikee wdyt?

thymikee:
>  IMO I would go with one PR, as it boils down to defining queryAll for each query type and proper exports. Tests and TS typedefs probably don't need to change.

👍🏼. We can, however, rearrange tests later to comply to the new structure, if needed :)

src/helpers/byTestId.js
MattAgn: I confirm, flow infers it 
image

PR #626, opened 2020-12-09
chore: bump React Native to 0.64 RC

suezzo:
### Summary

Making sure we're compatible with the latest upcoming RN and Flow versions.

### Test plan

CI


thymikee (approved):
Thanks!

thymikee:
Needs a rebase

src/waitForElementToBeRemoved.js
```diff
@@ -2,8 +2,9 @@
 import waitFor, { type WaitForOptions } from './waitFor';
 import { ErrorWithStack } from './helpers/errors';
 
-const isRemoved = (result) =>
-  !result || (Array.isArray(result) && !result.length);
+function isRemoved<T>(result: T): boolean {
```
suezzo, line 5: I had to replace arrow function, because vscode has some problems with them when using generic types (syntax isn't highlighted properly)

src/helpers/a11yAPI.js
```diff
@@ -6,11 +6,11 @@ import makeQuery from './makeQuery';
 type GetReturn = ReactTestInstance;
 type GetAllReturn = Array<ReactTestInstance>;
 type QueryReturn = ReactTestInstance | null;
-type QueryAllReturn = Array<ReactTestInstance> | [];
```
thymikee, line 9: Removed the `[]` fallback, as Flow treats it as a tuple and thinks that the array has no elements. Which may be true, but would require users to optional-chain a lot

src/helpers/a11yAPI.js
```diff
@@ -116,7 +116,7 @@ export function matchObject<T: {}>(prop?: T, matcher: T): boolean {
     : false;
 }
 
-const a11yAPI = (instance: ReactTestInstance): A11yAPI =>
+export const a11yAPI = (instance: ReactTestInstance): A11yAPI =>
```
thymikee, line 119: changed to be consistent with other queries

thymikee (approved):

PR #654, opened 2021-01-15
Finish refactoring queries using new query builder makeQueries

MattAgn:
### Summary

This PR aims to solve [this issue](https://github.com/callstack/react-native-testing-library/issues/325). It is the follow up of this [PR](https://github.com/callstack/react-native-testing-library/pull/608)

Tasks:

- [ ] group queries by their predicate type (byTestId, byText) instead of grouping them by their return 
type (different for get, find, query).
- [ ] maybe group the tests by predicate type instead of query type?

### Requirements
- [ ] The API should not change

### Test plan

### DIFF
--- a/src/__tests__/byTestId.test.js
+++ b/src/__tests__/byTestId.test.js
@@ -129,6 +129,13 @@ test('getByTestId, queryByTestId', () => {
 
   expect(getByTestId('bananaFresh')).toBe(component);
   expect(queryByTestId('InExistent')).toBeNull();
+
+  expect(() => getByTestId('duplicateText')).toThrow(
+    'Found multiple elements with testID: duplicateText'
+  );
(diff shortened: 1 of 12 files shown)


MattAgn:
@thymikee @mdjastrzebski I modified a bit the error messages to try to unify them for the different predicate types and the ones for byTestId which we modified on the previous PR

MattAgn:
I did not tackle the UNSAFE and a11y queries (since they have their own builder already), is it alright?

thymikee (approved):
This is great, thank you! 🚀

MattAgn:
thanks for the approval @thymikee! Should I group the tests by query in a new PR now? Because for now it is done for only one query (byTestId) and not the others

thymikee:
Sure, would be amazing to have the tests cleaned up as well :)

PR #655, opened 2021-01-21
feat: support pointerEvents

### LINKED ISSUE #650: Add support for pointerEvents
## Describe the Feature

Currently, when there is a wrapping `View` with `pointerEvents="none"` testing lib still allows pressing buttons that are rendered inside of that View.

Expected behaviour:

```js
// component
<View pointerEvents="none">
  <Touchable onPress={onPress}>Click me</Touchable>
</View>

...

// test
const mockOnPress = jest.fn();
const { getByText } = render(<Component onPress={mockOnPress} />);

fireEvent.press(getByText('Click me'));

expect(mockOnPress).not.toHaveBeenCalled();
```

## Possible Implementations

No idea, cause I am not familiar with the codebase :(

suezzo:
### Summary

This PR allows handling test cases that use `pointerEvents` which was mentioned in #650. I'm not sure if we should also additionally handle the `box-none` value. 

Fixes #650

### Test plan

I've added 3 tests that check pointerEvents `none` and `box-only` and also pointerEvents `none` with nested views.

### DIFF
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -234,7 +234,7 @@ test('should not fire on non-editable TextInput with nested Text', () => {
   expect(onChangeTextMock).not.toHaveBeenCalled();
 });
 
-test.only('should not fire on non-targetable View', () => {
+test('should not fire on none pointerEvents View', () => {
   const handlePress = jest.fn();
 
   const screen = render(
@@ -242,6 +242,38 @@ test.only('should not fire on non-targetable View', () => {
       <Pressable onPress={handlePress}>
         <Text>Trigger</Text>
       </Pressable>
+    </View>
+  );
+
+  fireEvent.press(screen.getByText('Trigger'));
+  expect(handlePress).not.toHaveBeenCalled();
+});
+
+test('should not fire on box-only pointerEvents View', () => {
+  const handlePress = jest.fn();
+
+  const screen = render(
+    <View pointerEvents="box-only">
+      <Pressable onPress={handlePress}>
+        <Text>Trigger</Text>
+      </Pressable>
+    </View>
+  );
+
+  fireEvent.press(screen.getByText('Trigger'));
+  expect(handlePress).not.toHaveBeenCalled();
+});
+
+test('should not fire on none pointerEvents View with nested elements', () => {
+  const handlePress = jest.fn();
+
+  const screen = render(
+    <View pointerEvents="box-only">
+      <View>
+        <Pressable onPress={handlePress}>
+          <Text>Trigger</Text>
+        </Pressable>
+      </View>
     </View>
   );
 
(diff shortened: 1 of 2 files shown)


src/fireEvent.js
```diff
@@ -17,11 +17,25 @@ const isTouchResponder = (element?: ReactTestInstance) => {
   return !!element?.props.onStartShouldSetResponder || isTextInput(element);
 };
 
+const isPointerEventEnabled = (element?: ReactTestInstance) => {
+  if (
+    element?.props.pointerEvents === 'none' ||
+    element?.props.pointerEvents === 'box-only'
```
satya164, line 23: `box-only` means this element doesn't disable pointers, but its children do. `box-none` means this element disables pointers, but it's children won't.

So with this code, if a `View` has `box-only`, and I try to trigger a press on it, then it would return `false` here whereas it's actually pressable.

Maybe there needs to be a second argument `isParent` or something, and then it could be something like:

```js
isParent
  ? element?.props.pointerEvents === 'box-only'
  : element?.props.pointerEvents === 'box-none'
```

thymikee:
Let's support `box-none` as well. It would be subpar experience to provide only partial support. So let's implement all cases specified here: https://reactnative.dev/docs/view#pointerevents

satya164 (commented):
Can you also add tests for following scenarios:

- Firing event on the same element which has `pointerEvents="none"` (`<Text onPress={handlePress} pointerEvents="none" />`)
- Firing event on the same element which has `pointerEvents="box-none"` (`<Text onPress={handlePress} pointerEvents="box-none" />`)
- Firing event on the same element which has `pointerEvents="box-only"` (`<Text onPress={handlePress} pointerEvents="box-only" />`)

suezzo:
I don't think that `<Text />` accepts `pointerEvents` prop. It's not mentioned in the docs - https://reactnative.dev/docs/textinput and type checker has a problem with it. Correct me if I'm wrong but only `<View />` component has `pointerEvents` prop?

src/__tests__/fireEvent.test.js
```diff
@@ -234,6 +234,83 @@ test('should not fire on non-editable TextInput with nested Text', () => {
...
+        <Text>Trigger</Text>
+      </Pressable>
+    </View>
+  );
+
+  fireEvent.press(screen.getByText('Trigger'));
+  expect(handlePress).toHaveBeenCalled();
+});
+
+test('should not fire on none pointerEvents View with nested elements', () => {
```
thymikee, line 297:
```suggestion
test('should not fire on box-only pointerEvents View with nested elements', () => {
```

thymikee (approved):

satya164 (commented):
I think it would still be nice to have some tests which trigger `onPress` on the same element that has `pointerEvents`

Issue #649, opened 2021-01-07
getByDisplayValue doesn't match TextInput when defaultValue is used


jdpagley:
## Describe the bug

getByDisplayValue doesn't find the TextInput when a defaultValue prop is used instead of value. It will match properly if I change `defaultValue` to just `value`

## Expected behavior

getByDisplayValue should match a TextInput component if it uses defaultValue or value prop.

## Steps to Reproduce

Try to use getByDisplayValue to match a TextInput with the value set via the defaultValue prop instead of value.

## Screenshots

The top of the screenshot shows the output of the `debug()` call. You can see that the query below doesn't match when defaultValue is used.
![image](https://user-images.githubusercontent.com/4016177/103942366-471e1080-50fe-11eb-9feb-b4ff5d825c71.png)

```
// Test Case
describe('TextField', function() {

  beforeEach(() => {

    this.props = {
      fieldId: 'fieldId',
      label: 'This is a test label',
      value: 'hello world',
      onChange: jest.fn(),
      onFocus: jest.fn(),
      onBlur: jest.fn(),
      onLayout: jest.fn(),
    }

  });

 it('should display defaultValue when passed in', () => {
    const { getByDisplayValue, debug } = render(<TextField {...this.props}/>);
    debug();
    expect(getByDisplayValue('hello world'));
 });

})

```

```
//Render method of component
  render() {

    const {
      fieldId,
      label,
      value,
      placeholder,
      disabled,
      onChange,
      onBlur,
      onFocus,
      onLayout
    } = this.props;

    return (
      <FieldHOC label={label} onLayout={onLayout}>
        <TextInput 
          testID={`${fieldId}-text-field`}
          style={Styles.field}
          disabled={disabled} 
          defaultValue={value}
          placeholder={placeholder}
          onBlur={(e) => {
            this.setState({focused: false});
            if (onBlur) onBlur(fieldId, e)
          }}
          onFocus={(e) => {
            this.setState({focused: true});
            if (onFocus) onFocus(fieldId, e)
          }}
          onChangeText={(changes) => {
            this.setState({changes});
            if (onChange) onChange(fieldId, changes);
          }}
        />
      </FieldHOC>
    );
  }
```
## Versions

![image](https://user-images.githubusercontent.com/4016177/103942627-ae3bc500-50fe-11eb-8fbc-6b762dfe5a60.png)

MattAgn:
[...]

PR #656, opened 2021-01-23
Fix issue of getByDisplayValue not checking TextInputs defaultValue

MattAgn:
### Summary

This PR aims to solve [this issue](https://github.com/callstack/react-native-testing-library/issues/649). Basically the issue to fix is that the byDisplayValue helpers don't check the defaultValue prop of TextInputs.

### Requirements

- [ ] If a TextInput only has a defaultValue and no value, getByDisplayValue should find it
- [ ] If a TextInput has a defaultValue AND a value, getByDisplayValue should only find it by its value and not its defaultValue (since if a value is provided, the default value won't be displayed)

### Test plan

- [ ] Check that getByDisplayValue and queryByDisplayValue get the TextInput by defaultValue if no value is provided
- [ ] Check that getByDisplayValue and queryByDisplayValue don't get the TextInput by defaultValue if a value is provided

### DIFF
--- a/src/helpers/getByAPI.js
+++ b/src/helpers/getByAPI.js
@@ -107,11 +107,10 @@ const getTextInputNodeByPlaceholderText = (node, placeholder) => {
 const getTextInputNodeByDisplayValue = (node, value) => {
   try {
     const { TextInput } = require('react-native');
+    const nodeValue = node.props.value || node.props.defaultValue;
     return (
       filterNodeByType(node, TextInput) &&
-      (typeof value === 'string'
-        ? value === node.props.value
-        : value.test(node.props.value))
+      (typeof value === 'string' ? value === nodeValue : value.test(nodeValue))
     );
   } catch (error) {
     throw createLibraryNotSupportedError(error);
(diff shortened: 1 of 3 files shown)


src/helpers/getByAPI.js
```diff
@@ -107,11 +107,10 @@ const getTextInputNodeByPlaceholderText = (node, placeholder) => {
 const getTextInputNodeByDisplayValue = (node, value) => {
   try {
     const { TextInput } = require('react-native');
+    const nodeValue = node.props.value || node.props.defaultValue;
```
thymikee, line 110: I think we want to be explicit here and check if `node.props.value === undefined`

MattAgn: indeed that would be better, which value do we use if value is empty string?

thymikee: I _think_ it's always gonna be a string for Text, but the only way to be sure would be to test it. 

Nevertheless, if the `value` is `""` the defaultValue is no longer used as the `value` is set to not-undefined

MattAgn: Oh alright I did not know that, I'll make the changes

MattAgn: Flow does not let me put 0 in the value prop of the TextInput, it wants a string. Should I ignore it or do we consider it's fine? (especially since we'll now be checking for undefined and not false values)

thymikee: No need to test the number then :)

thymikee (approved):
Thank you!

suezzo (approved):

PR #657, opened 2021-01-25
:books: fix typo in README.md

AbdellahRsmouki:
### Summary
The jest.config file extention should be ".json" instead of  ".js"


thymikee (approved):
Both js and json are correct, but since we're using json syntax, this is fine. Thanks!

PR #663, opened 2021-02-05
fix: add React.Fragment support for matching text

lachiet:
### Summary

When using getByText() on a custom component using nested Fragments, it should return the text content within the fragments.

```jsx
const CustomText = () => {
  return <React.Fragment>Hello World</React.Fragment>;
};

<Text>
  <CustomText />
</Text>
```

### Test plan

I've added an additional test and all other tests are passing. Let me know if there's any concerns or other things we can add to protect against regression.

### DIFF
--- a/src/helpers/getByAPI.js
+++ b/src/helpers/getByAPI.js
@@ -41,7 +41,7 @@ export type GetByAPI = {|
   getAllByPlaceholder: () => void,
 |};
 
-const filterNodeByType = (node, type) => node.type === type;
+const filterNodeByType = (node, type) => node?.type === type;
 
 const getNodeByText = (node, text) => {
   try {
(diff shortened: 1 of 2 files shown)


thymikee (commented):
Thanks for this improvement! What if, instead, we adjust the `getChildrenAsText` to detect Text with an array of strings?

lachiet:
> Thanks for this improvement! What if, instead, we adjust the `getChildrenAsText` to detect Text with an array of strings?

@thymikee no problem! I initially tried that approach but it seems like I can't access the children or props.children from the child object? This is what I get when I `console.log(child)` out before [this line](https://github.com/callstack/react-native-testing-library/blob/master/src/helpers/getByAPI.js#L77) - any ideas?

```
{
  '$$typeof': Symbol(react.element),
  type: [Function: CustomText],
  key: null,
  ref: null,
  props: {},
  _owner: null,
   _store: {}
}
```

src/helpers/getByAPI.js
```diff
@@ -56,6 +56,17 @@ const getNodeByText = (node, text) => {
           : text.test(textToTest);
       }
     }
+
+    const isParentTextComponent = filterNodeByType(node.parent?.parent, Text);
```
thymikee, line 60: checking parents non-recursively will result in a failure in e.g. this scenario:
```js
const Wrap = ({ children }) => <>{children}</>;

  const CustomText = () => {
    return <React.Fragment>Hello World</React.Fragment>;
  };

  expect(
    render(
      <Text>
        <Wrap>
          <CustomText />
        </Wrap>
      </Text>
    ).queryByText('Hello World')
  ).toBeTruthy();
```

thymikee: Please check this PR as well: https://github.com/callstack/react-native-testing-library/pull/554 as it's changing the way we're interfacing text to be similar to React Testing Library. Grabbing by text with non-direct string children is tricky and we'll likely need to trade-off some functionality, or come up with a more complex solution.

We want to solve this problem once for good, so if you're willing to help improve the library, please chime in there and see how we can support this scenario with string precision API

thymikee: FYI, #554 is now merged. Would you be still interested in pursuing this improvement? 😊

thymikee:
Closed via #934

PR #672, opened 2021-02-17
Add event blur and focus in fireEvent

misaku:
### Summary
i had to test state on TexInput has focus and blur, i try to change this code and work for me

### Test plan
```JS
   fireEvent.focus(getByTestId('elementTest'));
   fireEvent.blur(getByTestId('elementTest'));
```

### DIFF
--- a/src/fireEvent.js
+++ b/src/fireEvent.js
@@ -124,6 +124,10 @@ const toEventHandlerName = (eventName: string) =>
 
 const pressHandler = (element: ReactTestInstance): void =>
   invokeEvent(element, 'press', pressHandler);
+const focusHandler = (element: ReactTestInstance): void =>
+  invokeEvent(element, 'focus', pressHandler);
+const blurHandler = (element: ReactTestInstance): void =>
+  invokeEvent(element, 'blur', pressHandler);
 const changeTextHandler = (
   element: ReactTestInstance,
   ...data: Array<any>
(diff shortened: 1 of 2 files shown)


thymikee:
Thanks for you contribution! I'm not sure if we want to add more aliases to the `fireEvent` for now. You can use the generic form to construct any kind of event. And if we decided to add more, I'd like to avoid half-measures like adding only a few special cases and not documenting them.

misaku:
> Thanks for you contribution! I'm not sure if we want to add more aliases to the `fireEvent` for now. You can use the generic form to construct any kind of event. And if we decided to add more, I'd like to avoid half-measures like adding only a few special cases and not documenting them.

thanks, can you give one example how can i do this? im new in test ;)

thymikee:
Check the example here: https://callstack.github.io/react-native-testing-library/docs/api/#fireevent

eliw00d:
@thymikee The web version of TL has this section on focus: https://testing-library.com/docs/guide-events/#focusblur and awhile back there was a discussion on discouraging fireEvent for focus [here](https://github.com/testing-library/user-event/issues/350). I was just curious if eventually fireEvent will be more in-line with web and you would have to do something like `getBy*(...).focus()` in RNTL? Since, for example, [TextInput has .focus](https://reactnative.dev/docs/textinput#focus). It would be nice to follow the same patterns in both libraries.

thymikee:
I'm afraid the simulated Node.js environment we run in is not as sophisticated as JSDOM used by RTL. You should be good using `getBy*(...).focus()`, but I haven't tested it really. I'd appreciate som docs contribution in this topic, ideally alongside extended tests in the suite we have :)

PR #673, opened 2021-02-28
Fixed broken link of mock setup file

ThakurBallary:
### Summary

In documentation jest-setup.js file from the example, was mentioned as jest-mocks.js. 

### Test plan

Create a [mock file](https://github.com/callstack/react-native-testing-library/blob/master/examples/reactnavigation/jest-setup.js) necessary for your tests: (Now, on click redirects to correct file)


thymikee (approved):
Thanks!

PR #674, opened 2021-03-01
Make onPress more life-like

hontas:
Taking inspiration from [user-event](https://testing-library.com/docs/ecosystem-user-event) which tries to mimic the real event order when user interacts with an element.

### Summary

`fireEvent.press` currently only invoke the `onPress`-handler, and this PR aim to make it more real-life-like by also invoking `onPressIn`, `onPressOut` and `onFocus` if possible (silently failing if there are no registered handlers).

Also adding `fireEvent.longPress` for convenience.

### Test plan

Test added in [fireEvent.test.js](./src/__tests__/fireEvent.test.js)

### Thoughts

One could discuss that this should be another method called `tap` or something else that map closer to user intent than to event handler name. Looking forward to hear your feedback.

### DIFF
--- a/src/fireEvent.js
+++ b/src/fireEvent.js
@@ -5,6 +5,10 @@ const isHostElement = (element?: ReactTestInstance) => {
   return typeof element?.type === 'string';
 };
 
+const isFocused = (element?: ReactTestInstance) => {
+  return element?._component?.isFocused?.();
+};
+
 const isTextInput = (element?: ReactTestInstance) => {
   const { TextInput } = require('react-native');
   return element?.type === TextInput;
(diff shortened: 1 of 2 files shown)


src/fireEvent.js
```diff
@@ -119,11 +123,38 @@ const invokeEvent = (
   return returnValue;
 };
 
+const tryInvokeEvent = (...args) => {
+  try {
+    return invokeEvent(...args);
```
thymikee, line 128: We actually don't the `invokeEvent` to be throwable. It supposed to be there as a debugging aid, but the original Testing Library don't have this behavior and here it actually complicates things a little more. 

How about removing this behavior altogether in a separate PR and adjusting this one to keep the `invokeEvent` without the try/catch?

thymikee (commented):
Great addition, thank you!

src/fireEvent.js
hontas: Sure thing!
Just to make sure i understood 🤓 
- create a PR that removes the throwing code from `invokeEvent`
- remove the tryInvokeEvent from this PR

thymikee: Yup :)

hontas: Here it is: #691

hontas:
Removed the throwing behaviour from invokeEvent here: #691
That PR should be merged before this one :)

thymikee:
Mind rebasing? :)

hontas:
@thymikee done :)

rarenatoe:
Will this ever be merged? I need to be able to work with onFocus and I can't test it atm.

AugustinLF:
Long term, shouldn't we try to to either create `native-user-event` package, or at least export something specific for that in this library? I'm not sure that having a different behaviour between the web and native library when doing a simple `fireEvent` is the best.

mdjastrzebski:
Closing in favour of #1119 which would be more organised approach that is also similar to RTL.

PR #691, opened 2021-03-01
feat(breaking): do not throw because event handler was not found

hontas:
### Summary

Idea from: https://github.com/callstack/react-native-testing-library/pull/674#discussion_r584551609

Quote from @thymikee:
```
We actually don't the invokeEvent to be throwable. It supposed to be there as a debugging aid, but the original Testing Library don't have this behavior and here it actually complicates things a little more.

How about removing this behavior altogether in a separate PR and adjusting this one to keep the invokeEvent without the try/catch?
```

### DIFF
--- a/src/__tests__/fireEvent.test.js
+++ b/src/__tests__/fireEvent.test.js
@@ -62,6 +62,16 @@ describe('fireEvent', () => {
     expect(onPressMock).toHaveBeenCalled();
   });
 
+  test('should not fire if the press handler is not passed to children', () => {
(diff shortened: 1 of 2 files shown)


thymikee (approved):

thymikee:
Thank you! We are not yet able to get rid of the `parent.parent` check, because it's already buggy and we'd contribute further to breaking a broken behavior (!sic). Need to figure out a proper solution first :)

hontas:
Thanks for the quick feedback 😀

PR #697, opened 2021-03-03
Debug implicit any error

benjaminreid:
### Summary

In a project that didn’t allow implicit `any`s, this error would be thrown for `Debug`.

```
Call signature, which lacks return-type annotation, implicitly has an 'any' return type.
```

### Test plan

Use the `yarn run typescript-check`.

I’ve tested this change against a live project and it corrects the error, though perhaps there is a better type?

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -299,7 +299,7 @@ export interface RenderOptions {
 }
 
 type Debug = {
-  (message?: string): any;
+  (message?: string): void;
   shallow: (message?: string) => void;
 };
 
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -299,7 +299,7 @@ export interface RenderOptions {
 }
 
 type Debug = {
-  (message?: string);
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -299,7 +299,7 @@ export interface RenderOptions {
 }
 
 type Debug = {
-  (message?: string);
+  (message?: string): any;
```
thymikee, line 302: `void` would be a proper return type here, what do you think?
```suggestion
  (message?: string): void;
```

benjaminreid: Agreed! Will commit the suggestion, thank you 👍🏻

thymikee (approved):
Thanks!

tebatalla:
Do you mind sending out a release with this fix soon?

retyui:
WHEN THIS FIX WILL BE RELEASED?

thymikee:
When I publish it.

PR #704, opened 2021-03-19
Update react navigation  docs to support RN-0.64

McSam94:
### Summary

Update docs in the react-navigation example section that broke the test after upgraded to react-native v0.64


thymikee (approved):
Thank you!

thymikee:
Looking forward to future contributions 🙂

PR #751, opened 2021-06-04
docs: Expo with TypeScript getting started

louiechristie:
### Summary

Expo with TypeScript getting started instructions:

References:
    
    * https://forums.expo.io/t/tdd-typescript-testing-testing-library-react-native/42700
    * https://stackoverflow.com/questions/65549722/jest-expo-crashes-on-example-react-createelement-type-is-invalid-expected-a/65608967#65608967
    * https://docs.expo.io/guides/testing-with-jest/

### Test plan

N/A

### DIFF
--- a/website/docs/GettingStarted.md
+++ b/website/docs/GettingStarted.md
@@ -64,6 +64,28 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 }
 ```
 
+### Expo with TypeScript
+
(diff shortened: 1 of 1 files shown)


thymikee:
This includes some modules that are not necessarily used, like Sentry or Native Base. Since this is the Jest configuration issue, I prefer linking to the official docs on the matter to also educate users on why is that and how to resolve similar problems in the future: https://jestjs.io/docs/tutorial-react-native#transformignorepatterns-customization

thymikee:
Closing due to inactivity. Happy to reopen if you're interesting in pushing it forward :)

Issue #776, opened 2021-07-04
Bug: `Found multiple elements with text`


andreialecu:
## Describe the bug

It seems that in a certain case the library will find more than one matching element with `getByText`. See the reproduction.

## Expected behavior

Should find just one element.

## Steps to Reproduce

```js
function TestText({ children }) {
  return <Text>{children}</Text>;
}

function TestComponent() {
  return (
    <TestText>
      <TestText>Foo</TestText>
    </TestText>
  );
}

test('shows bug', () => {
  const component = render(<TestComponent />);
  component.getByText('Foo'); // Found multiple elements with text: Foo
});

function TestComponent2() {
  return (
    <Text>
      <Text>Foo</Text>
    </Text>
  );
}

test('works properly', () => {
  const component = render(<TestComponent2 />);
  component.getByText('Foo'); // OK
});
```

`debug` shows:
```
<Text>
  <Text>
    Foo
  </Text>
</Text>
```

Note that if `TestText` is replaced with just `Text`, so that it is used directly, the error goes away.

## Versions

npmPackages:
    @testing-library/react-native: ^7.2.0 => 7.2.0 
    react: 17.0.1 => 17.0.1 
    react-native: 0.64.1 => 0.64.1 
    react-test-renderer: 17.0.1 => 17.0.1

andreialecu:
Reproduction on this repo and branch:

https://github.com/andreialecu/rnavigation-jest-repro/tree/bug-gettext

andreialecu:
#554 fixes this. Just tested it.

milesj:
Just ran into this. Has there been any progress on landing that fix?

andreialecu:
@milesj I had a chat with @thymikee on the Callstack Discord and the current plan was to hold off from merging #554, because it is a breaking change. The plan was to release 8.0 next, without this fix, and then land the fix for this as part of a future version 9.0.

IMO the changes in #554 should likely not be breaking for too many people, and the benefits outweight the negatives.

So, unless there's a change of mind about including that fix in 8.0, it seems like this is a long way away from being fixed.

milesj:
That's unfortunate... If it's breaking, I don't see why it couldn't be included in 8 😓

prometheas:
Bumping this with a comment to get maintainer attention, as this is one of several issues I'm running into with my component tests after upgrading my React Native from `0.63.4` to `0.64.2`.

PR #778, opened 2021-07-06
test: add failing tests for #776

andreialecu:
### Summary

This shows two failing test cases for #776 

Note that #554 fixes both of them, so if #554 is merged, they can be enabled.

### Test plan

They test themselves 🙂 
 
Skipped because they fail on current `master`:

Screenshot 2021-07-06 at 17 11 32


thymikee (approved):
Thank you for contributing tests 🚀

Issue #659, opened 2021-01-26
Non-visible modal content shows up in queries


kirbycool:
## Describe the bug

Hi, thanks for taking a look at this!
I'm fairly new to react-native, but have a lot of experience with web testing-library variants and ran across some unexpected (to me) behavior. Query methods find nodes in non-visible modals.

I think this is what this user was getting at https://github.com/callstack/react-native-testing-library/issues/508

## Steps to Reproduce

Here's a simple example test I set up to test this out
```js
// App.js
import React from "react";
import { Modal, Text, View } from "react-native";

const App = () => {
  return (
    <View>
      <Modal visible={false}>
        <View>
          <Text>Exciting content</Text>
        </View>
      </Modal>
    </View>
  );
};

export default App;
```

```
// App.test.js
import React from "react";
import { render, fireEvent } from "@testing-library/react-native";
import App from "./App";

it("shows the modal on button click", () => {
  const { queryByText } = render(<App />);

  expect(queryByText("Exciting content")).toBeNull(); // fails
});
```

## Expected behavior

I'd expect this test to pass, since there's no modal content visible to the user while the modal is not visible. 

If this is expected behavior, is there a recommended way to test appearance/disappearance modals?

## Versions

@testing-library/react-native@7.1.0

MattAgn:
I think it's due to the fact that the getBy and queryBy queries parse the DOM itself. The modal is always in the DOM, its appearance to the user is dictated by its `visible` prop, which the queries like getBy don't take into account.

Personally what I usually tend to do in those cases is check the `visible` prop on the modal. 

I think a better solution might be to abstract this hack (checking the visible prop) in an extension of jest expect method, that could be put in the [jest native repo](https://github.com/testing-library/jest-native) for example.

kirbycool:
Yeah that makes sense. Thanks for taking a look. 

Testing the `visible` prop isn't ideal since it's an escape hatch from the "test from the user perspective" philosophy. Ideally we'd be able to query for "is this content on the screen", which is how the user sees it and makes the test more robust in case the content is refactored outside of a modal later on.

I suppose this is more of a feature request given the context (and would probably be a breaking change). Maybe this will bug me enough to try implementing it at some point 😄

MattAgn:
Totally understand your point of view. Unfortunately I think it's deeply tied to the modal implementation with the visible prop. In my mind, changing the getByText query to take into account the modal would imply checking all the parents of the node until you find (or not) a modal right? 
[...]

PR #787, opened 2021-07-27
feat: accessibility findAll

gabrielgrover:
### Summary

This PR adds a new option to the render function called respectAccessibility.  When this property is true a new implementation of ReactTestInstance.findAll() is 
used.  This new findAll() is defined in render.js inside the function `appendFindAllTrap`.  This new findAll ignores sub trees that have a root node with a prop
`accessibilityElementsHidden: true`.  This solution is an attempt to mimic the first approach found in [@MattAgn 's list](https://github.com/callstack/react-native-testing-library/issues/659#issuecomment-830781869).

### Test plan
Tests added in `src/__tests__/byTestId.test.js`, `src/__tests__/byPlaceholderText.test.js`, and `src/__tests__/byDisplayValue.test.js`
- run `yarn test`
  - tests should pass

### Scope
- [x] Implement `findAll` trap
- [x] Add tests
- [ ] Add flow & typescript type definitions
- [ ] Update documentation

### DIFF
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -111,7 +111,7 @@ test('all queries should respect accessibility', async () => {
   );
   const Comp = () => (
     <View>
-      <OtherComp accessibilityElementsHidden />
+      <OtherComp importantForAccessibility="no-hide-descendants" />
       <TextInput value="test_02" />
       <View>
         <TextInput value="test_02" />
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -102,12 +102,23 @@ test('findBy queries work asynchronously', async () => {
 }, 20000);
 
 test('all queries should respect accessibility', async () => {
-  const Comp = () => (
-    <View>
-      <View accessibilityElementsHidden>
+  const Input = () => (
+    <View >
+      <View>
         <TextInput value="test_01" />
       </View>
+    </View>
+  );
+  const Comp = () => (
+    <View>
+      <Input accessibilityElementsHidden />
       <TextInput value="test_02" />
+      <View>
+        <TextInput value="test_02" />
+      </View>
+      <View>
+        <TextInput accessibilityElementsHidden value="test_01" />
+      </View>
     </View>
   );
 
@@ -120,14 +131,30 @@ test('all queries should respect accessibility', async () => {
     queryByDisplayValue,
   } = render(<Comp />, { respectAccessibility: true });
 
-  await expect(findAllByDisplayValue('test_01')).rejects.toBeTruthy();
-  await expect(findByDisplayValue('test_01')).rejects.toBeTruthy();
+  await expect(getAllByDisplayValue("test_02")).toHaveLength(2);
   await expect(() => getAllByDisplayValue('test_01')).toThrow(
     'Unable to find an element with displayValue: test_01'
   );
-  await expect(queryAllByDisplayValue('test_01')).toHaveLength(0);
+  await expect(() => getByDisplayValue("test_02")).toThrow(
+    "Found multiple elements with display value: test_02 "
+  );
   await expect(() => getByDisplayValue('test_01')).toThrow(
     'Unable to find an element with displayValue: test_01'
   );
+  await expect(queryAllByDisplayValue('test_01')).toHaveLength(0);
+  await expect(queryAllByDisplayValue("test_02")).toHaveLength(2);
   await expect(queryByDisplayValue('test_01')).toBeNull();
+  await expect(() => queryByDisplayValue("test_02")).toThrow(
+    "Found multiple elements with display value: test_02 "
+  );
+  await expect(findAllByDisplayValue('test_01')).rejects.toEqual(
+    new Error("Unable to find an element with displayValue: test_01")
+  );
+  await expect(findAllByDisplayValue('test_02')).resolves.toHaveLength(2);
+  await expect(findByDisplayValue('test_01')).rejects.toEqual(
+    new Error("Unable to find an element with displayValue: test_01")
+  );
+  await expect(findByDisplayValue('test_02')).rejects.toEqual(
+    new Error("Found multiple elements with display value: test_02 ")
+  );
 });
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -107,13 +107,16 @@ test('all queries should respect accessibility', async () => {
       <View>
         <TextInput value="test_01" />
       </View>
+      <View accessibilityViewIsModal>
+        <TextInput value="test_02" />
+      </View>
     </View>
   );
   const Comp = () => (
     <View>
       <OtherComp importantForAccessibility="no-hide-descendants" />
       <OtherComp style={{display: "none"}} />
-      <TextInput value="test_02" />
+      <OtherComp />
       <View>
         <TextInput value="test_02" />
       </View>
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -1,7 +1,6 @@
 // @flow
 import * as React from 'react';
 import { View, TextInput } from 'react-native';
-import { isExportDeclaration } from 'typescript';
 
 import { render } from '..';
 
@@ -118,17 +117,17 @@ test('all queries should respect accessibility', async () => {
     getAllByDisplayValue,
     queryAllByDisplayValue,
     getByDisplayValue,
-    queryByDisplayValue
+    queryByDisplayValue,
   } = render(<Comp />, { respectAccessibility: true });
 
-  await expect(findAllByDisplayValue("test_01")).rejects.toBeTruthy();
-  await expect(findByDisplayValue("test_01")).rejects.toBeTruthy();
-  await expect(() => getAllByDisplayValue("test_01")).toThrow(
-    "Unable to find an element with displayValue: test_01"
+  await expect(findAllByDisplayValue('test_01')).rejects.toBeTruthy();
+  await expect(findByDisplayValue('test_01')).rejects.toBeTruthy();
+  await expect(() => getAllByDisplayValue('test_01')).toThrow(
+    'Unable to find an element with displayValue: test_01'
   );
-  await expect(queryAllByDisplayValue("test_01")).toHaveLength(0);
-  await expect(() => getByDisplayValue("test_01")).toThrow(
-    "Unable to find an element with displayValue: test_01"
+  await expect(queryAllByDisplayValue('test_01')).toHaveLength(0);
+  await expect(() => getByDisplayValue('test_01')).toThrow(
+    'Unable to find an element with displayValue: test_01'
   );
-  await expect(queryByDisplayValue("test_01")).toBeNull();
+  await expect(queryByDisplayValue('test_01')).toBeNull();
 });
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -102,7 +102,7 @@ test('findBy queries work asynchronously', async () => {
 }, 20000);
 
 test('all queries should respect accessibility', async () => {
-  const Input = () => (
+  const OtherComp = () => (
     <View >
       <View>
         <TextInput value="test_01" />
@@ -111,7 +111,7 @@ test('all queries should respect accessibility', async () => {
   );
   const Comp = () => (
     <View>
-      <Input accessibilityElementsHidden />
+      <OtherComp accessibilityElementsHidden />
       <TextInput value="test_02" />
       <View>
         <TextInput value="test_02" />
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -112,6 +112,7 @@ test('all queries should respect accessibility', async () => {
   const Comp = () => (
     <View>
       <OtherComp importantForAccessibility="no-hide-descendants" />
+      <OtherComp style={{display: "none"}} />
       <TextInput value="test_02" />
       <View>
         <TextInput value="test_02" />
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -119,7 +119,7 @@ test('all queries should respect accessibility', async () => {
     queryAllByDisplayValue,
     getByDisplayValue,
     queryByDisplayValue
-  } = render(<Comp />, { respectAccessibilityProps: true });
+  } = render(<Comp />, { respectAccessibility: true });
 
   await expect(findAllByDisplayValue("test_01")).rejects.toBeTruthy();
   await expect(findByDisplayValue("test_01")).rejects.toBeTruthy();
--- a/src/__tests__/byDisplayValue.test.js
+++ b/src/__tests__/byDisplayValue.test.js
@@ -1,6 +1,7 @@
 // @flow
 import * as React from 'react';
 import { View, TextInput } from 'react-native';
+import { isExportDeclaration } from 'typescript';
 
 import { render } from '..';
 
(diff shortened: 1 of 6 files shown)


thymikee:
cc @satya164. Would you like to check this out? Should make testing React Navigation more life-like :)

src/__tests__/byDisplayValue.test.js
```diff
@@ -100,3 +100,34 @@ test('findBy queries work asynchronously', async () => {
...
+  const {
+    findAllByDisplayValue,
+    findByDisplayValue,
+    getAllByDisplayValue,
+    queryAllByDisplayValue,
+    getByDisplayValue,
+    queryByDisplayValue,
+  } = render(<Comp />, { respectAccessibility: true });
+
+  await expect(findAllByDisplayValue('test_01')).rejects.toBeTruthy();
```
mdjastrzebski, line 123: consider `using rejects.toEqual(...)`

mdjastrzebski: nit: pls sort queries by they typical sort order: gets, query, find. Having them randomly shuffled raises question why they are in this particular order.

src/__tests__/byDisplayValue.test.js
```diff
@@ -100,3 +100,34 @@ test('findBy queries work asynchronously', async () => {
...
   await expect(findAllByDisplayValue('Display Value')).resolves.toHaveLength(1);
 }, 20000);
+
+test('all queries should respect accessibility', async () => {
+  const Comp = () => (
+    <View>
+      <View accessibilityElementsHidden>
+        <TextInput value="test_01" />
+      </View>
+      <TextInput value="test_02" />
```
mdjastrzebski, line 110: Pls add some more `expects` here that query test_02 so that we know that `accessibilityElementsHidden` did not affect sibling elements.

src/__tests__/byText.test.js
```diff
@@ -85,6 +85,73 @@ test('getAllByText, queryAllByText', () => {
...
+    'Unable to find an element with text: hello world'
+  );
+  await expect(() => getByText('hello world')).toThrow(
+    'Unable to find an element with text: hello world'
+  );
+  await expect(queryAllByText('hello world')).toHaveLength(0);
+  await expect(queryByText('hello world')).toBeNull();
+});
+
+test('queryAllByText returns an empty array if respectAccessibilityProps is true', () => {
```
mdjastrzebski, line 121: Seems like it covers the same cases the above test, pls merge these together.

src/render.js
```diff
@@ -107,3 +110,28 @@ function debug(
   debugImpl.shallow = (message) => debugShallow(instance, message);
   return debugImpl;
 }
+
+function appendFindAllTrap(renderer: ReactTestRenderer) {
+  return new Proxy(renderer.root, {
+    get(target, prop) {
```
mdjastrzebski, line 116: nit, readability: could we get some more type explicit type annotations here?

src/render.js
```diff
@@ -107,3 +110,28 @@ function debug(
...
+      const isFindAllProp = prop === 'findAll';
+      const newFindAll = (fn, node = target) => {
```
mdjastrzebski, line 118: can `newFindAll` be moved to external variable and be type-annotated?

src/render.js
```diff
@@ -107,3 +110,28 @@ function debug(
...
+        return node.children.reduce((result, child) => {
+          if (typeof child === 'string') {
+            return result;
+          }
+
+          return result.concat(newFindAll(fn, child));
+        }, initial);
+      };
+
+      return isFindAllProp ? newFindAll : Reflect.get(...arguments);
```
mdjastrzebski, line 134: Would that work?
```suggestion
      return isFindAllProp ? newFindAll : target[prop];
```

src/render.js
```diff
@@ -107,3 +110,28 @@ function debug(
...
+        }
+
+        const initial = fn(node) ? [node] : [];
+
+        return node.children.reduce((result, child) => {
+          if (typeof child === 'string') {
+            return result;
+          }
+
+          return result.concat(newFindAll(fn, child));
```
mdjastrzebski, line 130: nit:
```suggestion
          return [...result, ...newFindAll(fn, child)];
```

src/render.js
```diff
@@ -107,3 +110,28 @@ function debug(
...
+      const isFindAllProp = prop === 'findAll';
+      const newFindAll = (fn, node = target) => {
+        if (node.props.accessibilityElementsHidden) {
+          return [];
+        }
+
+        const initial = fn(node) ? [node] : [];
+
+        return node.children.reduce((result, child) => {
+          if (typeof child === 'string') {
```
mdjastrzebski, line 126: Seems like we need also to handle `number` type here as well (maybe also `boolean`). Please create tests with `<Text>{1}</Text>`, `{false}`, `{null}`, `{undefined}` and see if it crashes.

mdjastrzebski (commented):
I love the idea of this PR. Great work!

Couple of things to improve:
* `findAll` trap seems to work in simple cases but we should also check if it works in more complex cases by writing tests with:
  * more nested components (in both host & composite components)
  * `<Text>{1}</Text>`, `{null}`, etc (check detailed comments).
* add tests that cover finding of non-hidden elements in siblings of  `accessibilityElementsHidden` elements
* other minor stylistic issues (improved, typing, etc)

Also new params seems not added typescript defintions.

mdjastrzebski:
@thymikee I'm really impressed by this PR. After ironing our all of the potential edge cases we should consider this to be default option for next major release, as most of regular multi-screen tests would benefit from it.

mdjastrzebski:
As the intended use case if with React Navigation, perhaps we could find a way to actually test it with that library.

gabrielgrover:
Thanks for the detailed review @mdjastrzebski !  I will look through and address them after work today

mdjastrzebski:
@gabrielgrover after some more consideration of `findAll` proxy trap, I think there is a risk that our implementation will behave slightly differently than `findAll` from React Test Renderer. To overcome this I see two approaches:
1) base our implementation on code from React Test Renderer (i.e. copy-paste-modify) so that we match their behavior,
2) or rather call the original `findAll` but filter the results be doing recursive `parent` traversal and removing elements for which any parent has `accessibilityElementsHidden` set to true.

IMO I would lean to option 2, as it would allows as to remove proxy code and will allow us to rely on React Test Renderer `findAll` implementation, without copying their code.

CC: @thymikee

gabrielgrover:
> @gabrielgrover after some more consideration of `findAll` proxy trap, I think there is a risk that our implementation will behave slightly differently than `findAll` from React Test Renderer. To overcome this I see two approaches:
> 
> 1. base our implementation on code from React Test Renderer (i.e. copy-paste-modify) so that we match their behavior,
> 2. or rather call the original `findAll` but filter the results be doing recursive `parent` traversal and removing elements for which any parent has `accessibilityElementsHidden` set to true.
> 
> IMO I would lean to option 2, as it would allows as to remove proxy code and will allow us to rely on React Test Renderer `findAll` implementation, without copying their code.
> 
> CC: @thymikee

So I actually did option 1.  It is actually where i got the following line
```typescript
  if (typeof child === 'string') { ... }
```
ReactTestRenderer basically does a depth first search.  The only difference is my use of .reduce().  As for option 2 the only issue I see with calling their `findAll` is that there doesn't seem to be a way for me to tell the parent or child of a given node.  The call back for `findAll` is called by ReactTestRenderer with a node as the first argument.  Maybe I am missing something, but there doesn't seem to be anything in that node object that can uniquely identify itself nor its parent / children.  If I am wrong about this please let me know because I agree that it would be a better approach to lean on their implementation.  However if I am right then I think we would need to stick to option 1

satya164:
Excited for this PR!

I was thinking, since `accessibilityElementsHidden` is iOS only, that we'd need to handle few more accessibility-related props as well:
- Similar prop for Android: `importantForAccessibility="no-hide-descendants"`
- For all platforms: `display: 'none'` in styles
- On iOS: `accessibilityViewIsModal` - This should be equivalent to marking sibling elements as non-accessible

mdjastrzebski:
Re `findAll` ancestor traversal:  (@gabrielgrover) I'm not sure if i understood your comment correctly, but since original `findAll` is returning `TestInstance` array, we could use `TestInstance` `parent` prop to navigate to root ancestors and simply filter out these elements that have `accessibilityElementsHidden=true`.

Pseudo-code:
```javascript
function isVisibleToAccessibility(element: TestInstance) {
  const isElementVisible = !instance.accessibilityElementsHidden;
  const isParentVisible = !instance.parent || isVisibleToAccessibility(instance.parent);
  return isParentVisible && isElementVisible;
}

// Then
const elements = findAll(...);
const visibleElements = elements.filter(isVisibleToAccessibility);
```

mdjastrzebski:
Re `accessibilityViewIsModal` (@satya164): this prop details sound a little confusing. Could you clarify that I am understanding this correctly? 

If there is only one host component with `accessibilityViewIsModal` we threat it's sibling as they have `accessibilityElementsHidden == true`.

If there is more than one host components with `accessibilityViewIsModal` only one of these is actually filtering out its siblings. And the "effective" component with `accessibilityViewIsModal` is the one that is the deepest in the hierarchy?

mdjastrzebski:
(Closed by mistake, reopened this PR)

satya164:
> If there is only one host component with `accessibilityViewIsModal` we threat its sibling as they have `accessibilityElementsHidden == true`.

Yes, I think we can just treat `<A accessibilityViewIsModal /><B />` as `<A /><B accessibilityElementsHidden />`.

> If there is more than one host components with `accessibilityViewIsModal` only one of these is actually filtering out its siblings. And the "effective" component with `accessibilityViewIsModal` is the one that is the deepest in the hierarchy?

I don't know how it'll work with multiple siblings with `accessibilityViewIsModal`, normally there needs to be only one modal shown at a time. We can check on an iOS device to see how `VoiceOver` reacts to it. If there are views with `accessibilityViewIsModal` nested inside each other, then yea, as per my understanding, the innermost one is the active modal.

gabrielgrover:
> Re `findAll` ancestor traversal: (@gabrielgrover) I'm not sure if i understood your comment correctly, but since original `findAll` is returning `TestInstance` array, we could use `TestInstance` `parent` prop to navigate to root ancestors and simply filter out these elements that have `accessibilityElementsHidden=true`.
> 
> Pseudo-code:
> 
> ```js
> function isVisibleToAccessibility(element: TestInstance) {
>   const isElementVisible = !instance.accessibilityElementsHidden;
>   const isParentVisible = !instance.parent || isVisibleToAccessibility(instance.parent);
>   return isParentVisible && isElementVisible;
> }
> 
> // Then
> const elements = findAll(...);
> const visibleElements = elements.filter(isVisibleToAccessibility);
> ```

I see.  We started with different assumptions.  I was going about this problem with the assumption that we wouldn't want to traverse the tree more than once for each call to `render`.

mdjastrzebski:
I do not see any real problem (performance or other) with traversing the component tree bottom-up or top-down multiple times (as long as it's not an infinite loop ;-)). Performance impact will be limited to running the integration tests.

Digging a bit into details, traversing component tree bottom-up from given node should be cheaper than top-down, because going up you have only one parent, while going down you can have multiple children. Simplifying things a little bit, traversing up from given node should have `O(log n)` complexity while traversing down from root should have `O(n)` complexity, when using big O notation. Additionally we will do the traversal only for small number of components, those that are returned from `findAll` function, which should be in range from 0 to say 5, most of the time.

There for cost of these extra traversals bottom-up seems acceptable tradeoff compared to benefit of simpler and more robust code, which avoids patching renderer object and duplicating `findAll` algorithm.

@gabrielgrover: wdyt?

gabrielgrover:
> I do not see any real problem (performance or other) with traversing the component tree bottom-up or top-down multiple times (as long as it's not an infinite loop ;-)). Performance impact will be limited to running the integration tests.
> 
> Digging a bit into details, traversing component tree bottom-up from given node should be cheaper than top-down, because going up you have only one parent, while going down you can have multiple children. Simplifying things a little bit, traversing up from given node should have `O(log n)` complexity while traversing down from root should have `O(n)` complexity, when using big O notation. Additionally we will do the traversal only for small number of components, those that are returned from `findAll` function, which should be in range from 0 to say 5, most of the time.
> 
> There for cost of these extra traversals bottom-up seems acceptable tradeoff compared to benefit of simpler and more robust code, which avoids patching renderer object and duplicating `findAll` algorithm.
> 
> @gabrielgrover: wdyt?

Sounds good to me. Gonna work on it over the weekend.

gabrielgrover:
@satya164 @mdjastrzebski Sorry just now have been able to dedicate some more time to this PR.  Taking a look at the edge case for `accessibilityViewIsModal`.  I assume I can check siblings by doing something like this 

```typescript
  const currentNodeIsHidden = instance.parent.children.some(c => c.props.accessibilityViewIsModal);
```

However, this would cause the node with the property `accessibilityViewIsModal` to also be "hidden".  Is there someway I can identify a specific node so i can exclude it from the callback in the above code?  Or can I make some assumptions on the number of elements with `accessibilityViewIsModal`?

satya164:
> this would cause the node with the property accessibilityViewIsModal to also be "hidden".

I don't know any implementation details, but isn't something like this possible?

```js
const currentNodeIsHidden = instance.parent.children.some(
  c => c !== instance && c.props.accessibilityViewIsModal
);
```

gabrielgrover:
> > this would cause the node with the property accessibilityViewIsModal to also be "hidden".
> 
> I don't know any implementation details, but isn't something like this possible?
> 
> ```js
> const currentNodeIsHidden = instance.parent.children.some(
>   c => c !== instance && c.props.accessibilityViewIsModal
> );
> ```

`c` and `instance` are both ReactTestInstances which are objects.  There is no general way to compare objects.  That's why we need to find some property of these objects that has a primitive value and is unique.  Like an id of some sort. Or we can use JSON.stringify, but we will start hitting performance issues at that point.

gabrielgrover:
> > > this would cause the node with the property accessibilityViewIsModal to also be "hidden".
> > 
> > 
> > I don't know any implementation details, but isn't something like this possible?
> > ```js
> > const currentNodeIsHidden = instance.parent.children.some(
> >   c => c !== instance && c.props.accessibilityViewIsModal
> > );
> > ```
> 
> `c` and `instance` are both ReactTestInstances which are objects. There is no general way to compare objects. That's why we need to find some property of these objects that has a primitive value and is unique. Like an id of some sort. Or we can use JSON.stringify, but we will start hitting performance issues at that point.

Actually it looks like using Object.is() will work.

```typescript
const currentNodeIsHidden = instance.parent.children.some(
  c => !Object.is(c, instance) && c.props.accessibilityViewIsModal
);

```

gabrielgrover:
@mdjastrzebski @satya164 pushed changes.  Let me know what you think

src/render.js
```diff
@@ -107,3 +110,64 @@ function debug(
   debugImpl.shallow = (message) => debugShallow(instance, message);
   return debugImpl;
 }
+
+function appendFindAllTrap(renderer: ReactTestRenderer) {
```
mdjastrzebski, line 114: I would suggest removing `Proxy` code, and using `filter(isReactTestElementVisibleToAccessibility)` in the queries. This way `respectAccessibility` would be moved from `render` option to query option, which would better reflect the fact that we are apply query modifier and not render modifier.

We should have some base `QueryOptions` type that would atm be `{ respectAccessibility: boolean }` and would be passed as a second argument to all `get/getAll/query`/etc/ queries. Note, for `findBy` & `findAllBy` we would add the existing `waitForOptions` there.

mdjastrzebski (commented):
Thanks for building up this PR.

I've noticed two points that would like to discuss:
1) Atm `respectAssesibility` is a `render` option. However, it does not actual modify the render output (other than `findAll` method). IMO it would be better to make it queries option (`get/getAll/query`/etc). Each query would accept a 2nd optional alrgument of options and `respectAssesibility` would be the first available one.
2) In code checking accessibility props, you currently allow accessibility props to be added to composite components, this is especially visible in test code. However in order to reflect how RN renderer works, we should only take into account props set on host components (`View`, `Text`, etc). This is similar to styling composite components, you can have a `style` prop there, but the UI will not be styled unless you somehow apply that `style` to host component.

@gabrielgrover Pls let me know if you need any coding help in this PR. I'm eager to help :-)

mdjastrzebski:
I'm closing this PR as stale, but since the idea is work pursuing I've create #970 issue which references this PR.

PR #812, opened 2021-09-09
Update index.d.ts

thatsLegit:
toJSON function can return a list of ReactTestRendererJSON

While I was testing my application, I realized 
Screenshot 2021-09-09 at 16 38 32
Screenshot 2021-09-09 at 16 39 16


thymikee (approved):
Thanks!

PR #876, opened 2021-12-03
Add v9 release docs

AugustinLF:
As discussed in https://github.com/callstack/react-native-testing-library/discussions/862#discussioncomment-1705927, here is a proposal for the v9 docs. @thymikee let me know if you had something different in mind.

### DIFF
--- a/website/docs/MigrationV9.md
+++ b/website/docs/MigrationV9.md
@@ -3,13 +3,13 @@ id: migration-v9
 title: Migration to 9.0
 ---
 
-Version 7.0 brought this package as the standard `react-native` implementation in the `@testing-library`. However it has been implemented independently from its web counterpart, therefore there were some differences in API an behaviour. Version 9.0 solves several of these problems.
+Version 7.0 brought React Native Testing Library into the `@testing-library` family. Since it has been implemented independently from its web counterpart – the React Testing Library – there are some differences in the API and behavior. Version 9.0 solves several of these problems.
 
-# API changes
+## Support for text match options a.k.a string precision API
 
-## Support for text match options a.k.a string precision API (backward compatible change)
+This is a backward compatible change.
 
-When querying text, it is now possible to pass a [`TextMatch`](https://callstack.github.io/react-native-testing-library/docs/api-queries/#textmatch) to most text based queries, which lets you configure how `@testing-library/react-native` should match your text. For instance, passing `exact: false` will allow matching substrings and will ignore case.
+When querying text, it is now possible to pass a [`TextMatch`](https://callstack.github.io/react-native-testing-library/docs/api-queries/#textmatch) to most text based queries, which lets you configure how `@testing-library/react-native` should match your text. For instance, passing `exact: false` will allow matching substrings and will ignore case:
 
 ```jsx
 const { getByText } = render(<Text>Hello World</Text>);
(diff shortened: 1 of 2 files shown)


website/docs/MigrationV9.md
```diff
@@ -0,0 +1,40 @@
...
+
+const {getByText} = render(<Text>Hello <Text>world</Text</Text>)
+
+// exact: false
+within(getByText('Hello', {exact: false})).getByText('world')
+```
+
+# Future plans
+
+This release changes a lot of internal logic in `@testing-library/react-native`, paving the way for more more improvements to bring us closer from our web counterpart, with for instance a better support for accessibility queries.
```
thymikee, line 40:
```suggestion
This release changes a lot of internal logic in the library, paving the way for more improvements to bring us closer to our web counterpart, with a possibly better story for accessibility queries.
```

website/docs/MigrationV9.md
```diff
@@ -0,0 +1,40 @@
...
+
+When text is spread across several nodes `<Text>Hello <Text>world</Text</Text>`, `@testing-library/react-native` used to allow matching the full string `getByText('Hello world')`. However this behaviour was different than the web one, and wouldn't always be straightforward to reason about (it could match text nodes far from each other on the screen). From v9, this type of match will not work. A work around is to use `within`:
+
+```tsx
+import {Text} from 'react-native'
+import {render, within} from '@testing-library/react-native'
+
+const {getByText} = render(<Text>Hello <Text>world</Text</Text>)
+
+// exact: false
```
thymikee, line 34:
```suggestion
```

website/docs/MigrationV9.md
```diff
@@ -0,0 +1,40 @@
...
+getByText('Hello World'); // Matches
+getByText('Hello'); // Doesn't match
+getByText('hello', { exact: false }); // ignore case-sensitivity and does partial matching
+```
+
+`findBy*` queries used to take a `waitForOptions` parameter (e.g. `findByText('Hello world', {timeout: 3000})`), that parameter has been moved to the third place: `findByText('Hello world', {exact: true}, {timeout: 3000})`. For backward compatibility, `@testing-library/react-native@9.0` still reads `waitForOptions` from the second parameter but will print a deprecation warning: `findByText('Hello world', {timeout: 3000, exact: false})`.
+
+## Stop matching across several nodes (breaking change)
+
+When text is spread across several nodes `<Text>Hello <Text>world</Text</Text>`, `@testing-library/react-native` used to allow matching the full string `getByText('Hello world')`. However this behaviour was different than the web one, and wouldn't always be straightforward to reason about (it could match text nodes far from each other on the screen). From v9, this type of match will not work. A work around is to use `within`:
```
thymikee, line 26:
```suggestion
When text is spread across several nodes, like: 
```tsx
<Text>Hello <Text>world</Text</Text>
```
`@testing-library/react-native` used to allow matching the full string using `getByText('Hello world')`. However this behaviour was different than the web one, and wouldn't always be straightforward to reason about. For instance it could match text nodes far from each other on the screen. From v9, this type of match will not work. A work around is to use `within`:
```

File: website/docs/MigrationV9.md [hunk 4]
```diff
@@ -0,0 +1,40 @@
...
+
+```jsx
+const { getByText } = render(<Text>Hello World</Text>);
+
+getByText('Hello World'); // Matches
+getByText('Hello'); // Doesn't match
+getByText('hello', { exact: false }); // ignore case-sensitivity and does partial matching
+```
+
+`findBy*` queries used to take a `waitForOptions` parameter (e.g. `findByText('Hello world', {timeout: 3000})`), that parameter has been moved to the third place: `findByText('Hello world', {exact: true}, {timeout: 3000})`. For backward compatibility, `@testing-library/react-native@9.0` still reads `waitForOptions` from the second parameter but will print a deprecation warning: `findByText('Hello world', {timeout: 3000, exact: false})`.
```
thymikee, line 22:
```suggestion
The `findBy*` queries used to take a `waitForOptions` parameter as a second argument, e.g. `findByText('Hello world', {timeout: 3000})`. It has now been moved to the third argument: `findByText('Hello world', {exact: true}, {timeout: 3000})`. For backward compatibility, `@testing-library/react-native@9.0` can still read `waitForOptions` from the second argument but will print a deprecation warning.
```

thymikee (approved):
LGTM 👍🏼

PR #881, opened 2021-12-06
feat(breaking): migrate Flow to TS; remove some deprecated functions and aliases

### LINKED ISSUE #877: Native TS support
## Describe the Feature
As quickly discussed in a [PR comment](https://github.com/callstack/react-native-testing-library/pull/554#issuecomment-976719551), I'll see to port the library to use TS instead of flow. Unsure if it's needed to list the benefits of the move, but here are some of them:
- TS is nowadays more common than flow, with native IDE support, which makes contributing easier with TS
- Flow is _not really_ an open source project, the [flow team stated](https://medium.com/flow-type/clarity-on-flows-direction-and-open-source-engagement-e721a4eb4d8b) that their focus is Facebook's codebase, and that community isn't.
- Callstack seems to use TS now for most of its JS based tooling, so consistency?

First question is what about Flow support? Personally I don't care, and I'm not sure it's a big priority for the community. So what I'd suggest is for us to publish the typings to flow-typed before removing flow support. That'll make long term maintenance simpler, but that'd be a breaking change. I personally would like to first release the v9 before this release, so we'd talk about a v10.

If we're considering a breaking change, I'd suggest to use that as an opportunity to clean up the deprecated APIs (from both v2, v7, v9) features, such `flushMicrotasksQueue`, etc.

We can also of course keep the flow declaration in the same repo for now, and don't clean up anything. It'll make the TS migration a bit more complicated, but that's not the end of the world.

What do you think of it?

AugustinLF:
This is the first draft of a PR to switch from flow to TS. The PR includes a green build for typescript and tests. It also includes the deletion of several deprecated APIs as discussed in #877.

There is still a bit left to do, but I'd like to have to have a 👍  before fixing those:
- [x] flow types (like we used to do with TS)
- [x] Build set up: I'm unsure what strategy we'd like to use to publish a typeless build + expose the TS types, if not I'll try to follow repack's one

Removed functionality:
- deprecation messages for all fns deprecated in v2 (getByName, getByType, etc)
- `waitForElement`
- `shallow`
- `flushMicrotasksQueue`

Closes #877


babel.config.js
```diff
@@ -12,5 +12,8 @@ module.exports = {
       },
     ],
   ],
-  plugins: ['@babel/plugin-proposal-class-properties'],
+  plugins: [
+    '@babel/plugin-proposal-class-properties',
+    '@babel/plugin-transform-flow-strip-types',
```
AugustinLF, line 17: I've been struggling with this issue https://github.com/react-native-community/upgrade-support/issues/152. There is probably a cleaner way to fix that, would love to get some guidance

src/__tests__/a11yAPI.test.tsx
```diff
@@ -29,11 +28,11 @@ const waitForOptions = { timeout: 10 };
 class Button extends React.Component<any> {
   render() {
     return (
-      // $FlowFixMe - accessibilityStates removed in RN 0.62
       <TouchableOpacity
         accessibilityHint={BUTTON_HINT}
         accessibilityLabel={BUTTON_LABEL}
         accessibilityRole="button"
+        // @ts-ignore - accessibilityStates removed in RN 0.62
```
AugustinLF, line 35: Should we consider dropping support for that? Just curious

src/__tests__/a11yAPI.test.tsx
```diff
@@ -220,14 +219,15 @@ test.skip('getByA11yStates, queryByA11yStates', () => {
...
-  expect(disabled && disabled.props.accessibilityStates).toMatchObject([
+  expect(disabled?.props.accessibilityStates).toMatchObject([
     'selected',
     'disabled',
   ]);
   const disabledSelected = queryByA11yStates(['selected', 'disabled']);
-  expect(
-    disabledSelected && disabledSelected.props.accessibilityStates
-  ).toEqual(['selected', 'disabled']);
+  expect(disabledSelected?.props.accessibilityStates).toEqual([
```
AugustinLF, line 227: It runs TS linter rules now, is it something we want? Perhaps we need to make sure that this doesn't have more impact (some JS rules disappearing, or something like that?)

src/flushMicroTasks.js
```diff
@@ -1,23 +0,0 @@
-// @flow
-import { printDeprecationWarning } from './helpers/errors';
-import { setImmediate } from './helpers/timers';
-
-type Thenable<T> = { then: (() => T) => mixed };
-
-/**
- * Wait for microtasks queue to flush
- */
-export default function flushMicrotasksQueue<T>(): Thenable<T> {
```
AugustinLF, line 10: Deleted this one

src/helpers/a11yAPI.ts
```diff
@@ -1,85 +1,157 @@
...
-  getAllByLabelText: (string | RegExp) => GetAllReturn,
-  queryByA11yLabel: (string | RegExp) => QueryReturn,
-  queryByLabelText: (string | RegExp) => QueryReturn,
-  queryAllByA11yLabel: (string | RegExp) => QueryAllReturn,
-  queryAllByLabelText: (string | RegExp) => QueryAllReturn,
-  findByA11yLabel: (string | RegExp, ?WaitForOptions) => FindReturn,
-  findByLabelText: (string | RegExp, ?WaitForOptions) => FindReturn,
-  findAllByA11yLabel: (string | RegExp, ?WaitForOptions) => FindAllReturn,
-  findAllByLabelText: (string | RegExp, ?WaitForOptions) => FindAllReturn,
+  getByA11yLabel: (label: string | RegExp) => GetReturn;
```
AugustinLF, line 23: Should we consider deprecating the aliases?

src/helpers/byDisplayValue.ts
```diff
@@ -23,7 +23,10 @@ const getTextInputNodeByDisplayValue = (
       matches(value, nodeValue, normalizer, exact)
     );
   } catch (error) {
-    throw createLibraryNotSupportedError(error);
+    if (error instanceof Error) {
```
AugustinLF, line 26: Common theme in this PR. `catch` provides `error: unknown` in recent TS, so I've been trying to fix the code around, feel free to chime in and let me know if you have different opinion on the workarounds

src/helpers/debugDeep.ts
```diff
@@ -0,0 +1,21 @@
+import type { ReactTestRendererJSON } from 'react-test-renderer';
+import format from './format';
+
+/**
+ * Log pretty-printed deep test component instance
+ */
+export default function debugDeep(
+  instance: ReactTestRendererJSON | ReactTestRendererJSON[],
```
AugustinLF, line 8: Evolved to handle an array, to make the call-sites happy

src/helpers/getByAPI.ts
```diff
@@ -16,58 +11,50 @@ import {
...
+  ) => Array<ReactTestInstance>;
   getAllByDisplayValue: (
     value: string | RegExp,
     queryOptions?: TextMatchOptions
-  ) => Array<ReactTestInstance>,
+  ) => Array<ReactTestInstance>;
 
   // Unsafe aliases
-  UNSAFE_getByType: <P>(type: React.ComponentType<P>) => ReactTestInstance,
+  UNSAFE_getByType: <P>(type: React.ComponentType<P>) => ReactTestInstance;
```
AugustinLF, line 49: Should we also deprecated those?

src/__tests__/a11yAPI.test.tsx
pierrezimmermannbam: maybe try accessibilityState={{ selected: true }}

AugustinLF: We do support `accessibilityState` (see [below](https://github.com/callstack/react-native-testing-library/pull/881/files/0ad5e48dc9d1fbaa8bb382708a93dbb7dbef68d2#diff-7a8cb5b84e76140224edf2c69459eef7246f003a6ddd3b67b03660f86dc68681R62) for instance, my question is that since RN 0.62 doesn't support `accessibilityStates` anymore, should we support it? And if yes, for how long?

.circleci/config.yml
```diff
@@ -34,21 +34,21 @@ jobs:
...
+  lint-and-typescript:
     <<: *defaults
     steps:
       - attach_workspace:
           at: ~/react-native-testing-library
       - run: |
           yarn lint
-          yarn flow-check
-  typescript:
+          yarn typecheck
```
mdjastrzebski, line 44:
```suggestion
          yarn typescript-check
```
This needs to match package.json script name.

src/__tests__/a11yAPI.test.tsx
```diff
@@ -79,7 +78,7 @@ test('getByA11yLabel, queryByA11yLabel, findByA11yLabel', async () => {
     BUTTON_LABEL
   );
   const button = queryByA11yLabel(/button/g);
-  expect(button && button.props.accessibilityLabel).toEqual(BUTTON_LABEL);
+  expect(button?.props.accessibilityLabel).toEqual(BUTTON_LABEL);
```
mdjastrzebski, line 81: optional chaining 👍🏻

src/helpers/makeA11yQuery.ts
```diff
@@ -18,29 +18,33 @@ function makeAliases(aliases: Array<string>, query: Function) {
...
       return instance.find(
         (node) => isNodeValid(node) && matcherFn(node.props[name], matcher)
       );
     } catch (error) {
       throw new ErrorWithStack(
-        prepareErrorMessage(error, name, matcher),
+        prepareErrorMessage(
+          error instanceof Error
+            ? error
+            : new Error('Something wrong happened'),
```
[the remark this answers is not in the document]
mdjastrzebski: I'd suggest moving non-`Error` `error` handling inside the `preapareErrorMessage` so that it accepts `unknown` and then makes a check inside. As for other potential received values we could just call `error?.toString()` to get their printable value. @AugustinLF wdyt?

src/helpers/getByAPI.ts
```diff
@@ -87,7 +80,14 @@ export const UNSAFE_getByProps = (
     try {
       return instance.findByProps(props);
     } catch (error) {
-      throw new ErrorWithStack(prepareErrorMessage(error), getByPropsFn);
+      if (error instanceof Error) {
```
mdjastrzebski, line 83: I'd suggest make `prepareErrorMessage` accept `unknown` as first argument, and move the non-`Error` handling there.

src/helpers/errors.ts
```diff
@@ -32,21 +31,30 @@ export const prepareErrorMessage = (
...
+): null => {
+  if (error instanceof Error) {
+    if (error.message.includes('No instances found')) {
+      return null;
+    }
+    throw new ErrorWithStack(error.message, callsite);
   }
-  throw new ErrorWithStack(error.message, callsite);
+
+  throw new ErrorWithStack('Something wrong happened when querying', callsite);
```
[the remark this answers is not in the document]
mdjastrzebski: Consider something along the lines:

```suggestion
  throw new ErrorWithStack(`Query: caught unknown error type: ${typeof error}, value: ${error?.toString()}`, callsite);
```

src/helpers/errors.ts
```diff
@@ -17,8 +16,8 @@ export const createLibraryNotSupportedError = (error: Error): Error =>
 
 export const prepareErrorMessage = (
   error: Error,
-  name: ?string,
-  value: ?mixed
+  name?: string | undefined,
+  value?: unknown | undefined
```
mdjastrzebski, line 20: AFAIR optional argument implies `undefined`

```suggestion
  name?: string,
  value?: unknown
```

src/helpers/debugShallow.ts
```diff
@@ -1,20 +1,22 @@
...
  */
 export default function debugShallow(
-  instance: ReactTestInstance | React.Element<any>,
+  instance: ReactTestInstance | React.ReactElement<any>,
   message?: string
 ) {
   const { output } = shallowInternal(instance);
 
   if (message) {
+    // eslint-disable-next-line no-console
```
mdjastrzebski, line 16: Since were library that sometimes need to communicate with user, maybe we could disable `no-console` rule in ESLint config. @thymikee wdyt?

src/helpers/debugDeep.ts
```diff
@@ -0,0 +1,21 @@
...
+  message?: string
+) {
+  const instances = Array.isArray(instance) ? instance : [instance];
```
mdjastrzebski, line 11: I wonder if `format` (and underlying `prettyFormat`) could actually handle array version. If so then maybe we should just pass it there. wdyt?

src/helpers/byText.ts
```diff
@@ -59,7 +62,10 @@ const getNodeByText = (
     }
     return false;
   } catch (error) {
-    throw createLibraryNotSupportedError(error);
+    if (error instanceof Error) {
+      throw createLibraryNotSupportedError(error);
```
mdjastrzebski, line 66: IMO we should also throw for non-`Error` errors. I'd suggest changing `createLibraryNotSupportedError` to accept `unknown` instead of `Error` and then do some checks inside.

mdjastrzebski: This relates to all similar cases in al of the query files.

src/helpers/byDisplayValue.ts
mdjastrzebski: We should not silently ignore the non-`Error` values. It should be better to just analyse the `error` type inside the `createLibraryNotSupportedError`.

src/helpers/a11yAPI.ts
mdjastrzebski: I guess I would keep them for the time being, as not all a11y queries have their RTL-like counterpart.
@thymikee wdyt?

src/fireEvent.ts
```diff
@@ -53,7 +53,7 @@ const findEventHandler = (
   eventName: string,
   callsite?: any,
   nearestTouchResponder?: ReactTestInstance
-) => {
+): Function | null => {
```
mdjastrzebski, line 56: Maybe we should define a type for event handlers?. We could start with non-generic version like `type EventHandler = (...args: any) => unknown`.

src/__tests__/questionsBoard.test.tsx
```diff
@@ -9,7 +8,13 @@ import {
 } from 'react-native';
 import { render, fireEvent } from '..';
 
-function QuestionsBoard({ questions, onSubmit }) {
+function QuestionsBoard({
```
mdjastrzebski, line 11: Consider extracting `QuestionsBoardProps` type

src/__tests__/fireEvent.test.tsx
```diff
@@ -11,33 +10,49 @@ import {
 } from 'react-native';
 import { render, fireEvent } from '..';
 
-const OnPressComponent = ({ onPress, text }) => (
+const OnPressComponent = ({
```
mdjastrzebski, line 13: Consider extracting `XxxProps` types for all of these components, pls place them next to each component.

mdjastrzebski (commented):
Great work @AugustinLF! I strongly support moving from Flow to Typescript as a better and de-facto standard type-checking solution.

src/helpers/getByAPI.ts
AugustinLF: Yep, I've actually started doing that in other functions, I'll switch to this approach everywhere

src/helpers/debugDeep.ts
AugustinLF: `prettyFormat` takes `unknown` as a parameter, so will follow your suggestion

src/helpers/a11yAPI.ts
AugustinLF: In my ideal world each query should exists only once (no alias) and have the same behaviour as on web. For instance, we could drop `getByA11yLabel` and keep `getByLabelText`, since this one is supported on both platforms.

thymikee: Let's get rid of aliases for queries that are available on RTL, like ByLabelText for example. It's a simple search-replace change, so no biggie

typings/test_index.js
```diff
@@ -11,15 +10,15 @@ import {
   within,
   getQueriesForElement,
   getDefaultNormalizer,
-} from '../..';
+} from '@testing-library/react-native';
```
AugustinLF, line 13: @thymikee you suggested to export flow types the same way we did with TS, but I can't get the import here to work (hence the red build), are you familiar with how to do that?

src/helpers/a11yAPI.ts
AugustinLF: Removed the `*LabelText` and `*Role` aliases.

babel.config.js
thymikee: right, we could configure this through `env: {test: {...}}`

AugustinLF: Yep, good call, fixed in fa9a1c594fff47283a7a34cab38380f065b337ec

AugustinLF:
@thymikee I just rebased to fix the conflict, and I believe I addressed all the comments (only left one is mine https://github.com/callstack/react-native-testing-library/pull/881#discussion_r773872445 waiting for your input). Is there anything else you'd like me to do?

thymikee:
@AugustinLF I see that Flow check is failing, would be cool to remove it :) Other than that please just wait a bit more for me to find time and review this. Thanks so much for the work you've done already!

AugustinLF:
Yeah, the issue with the flow check is that I don't know how to make it possible to "test" the flow types https://github.com/callstack/react-native-testing-library/pull/881#discussion_r773872445. If I remove the test file for the types the build pass, but then we have no guarantee that the flow types are correct.

Thanks for taking the time reviewing :)

src/__tests__/a11yAPI.test.tsx
Killavus: Since the oldest React-Native version mentioned in docs is `>= 0.61` and I see no clear "minimum supported React Native version" info, I'd assume people can be using RNTL in 0.61~ versions where `accessibilityStates` prop is still around. Thus I'd keep it if it does not incur big maintenance cost.

src/__tests__/a11yAPI.test.tsx
```diff
@@ -70,51 +69,51 @@ function Section() {
   );
 }
 
-test('getByA11yLabel, queryByA11yLabel, findByA11yLabel', async () => {
-  const { getByA11yLabel, queryByA11yLabel, findByA11yLabel } = render(
+test('getByLabelText, queryByLabelText, findByLabelText', async () => {
+  const { getByLabelText, queryByLabelText, findByLabelText } = render(
```
Killavus, line 73: I wonder why we are changing alias for less `a11y` name despite it is `a11yAPI` test. Is there any argument for it or only preference?

src/__tests__/a11yAPI.test.tsx
Killavus: Since it extends `callstack` ESLint preset I think it's more of a concern of `@callstack/eslint-config` to maintain parity between the two? I've quickly checked and I don't see glaring differences between JS/TS configs. @thymikee do you think it may be a problem?

src/__tests__/a11yAPI.test.tsx
```diff
@@ -396,31 +392,6 @@ test('getAllByA11yValue, queryAllByA11yValue, findAllByA11yValue', async () => {
...
-  } = render(<Section />);
-
-  // Assert that query aliases are referencing the same function
-  expect(getByA11yLabel).toBe(getByLabelText);
-  expect(queryByA11yLabel).toBe(queryByLabelText);
-  expect(findByA11yLabel).toBe(findByLabelText);
-  expect(getAllByA11yLabel).toBe(getAllByLabelText);
-  expect(queryAllByA11yLabel).toBe(queryAllByLabelText);
-  expect(findAllByA11yLabel).toBe(findAllByLabelText);
-});
```
Killavus, line 422: While we may discuss about usefulness of this test, why it gets removed with this PR? :)

src/__tests__/act.test.tsx
```diff
@@ -1,12 +1,12 @@
-// @flow
 import * as React from 'react';
 import { Text } from 'react-native';
 import ReactTestRenderer from 'react-test-renderer';
 import act from '../act';
 import render from '../render';
 import fireEvent from '../fireEvent';
 
-const UseEffect = ({ callback }: { callback: Function }) => {
+type UseEffectProps = { callback: () => void };
```
Killavus, line 8: small improvement (?):

```suggestion
type UseEffectProps = { callback(): void };
```

src/__tests__/auto-cleanup-skip.test.tsx
```diff
@@ -11,7 +10,7 @@ beforeAll(() => {
 
 let isMounted = false;
 
-class Test extends React.Component<*> {
+class Test extends React.Component<any> {
```
Killavus, line 13:
```suggestion
class Test extends React.Component {
```

src/__tests__/auto-cleanup.test.tsx
```diff
@@ -1,11 +1,10 @@
-// @flow
 import * as React from 'react';
 import { View } from 'react-native';
 import { render } from '..';
 
 let isMounted = false;
 
-class Test extends React.Component<*> {
+class Test extends React.Component<{ onUnmount?: () => void }> {
```
Killavus, line 7: Here you can also do [the same](https://github.com/callstack/react-native-testing-library/pull/881/files#r792167435) treatment `onUnmount?(): void` or keep it as-is

src/__tests__/byText.test.tsx
```diff
@@ -23,9 +26,7 @@ const Banana = () => {
       <Text>Is the banana fresh?</Text>
       <Text testID="bananaFresh">not fresh</Text>
 
-      <MyButton onPress={jest.fn()} type="primary">
-        Change freshness!
-      </MyButton>
+      <MyButton onPress={jest.fn()}>Change freshness!</MyButton>
```
Killavus, line 29: Why `type="primary"` is gone here?

src/__tests__/byText.test.tsx
```diff
@@ -146,7 +148,8 @@ describe('findBy options deprecations', () => {
 });
 
 test.skip('getByText works properly with custom text component', () => {
-  function BoldText({ children }) {
+  type BoldTextProps = { children: React.ReactNode };
```
Killavus, line 151: How about extracting this to higher scope:

```
type ChildrenProps = { children: React.ReactNode }
```

And then reuse for all tests expecting `children`?

src/__tests__/timerUtils.ts
```diff
@@ -1,11 +1,9 @@
-// @flow
-
 import { setTimeout } from '../helpers/timers';
 
 const TimerMode = {
   Legacy: 'legacy',
   Modern: 'modern', // broken for now
```
Killavus, line 5: Still broken? :)

src/flushMicroTasks.ts
```diff
@@ -0,0 +1,13 @@
+import { setImmediate } from './helpers/timers';
+
+type Thenable<T> = { then: (callback: () => T) => unknown };
+
+export function flushMicroTasks<T>(): Thenable<T> {
+  return {
+    // using "thenable" instead of a Promise, because otherwise it breaks when
+    // using "modern" fake timers
```
Killavus, line 8: Can you check if it's still the case with TypeScript bindings for `jest`?

src/helpers/byDisplayValue.ts
```diff
@@ -7,9 +7,9 @@ import { createLibraryNotSupportedError } from './errors';
 import type { TextMatchOptions } from './byText';
 
 const getTextInputNodeByDisplayValue = (
-  node,
-  value,
-  options?: TextMatchOptions = {}
+  node: ReactTestInstance,
+  value: string | RegExp,
```
Killavus, line 11: I think this particular union may deserve an alias because of its widespread use. WDYT @thymikee?

src/helpers/byText.ts
```diff
@@ -8,12 +8,15 @@ import { filterNodeByType } from './filterNodeByType';
...
-  exact?: boolean,
-  normalizer?: NormalizerFn,
+  exact?: boolean;
+  normalizer?: NormalizerFn;
 };
 
-const getChildrenAsText = (children, TextComponent) => {
-  const textContent = [];
+const getChildrenAsText = (
+  children: (string | number | React.ReactElement)[],
```
Killavus, line 16:
```suggestion
  children: React.ReactChild[],
```

src/helpers/errors.ts
```diff
@@ -0,0 +1,106 @@
...
+export const createLibraryNotSupportedError = (error: unknown): Error =>
+  new Error(
+    `Currently the only supported library to search by text is "react-native".\n\n${
+      error instanceof Error ? error.message : ''
+    }`
+  );
+
+export const prepareErrorMessage = (
+  // TS states that error caught in a catch close are of type `unknown`
+  // most real cases will be `Error`, but better safe than sorry
```
Killavus, line 21: 💯

src/helpers/errors.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (error.message.includes('No instances found')) {
+      return null;
+    }
+    throw new ErrorWithStack(error.message, callsite);
+  }
+
+  throw new ErrorWithStack(
+    // generic refining of `unknown` is very hard, you cannot do `'toString' in error` or anything like that
+    // Converting as any with extra safe optional chaining will do the job just as well
+    `Query: caught unknown error type: ${typeof error}, value: ${(error as any)?.toString?.()}`,
```
Killavus, line 59: Maybe it's nice to introduce some marker value using `??` operator like ` ?? "SHOULD_NEVER_HAPPEN"` so when someone comes to issues with this it'll be immediately greppable through the codebase? @thymikee

src/helpers/timers.ts
```diff
@@ -30,20 +28,21 @@ function getJestFakeTimersType() {
   }
 
   if (
+    // @ts-ignore jest mutates setTimeout
```
Killavus, line 31: I think it's a good place to use `@ts-expect-error` instead of `@ts-ignore`.

src/waitFor.ts
```diff
@@ -203,24 +200,10 @@ export default async function waitFor<T>(
...
   let result: T;
 
-  //$FlowFixMe: `act` has incorrect flow typing
   await act(async () => {
     result = await waitForInternal(expectation, optionsWithStackTrace);
   });
 
-  //$FlowFixMe: either we have result or `waitFor` threw error
+  // @ts-ignore either we have result or `waitFor` threw error
   return result;
```
Killavus, line 208: I think you can use `!` operator to force TS to trust you it's always valid. You can keep the comment though.

Killavus (commented):
Looks mostly good to me. I'd definitely take a look whether we can get rid of `Thenable` in https://github.com/callstack/react-native-testing-library/pull/881/files#r792187961 `flushMicroTasks` after migration from Flow to TypeScript.

Also `string | RegExp` type may be aliased to something more in `testing-library` nomenclature. The rest of my comments are small stylistic improvements & small type improvements (like using `React.ReactChild` instead of enumerating possibilities, or restricting `any` in some places).

src/__tests__/a11yAPI.test.tsx
AugustinLF: I guess I should actually rename the test. It was discussed a bit earlier, the point was to remove the aliases, currently there was `ByA11yLabel`, `ByAccessibilityLabel`, `ByLabelText`. RTL web only supports `ByLabelText`, so here we would align the API.

src/__tests__/a11yAPI.test.tsx
AugustinLF: This test just checks that aliases work and we removed them

https://github.com/callstack/react-native-testing-library/pull/881#discussion_r768432588

src/__tests__/byText.test.tsx
AugustinLF: The prop wasn't used, the `MyButton` component doesn't have a `type` prop, so TS would complain. I can add it there, but then it'd probably be ignored? My guess was that it'd be simpler to just remove it since it didn't seem to have any impact on the test.

src/__tests__/timerUtils.ts
AugustinLF: No idea to be honest, I'm not super familiar with this modern/legacy timer. But I guess it'd be worth investigating after this PR!

src/helpers/errors.ts
AugustinLF: Please let me know if you confirm that and I'll add that!

AugustinLF:
@Killavus I should have addressed all your comments (or answered). I just let one hanging, waiting for a confirmation. Please let me know if you need anything else!

AugustinLF:
@Killavus can I be of any help, or do anything to help getting that merged?

src/helpers/debugShallow.ts
thymikee: Nah, I'd rather create a dummy logger helper and disable the rule in one place. It's still possible to leave some debug code using console.logs which would be super annoying if happens

src/helpers/errors.ts
thymikee: I don't mind, but can be done separately

src/helpers/getByAPI.ts
thymikee: No, let's keep them.

thymikee (approved):
Looks good and ready to be merged. However, since this is a breaking change, I wonder if we shouldn't tackle https://github.com/callstack/react-native-testing-library/pull/934 first and release it in a minor release. WDYT?

AugustinLF:
Sounds fair to me!

thymikee:
@AugustinLF v9.1.0 is released so let's rebase this and we can merge this work 🚀

AugustinLF:
@thymikee rebased. However CI didn't not thing to have been triggered, did you change anything recently?

thymikee:
@AugustinLF nothing else than merging a bunch of Dependabot PRs

thymikee:
There's some outage around self-hosted runners, may be related: https://status.circleci.com

src/helpers/timers.ts
```diff
@@ -2,6 +2,8 @@
 /* globals jest */
 const globalObj = typeof window === 'undefined' ? global : window;
 
+// trigger pipeline
+
```
thymikee, line 6: looks like it worked this time. I wholeheartedly recommend using `git commit --allow-empty` for such cases when playing with CI (we all love it, don't we?) :D 
```suggestion
```

thymikee:
Thank you @AugustinLF!

src/helpers/timers.ts
AugustinLF: Aha didn't think about that, was planning to remove the commit and push force^^

PR #883, opened 2021-12-17
fix: add React.Fragment support for matching text

Alvindelito:
#### This PR builds upon #663

I have used the code provided by @lachiet with an added conditional.

Searching for text by using `getByText()` should now detect the text inside of a `<React.Fragment>` nested in a `<Text>` component.
```javascript
<Text>
  <React.Fragment>Hello world</React.Fragment>
</Text>
```

### Test plan

I have included a test to test the new change and all other tests have passed.

### DIFF
--- a/src/__tests__/byText.test.js
+++ b/src/__tests__/byText.test.js
@@ -176,6 +176,20 @@ test.skip('getByText works properly with custom text container', () => {
   ).toBeTruthy();
 });
 
+test('getByText works properly with React.Fragment nested in <Text>', () => {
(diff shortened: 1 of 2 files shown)


thymikee:
Thanks for your contribution @Alvindelito! Can you test if this works with nested fragments as well? Checking for `parent.parent` proved to be buggy in the past. Also, have you considered updating the `getChildrenAsText` method to check for `Fragment` type? We could use `react-is` for this.

thymikee:
Closed via #934

PR #885, opened 2021-12-27
feat: adds onTimeout option to waitFor

jmbeach:
### Summary

Adds the option `onTimeout?: (error: Error) => Error` to the `waitFor` function to match the option in [the non-native React Testing Library](https://testing-library.com/docs/dom-testing-library/api-async/).

### Test plan

I added a new test for this as well.

Useful so that you can do stuff like this:

```ts
await waitFor(() => expect(getSlotMachine().props.accessibilityState.disabled).toBeFalsy(), {
    onTimeout: (e) => {
      console.log(getSlotMachine().props.accessibilityState.disabled)
      expect(wrapper).toMatchSnapshot();
      return e;
    }
  });
```

### DIFF
--- a/src/__tests__/waitFor.test.js
+++ b/src/__tests__/waitFor.test.js
@@ -115,6 +115,32 @@ test.each([TimerMode.Legacy, TimerMode.Modern])(
   }
 );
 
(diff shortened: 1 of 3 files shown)


thymikee (approved):
Looks good, thanks!

PR #906, opened 2022-01-22
Trigger non-touch events on box-none targets

### LINKED ISSUE #897: Setting pointerEvents to "box-none" prevents triggering onTouch* and onLayout events on the element
## Describe the bug

Setting `pointerEvents: 'box-none'` on a `View` prohibits triggering `onTouch*` or `onLayout` events on the `View`. This behavior differs from React Native itself. This [Expo Snack](https://snack.expo.dev/@davidcalhoun/pointerevents-and-ontouch-events) showcases that pressing the child `Pressable` of a `<View pointerEvents="box-none" />` still triggers the `onTouchStart` and `onTouchEnd` events. This change in behavior, which appears to differ from React Native, began in with `7.2.0` due to changes in https://github.com/callstack/react-native-testing-library/pull/655.

## Expected behavior

Invoking `fireEvent.trigger(viewWithPointerEventsBoxNone, "onTouchStart")` results in triggering the `View`'s `onTouchStart` handler. 

Similarly, invoking `fireEvent.trigger(viewWithPointerEventsBoxNone, "onLayout")` results in triggering the `View`'s `onLayout` handler. 

## Steps to Reproduce

1. Press the "Press me" text in the [Expo Snack](https://snack.expo.dev/@davidcalhoun/pointerevents-and-ontouch-events) to observe the expected behavior. 
2. Clone the [reproduction case](https://github.com/dcalhoun/rntl-reproduction-cases/tree/pointerevents-box-none-and-ontouchevents), specifically the `pointerevents-box-none-and-ontouchevents` branch. 
3. `npm install`
4. `npm run test`

## Screenshots

n/a

## Versions

```
  npmPackages:
    @testing-library/react-native: ^9.0.0 => 9.0.0
    react: 17.0.2 => 17.0.2
    react-native: 0.66.4 => 0.66.4
    react-test-renderer: 17.0.2 => 17.0.2
```

dcalhoun:
### Summary
Fixes #897. Currently, this library disables _all_ events for a target with `pointerEvents="box-none"`. This behavior is counter to how React Native itself functions. This change continues to disable touch events, e.g. `press`, for targets set to "box-none", but allows triggering non-touch events, e.g. `layout`, `touchStart`.

### Test plan

Verify all CI tests pass, including the new tests within this pull request.


src/__tests__/fireEvent.test.js
```diff
@@ -317,6 +317,36 @@ test('should not fire on box-only pointerEvents View with nested elements', () =
   expect(handlePress).not.toHaveBeenCalled();
 });
 
+test('should fire non-terminal touch events on box-none pointerEvents View', () => {
```
dcalhoun, line 320: The name "non-terminal" was an attempt to communicate the subject even is not "press," with is the "ultimate" or "primary" touch event. Definitely open to feedback on better names.

src/fireEvent.js
```diff
@@ -33,12 +33,17 @@ const isPointerEventEnabled = (
   return isPointerEventEnabled(element.parent, true);
 };
 
+const isTouchEvent = (eventName?: string) => {
```
dcalhoun, line 36: The name `isTouchEvent` feels odd given that `onTouch*` is not considered to be `isTouchEvent` here. However, I chose the name `isTouchEvent` based upon the language in the [`pointerEvents` documentation](https://reactnative.dev/docs/view#pointerevents). Open to feedback on this naming.

src/fireEvent.js
```diff
@@ -33,12 +33,17 @@ const isPointerEventEnabled = (
...
+  return eventName === 'press';
```
dcalhoun, line 37: Originally, I structured this as `return ['press'].includes(eventName);` to allow easy additions to this allow list, but opted for a simple comparison for now.

src/fireEvent.js
Killavus: I'm not sure whether this list is complete and/or even can be a static one because it may be platform-dependent.

On Android setting `pointerEvents` works by intercepting and discarding `MotionEvent`s. These include  at least `touchstart`, `touchend`, `touchmove` and `touchcancel` events. See: https://github.com/facebook/react-native/blob/main/ReactAndroid/src/main/java/com/facebook/react/views/view/ReactViewGroup.java#L220-L245.

In case of iOS it is setting this: https://developer.apple.com/documentation/uikit/uiview/1622577-userinteractionenabled?language=objc (related RN code: https://github.com/facebook/react-native/blob/main/React/Views/RCTView.m#L151-L158) to `false`, so it includes `touch, press, keyboard, and focus events`.

I'd experiment with both platforms and see what superset of events is covering both platforms.

src/__tests__/fireEvent.test.js
Killavus: How about `pointer` events? While they are not mentioned directly in React-Native docs (which uses the term 'touch events') it is rather well-defined in the web version of the same concept: https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent

Killavus (commented):
@thymikee I think code-wise it looks good but I don't think it's right now doing the right thing semantically. As explained in https://github.com/callstack/react-native-testing-library/pull/906/files#r792240808 it may be needed to find a superset of events this option is supressing before merge.

src/fireEvent.js
Killavus: That seems like a good approach for me. @thymikee has a final say here though!

thymikee: Still haven't found time for this, please bare with me! :D

dcalhoun: Hey @thymikee. 👋🏻 I wanted to check back in with you. I am still interested merging this work to fix some tests in [WordPress/Gutenberg](https://github.com/wordpress/gutenberg). It is completely understandable if you do not have time to review. I merely wanted to make sure the PR was not forgotten. Thanks! 🙇🏻

thymikee: Hey @dcalhoun, I'll need more time to process it. Right now I focus on releasing a new library and needed to postpone this. It's not forgotten though.

In the meantime, I'll ask @AugustinLF, maybe he'll be able to check it.

dcalhoun: Thanks, @thymikee. That makes sense. I appreciate your time. 

P.S. - feel free to ignore my ping from a different PR.

AugustinLF: So if I understand correctly, your suggestion of implementation is strictly more correct than what we have, but is missing some additional cases? i.e. it makes things better, but doesn't fix everything, right?

If that's the case, I'm fine with merging this change. I'd like to have documentation (code and/or through a detailed issue) on the missing behaviour, so we can fix it going forward. Does that seem reasonable?

AugustinLF: I'd suggest to merge.

AugustinLF (approved):
As discussed in this thread https://github.com/callstack/react-native-testing-library/pull/906#discussion_r910126921, we should look to better document the platform difference to get a semantically more correct behaviour, but this pull request is still an improvement on the current status, so I'd say let's merge.

thymikee:
Ok, let's do this. We'll fit it into the next major release. 

Btw, please check the work that the core team is doing on the pointer events recently: https://github.com/facebook/react-native/search?q=pointer+events&type=commits. They're making commitments to have pointer events behave more web-like.

Issue #869, opened 2021-11-28
Implement `renderHook`


thymikee:
## Describe the Feature

See https://github.com/testing-library/react-testing-library/pull/991 for details

## Possible Implementations

Copy the implementation from RTL.

MattAgn:
Hi @thymikee! I'd be glad to help on this, I'll make a  PR shortly

pierrezimmermannbam:
Hi @thymikee, @MattAgn told me he finally couldn't find time for this so I made a pr

manugo-dev:
Hello there!

How about the waitForNextUpdate function ? Im trying to figured out how to upgrade my tests to be compatible with the new renderHook on the react-testing-library

<img width="533" alt="image" src="https://user-images.githubusercontent.com/17772376/164291396-747e3bc1-c646-411f-a6cb-cbd9e9ccbd10.png">

pierrezimmermannbam:
@manugo-dev  you should be able to use react-native-testing-library's waitfor instead. If I'm correct, being able to use the async utils from the library instead of utils such as waitForNextUpdate that are a bit too focused on implementation detailswas one of the reasons why it was decided to implement a renderHook equivalent on react-native-testing-library and on react-testing-library's side

mdjastrzebski:
Closing as implemented by #963.

kylebake:
> @manugo-dev you should be able to use react-native-testing-library's waitfor instead. If I'm correct, being able to use the async utils from the library instead of utils such as waitForNextUpdate that are a bit too focused on implementation detailswas one of the reasons why it was decided to implement a renderHook equivalent on react-native-testing-library and on react-testing-library's side

Am I correct in assuming that what you mean is there's no built in way to `waitFor` the return values of hooks directly anymore and we will need to always test the consumers of the hooks and have to `waitFor` elements to show/hide based on the values of the hook?

AugustinLF:
I believe what they mean is that you can simply use `waitFor` to wait for the value of the hook to change. Something like that:
```
import {waitFor, renderHook} from '@testing-library/react-native'
import myHook from './myHook'

test('myHook', async () => {
  const {result} = renderHook(() => myHook())
  await waitFor(() => {
    expect(result.current).toBe(1)
  })
})
```

kylebake:
[...]

PR #923, opened 2022-02-22
feat: add renderHook function

pierrezimmermannbam:
### Summary

This pr provides a basic implementation of renderHook as requested in #869. The solution is almost a copy past from the one made in react testing library

### Test plan

Some test cases were added, also inspired from the implementation from react testing library

### DIFF
--- a/src/__tests__/renderHook.test.tsx
+++ b/src/__tests__/renderHook.test.tsx
@@ -1,0 +1,62 @@
+import React, { ReactNode } from 'react';
+import { renderHook } from '../pure';
+
+test('gives comitted result', () => {
+  const { result } = renderHook(() => {
+    const [state, setState] = React.useState(1);
+
+    React.useEffect(() => {
+      setState(2);
+    }, []);
+
+    return [state, setState];
+  });
+
+  expect(result.current).toEqual([2, expect.any(Function)]);
+});
+
+test('allows rerendering', () => {
+  const { result, rerender } = renderHook(
+    (props: { branch: 'left' | 'right' }) => {
+      const [left, setLeft] = React.useState('left');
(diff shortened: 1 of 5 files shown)


AugustinLF:
The `@testing-library/react` [PR](https://github.com/testing-library/react-testing-library/pull/991) still hasn't been merged, so I'd like for it to be merged first before getting that in. I'm monitoring said PR, but don't hesitate to ping me once it's merged if I miss it.

mpeyper:
The RTL PR has been merged now. Please let me know if there is anything I can do to progress this.

AugustinLF (changes requested):
Implementation and tests look correct to me. Could you fix the conflicts and also add documentation for it in https://github.com/callstack/react-native-testing-library/blob/main/website/docs/API.md

pierrezimmermannbam:
I'm working on migrating to typescript, I'll add documentation afterwards, it shouldn't take long

pierrezimmermannbam:
It's done, there is a typing issue because the option param is optional so initialProps may not be defined when it needs to be but changing the typing would mean having a different api from the one of react-hooks-testing-library and I'm not sure it's something we want. 

I also added documentation, I used the documentation I found on the react-hooks-testing-library repo and added some examples

AugustinLF:
Oops forgot to ask for flow types https://github.com/callstack/react-native-testing-library/blob/main/typings/index.flow.js But after that we should be good I believe

pierrezimmermannbam:
@AugustinLF I added the flow types

AugustinLF (approved):

thymikee (approved):
Pushed some documentation updates. Looking pretty!

Issue #866, opened 2021-11-23
Can't getByText for component that returns a fragment


joebourne:
## Describe the bug

Using the `T` component from `@transifex/native`, I am unable to `getByText()` any child component of `<Text />` whose return is simply a fragment or string.

## Expected behavior

The inner text of the fragment is found by `getByText` and returned.

## Steps to Reproduce

```jsx
import React from 'react';
import { Text } from 'react-native';
import { render } from '@testing-library/react-native';

// This is mocking T from @transifex/react that has similar but more complex functionality
function T({ _str }) {
  return _str; // same if returning <>{_str}</>
}

function Test() {
  return (
    <Text>
      <T _str="My String" />
    </Text>
  );
}

describe('Test', () => {
  it('should render the correct text', () => {
    const { debug, getByText, toJSON } = render(<Test />);

    debug();

    console.log(toJSON());

    expect(getByText('My String')).toBeTruthy();
  });
});
```

Output of `console.log(toJSON())` is `{ type: 'Text', props: {}, children: [ 'My String' ] }` and output of `debug()` is 
```
<Text>
  My String
</Text> 
```

## Versions

npmPackages:
    @testing-library/react-native: ^8.0.0 => 8.0.0 
    react: ^16.13.1 => 16.14.0 
    react-native: ~0.63.4 => 0.63.4

thymikee:
Does it happen on `9.0.0-alpha.0`?

Joebourne:
Thanks for the quick reply. Just tried and still happens on `9.0.0-alpha.0`.

AugustinLF:
This should be fixable. There are two cases to consider here:
- `<Text><>Hey hey</></Text>`
- `<Text>Hey <>hey</></Text>`

The first case is the one you raised, and is probably _simpler_ to fix, and if we can, I don't see a reason why we shouldn't. The second is actually trickier, because on purpose we don't match across several react node, but it does make sense to consider that fragments should be ignored.

Joebourne:
@AugustinLF I found the PR (https://github.com/callstack/react-native-testing-library/pull/663) and it seems to be stale. Do you have any inside info on why it was dropped?

AugustinLF:
The author of the PR[ has just been pinged](https://github.com/callstack/react-native-testing-library/pull/663#discussion_r757337751), if they're not available anymore I guess we should be able to pursue the work on their PR.

Joebourne:
[...]

PR #934, opened 2022-03-01
feat: allow getByText and queryByText to find text nested in fragments

pierrezimmermannbam:
### Summary

This fixes the issue #866 

There are already two open prs for this issue, #663 and #883 but they seem to be stale. 

As suggested in this comment https://github.com/callstack/react-native-testing-library/pull/883#issuecomment-1008809986, this implementation modifies getChildrenAsText function and tests for Fragment type.

### Test plan

I added several test cases with fragments. It should work as if fragments were not there as far as getByText and queryByText are concerned

### DIFF
--- a/src/__tests__/byText.test.js
+++ b/src/__tests__/byText.test.js
@@ -436,3 +436,36 @@ describe('Supports normalization', () => {
     expect(getByText('A text', { normalizer: normalizerFn })).toBeTruthy();
   });
 });
+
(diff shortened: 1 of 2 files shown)


AugustinLF (approved):
Looks good to me, exhaustive tests and maintainable implementation. Thanks for removing the useless call to `getChildrenAsText`.

thymikee (approved):

PR #936, opened 2022-03-18
Update index.d.ts

Norfeldt:
Found this type to be much more useful and not throwing a type error if I do something like

```
expect(unlockedUtils.UNSAFE_getAllByType('Text').length).toBe(1)
```


Norfeldt:
Sorry, just realised that the right way of calling it is by

```js
import { Text } from 'react-native'

expect(unlockedUtils.UNSAFE_getAllByType(Text).length).toBe(1)
```

(thank you typescript CI)

PR #957, opened 2022-04-14
fix: missing jsx/tsx in build

carlosdp:
### Summary
Currently, `src/render.tsx` is not being built by Babel, because it was changes to a `tsx` extension, but Babel is only told to build `js` and `ts`.

Also, it seems the out directory is out of date in `package.json`.

### Test plan

It now builds everything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
   "name": "@testing-library/react-native",
   "version": "9.1.0",
   "description": "Simple and complete React Native testing utilities that encourage good testing practices.",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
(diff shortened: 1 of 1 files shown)


AugustinLF (approved):
Good catch.

AugustinLF:
@thymikee should we consider publishing the TS version? Will help to make sure that nothing else was missed.

thymikee (approved):
Thanks!

thymikee:
> should we consider publishing the TS version? Will help to make sure that nothing else was missed.

@AugustinLF I don't see it's needed, as it was a one-time thing. It would bloat the npm package and I'd like to avoid this if possible.

PR #963, opened 2022-04-27
docs(renderHook): remove references to error and all from renderHook API

mpeyper:
### Summary

Correct the `renderHook` API docs as `result.error` and `result.all` were not included in the new RTL version of `renderHook`.  I also fixed a typo and renamed a test example to better describe what it is showing (both in the same file).

### Test plan

No testing is required other than reviewing the documentation changes.

### DIFF
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -471,10 +471,10 @@ Useful function to help testing components that use hooks API. By default any `r
 Defined as:
 
 ```ts
-function renderHook(
-  callback: (props?: any) => any,
-  options?: RenderHookOptions
-): RenderHookResult;
+function renderHook<Result, Props>(
+  callback: (props?: Props) => Result,
+  options?: RenderHookOptions<Props>
+): RenderHookResult<Result, Props>;
 ```
 
 Renders a test component that will call the provided `callback`, including any hooks it calls, every time it renders. Returns [`RenderHookResult`](#renderhookresult-object) object, which you can interact with.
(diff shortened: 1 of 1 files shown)


website/docs/API.md
```diff
@@ -533,13 +533,11 @@ The `renderHook` function returns an object that has the following properties:
 
 ```jsx
 {
-  all: Array<any>
-  current: any,
-  error: Error
+  current: any
```
mdjastrzebski, line 536: Since `RenderHookResult` accepts generic `Result` parameter we should use it instead of `any` here.

website/docs/API.md
```diff
@@ -515,41 +515,43 @@ The `props` passed into the callback will be the `initialProps` provided in the
...
 
 #### `initialProps`
 
-The initial values to pass as `props` to the `callback` function of `renderHook`.
+The initial values to pass as `props` to the `callback` function of `renderHook`.  The `Props` type is determined by the type passed to or inferred by the `renderHook` call.
 
 #### `wrapper`
 
-A React component to wrap the test component in when rendering. This is usually used to add context providers from `React.createContext` for the hook to access with `useContext`. `initialProps` and props subsequently set by `rerender` will be provided to the wrapper.
+A React component to wrap the test component in when rendering. This is usually used to add context providers from `React.createContext` for the hook to access with `useContext`.
```
mpeyper, line 526: While fixing the types, I also noticed that this is not true for your implementation either.

website/docs/API.md
mpeyper: I hope my changes align with what you were thinking for this.

mdjastrzebski (approved):
Looks good! Thanks for applying type comments.

AugustinLF (approved):

PR #965, opened 2022-04-29
refactor: re-organise regular (non A11y) queries by predicate

mdjastrzebski:
### Summary

Re-organise internal queries code by predicate typ #325

### Scope of changes
- [x] Replace `GetByAPI`, `QueryByAPI` and `FindByAPI` types and functions with per query type and functions (`ByXxxQueries`, `bindByXxxQueries`)
- [x] Move query implementations from `src/helpers` to `src/queries`
- [x] Simplified query type definitions while making query options generic to allow greater flexibility

Note: Changes consist of moving types around and tweaking exports. In order to reduce PR size and potential for bugs I refrained from doing any query logic changes.

Note 2: This is the reason why I intend to make a separate PR for A11y queries as there we would require some code changes.

Note 3: There are certain inconsistencies when it comes to existence (or not) of certain deprecated query aliases. I did not make any changes but kept the existing status as is. Let me know if I should remove old aliases or fill in the missing ones

Note 4: We have really awesome automated tests!

### Test plan
- [x] All tests pass (no tests changed)
- [x] TypeScript and Flow checks pass


src/queries/unsafeProps.ts
```diff
@@ -0,0 +1,103 @@
...
+  UNSAFE_queryByProps: (props: {
+    [key: string]: any;
+  }) => ReactTestInstance | null;
+  UNSAFE_queryAllByProps: (props: {
+    [key: string]: any;
+  }) => Array<ReactTestInstance>;
+
+  // Removed aliases
+  // TODO: remove in the next release
+  queryByName: () => void;
```
MattAgn, line 72: this should be in the unsafeByName file I guess (same for the other mentions of byName in the rest of the file)

src/queries/placeholderText.ts
```diff
@@ -0,0 +1,105 @@
+import type { ReactTestInstance } from 'react-test-renderer';
```
MattAgn, line 1: I like the new simpler naming!

we should also rename the files in the __test__ folders, wdyt ?

MattAgn: maybe even make a queries subfolder in the test folder

MattAgn: totally! if we do that, shouldn't we also do it for the helpers? (to stay cohesive?)

src/queries/unsafeProps.ts
mdjastrzebski: This is unmodified name from before the PR and it's there for easier code migration for users.

That being said this code deprecation is quite old now, all users should have migrated a long time ago, so I would be for removing these legacy aliases. @thymikee @AugustinLF wdyt about remove all these deprecated aliases?

src/queries/placeholderText.ts
mdjastrzebski: moved queries tests to `src/queries/__tests__` directory.

src/queries/unsafeProps.ts
mdjastrzebski: I've removed all removed functions error messages.

AugustinLF (approved):
Looks good to me, glad that we're getting this in :)

thymikee (approved):
🚀

Issue #971, opened 2022-05-04
Allow providing custom queries


mdjastrzebski:
## Describe the Feature
Allow to add custom queries in `render` call. Similar to [RTL API](https://testing-library.com/docs/react-testing-library/api#queries).

## Possible Implementations
Option 1 (RTL-like):
* Provide a new `queries` option to `render` & `within` functions as in RTL,.
* Since RTL API requires you to import original `queries` if you want to retain them (and you generally want this), so we probably need to provider such export. 
* Correctly managing TS types might be the most difficult part of the task since we would want `render` and `within` output to be typed with the actual queries passed
* Implementation might consider changing way we internally export the queries from `bindXxxQueries` to named particular queries, e.g. `bindGetByText: (instance) => (text, options) => ReactTestInstance` or event introducing "unbound" queries (`getByText: (instance, text, options) => ReactTestInstance`) as in Dom Testing Library if that simplifies the exposed API / type management 
* Following RTL we should also export `makeQueries` internal utils, so that users need only to prepare `queryAll` variant of the query.

Option 2 (roll our own API):
* We could probably simplify the API surface by accepting list of additional queries only in `queryAllBy` verb and then invoking `makeQuery` internally.
* We could automatically forward all built-in queries and only append new queries from the user. That we we would keep the strong typing for built-in queries with potentially more lax typing of user provided queries (more usage `any` for query predicate / options).
* This options seems to have much better trade off between lower increasing code complexity & niche popularity of the feature. However it is not API compatible with RTL.

Option 3 (expose generic query function accepting user predicate):
* We could expose a full set of queries (`get`, etc) for `ByPredicate` query accepting a function `(instance: ReactTestInstance) => boolean`.
* User then could define his own predicates and feed them to our `getByPredicate`, etc queries. That would give user total flexibility in types of built queries: number of accepted params, typing, logic, etc
* These custom queries would do thing like filtering our composite components for the user, so he can focus on query logic.
* Pros: least increase of code & API surface complexity, proper typing for all queries
* Cons: not compatible with RTL, but seems so good that we can suggest RTL to add it ;-)
* Quick and dirty POC implementation of this option: #973 

## Related Issues
* Initial implementation attempt: #573 but the code is stale. Should be easy to re-apply. Some implementation and tests can be salved from that PR

mdjastrzebski:
@thymikee @AugustinLF @pvinis: what do you think?

V1shvesh:
[...]

PR #973, opened 2022-05-05
feat: predicate query

mdjastrzebski:
Implementation of custom queries based on option 3 from #971. This is different from custom query impl. in RTL, but because of the reasons explained in #971 I believe that their implementation has significant flaws, and we can provide a better API for such feature.

### Summary
* family of queries: `*ByPredicate(predicate: (instance: ReactTestInstance) => boolean)` 

### Test plan
* tests for matching proper elements
* tests for negative cases incl. error messages

Looking forward for your comments.

### DIFF
--- a/src/queries/predicate.ts
+++ b/src/queries/predicate.ts
@@ -1,4 +1,6 @@
 import type { ReactTestInstance } from 'react-test-renderer';
+import { isHostElement } from '../helpers/component-tree';
+import { findAll } from '../helpers/findAll';
 import { makeQueries } from './makeQueries';
 import type {
   FindAllByQuery,
@@ -8,24 +10,25 @@ import type {
   QueryAllByQuery,
   QueryByQuery,
 } from './makeQueries';
+import { CommonQueryOptions } from './options';
 
 type PredicateFn = (instance: ReactTestInstance) => boolean;
-// Not used so far
-type PredicateQueryOptions = void;
+type ByPredicateQueryOptions = CommonQueryOptions;
 
-const queryAllByPredicate = (
-  instance: ReactTestInstance
-): ((
-  predicate: PredicateFn,
-  options?: PredicateQueryOptions
-) => Array<ReactTestInstance>) =>
-  function queryAllByTestIdFn(predicate) {
-    const results = instance.findAll(
-      (node) => typeof node.type === 'string' && predicate(node)
+function queryAllByPredicate(instance: ReactTestInstance) {
+  return function queryAllByPredicateFn(
+    predicate: PredicateFn,
+    options?: ByPredicateQueryOptions
+  ): Array<ReactTestInstance> {
+    const results = findAll(
+      instance,
+      (node) => isHostElement(node) && predicate(node),
+      options
     );
 
     return results;
   };
+}
 
 const getMultipleError = (predicate: PredicateFn) =>
   `Found multiple elements matching predicate: ${predicate}`;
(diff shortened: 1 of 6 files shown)


src/queries/predicate.ts
```diff
@@ -0,0 +1,60 @@
...
+
+const queryAllByPredicate = (
+  instance: ReactTestInstance
+): ((
+  predicate: PredicateFn,
+  options?: PredicateQueryOptions
+) => Array<ReactTestInstance>) =>
+  function queryAllByTestIdFn(predicate) {
+    const results = instance.findAll(
+      (node) => typeof node.type === 'string' && predicate(node)
```
pierrezimmermannbam, line 24: we could use isHostElement method here

pierrezimmermannbam (changes requested):
Well we may not have the same definition for quick and dirty, looks rather clean to me ! What I think is missing for now is some documentation, an update of flow types and maybe an additional test case for within but otherwise the feature looks complete to me and could be released and iterated on later if needed

mdjastrzebski:
@pierrezimmermannbam yeah agree, it's pretty advanced POC 😂

I will try to revisit it soon, and add missing parts.

mdjastrzebski:
@pierrezimmermannbam I've rebased and updated the PR with missing PR. Looking forward for your review.

typings/index.flow.js
```diff
@@ -389,6 +389,37 @@ interface A11yAPI {
   ) => FindAllReturn;
 }
 
+type ByPredicateOptions = CommonQueryOptions;
+
+interface ByPredicateQueries {
+  getByPredicate: (
+    placeholder: TextMatch,
```
MattAgn, line 396:
```suggestion
      predicate: (element: ReactTestInstance) => boolean
```

website/docs/Queries.md
```diff
@@ -391,6 +392,23 @@ const element = screen.getByA11yValue({ now: 25 });
...
+  predicate: (element: ReactTestInstance) => boolean,
+  options?: {
+    includeHiddenElements?: boolean;
+  }
+): ReactTestInstance;
+```
+
+Returns a host element matching a custom `predicate` function.
+
+This query type is an escape hatch and should be used with care. In most cases you using the standard queries like `getByRole` or `getByText` will lead to test more-resembling user perspective. Use this query with care in rare cases where more flexibility is needed.
```
MattAgn, line 410: could we provide an example of a real scenario where we might need it and how to use it in that case? (like it's done for ByAccessibilityValue above)

typings/index.flow.js
mdjastrzebski: Good catch!

website/docs/Queries.md
MattAgn: i see what you mean, but then RNTL only provides queries for common case scenarios. We could find a legit scenario  that's very uncommon maybe? Like find all underline text? (so hard to find a rare but legit scenario haha ^^')

mdjastrzebski:
@thymikee @AugustinLF @MattAgn @pierrezimmermannbam feel free to express whether our users would actually benefit from this feature or it might be just unused piece of coda that we would have to maintain.

I treat this feature as an escape hatch for more advanced users who need more flexibility in querying items. We already have one escape hatch in form of *ByTestID queries. The fact that RTL does have custom queries (although in a bit different shape but functionally equivalent) does not mean that we need to add this as well if there is no valid use case for it. I have my own doubts here, so looking forward for your take.

website/docs/Queries.md
mdjastrzebski: Yeah, we can build examples to showcase how you can recreate some existing queries like *ByTestId.

I went back to the source, [DTL docs](https://testing-library.com/docs/dom-testing-library/api-custom-queries), to check what examples they give, but they are basically implementing `*ByTestId` queries using some `queryByAttribute` function. All that seems like legacy stuff that has no good examples.

AugustinLF:
I haven't had the need for it since we went for feature parity with RTL on accessibility related matchers.

MattAgn:
Same for me or other projects at my company, we don't remember having encountered such a scenario

mdjastrzebski:
Closing as not really needed. Having this API would allow extra flexibility for users, but we haven't really found any good example for that. I can such case appears in the future we can came back to this PR.

PR #976, opened 2022-05-10
docs: fix typo in redux integration

rvasseur31:
### Summary

We will test 'AddTodo' and not 'TodoElem'
Add '.test' to the second file as well as for 'AddTodo.test.js'

### Test plan

Not necessary as long as the change only concerns the documentation.


thymikee (approved):
Thank you!

PR #977, opened 2022-05-14
Organise a11y queries by predicate

MattAgn:
### Summary

This PR aims to continue solving [this issue](https://github.com/callstack/react-native-testing-library/issues/325) by using makeQueries for all a11yQueries.

Tasks:

- [x] group all a11y queries by their predicate type (byRole, byA11yState...) by using makeQueries instead of makeA11yQueries
- [x] split the tests for each query as well
- [x] remove the old makeA11yQueries helper

### Warning
- [x] The API has changed, I aligned the error messages to the same ones used for the other (non A11y) queries

### Test plan


MattAgn:
Hi @mdjastrzebski, I'm mostly done with this PR, just got a couple questions for you: 
- What do you think about extracting the helpers matchString and matchObject in the `helpers` folder? We use them respectively 3 and 2 times
- Now that we've split all the tests in their separate files, I think it could be nice to split them a little bit more by use case scenarios. I've done an example [here](https://github.com/callstack/react-native-testing-library/pull/977/commits/424d4052994f1ecc5fcdc91f0e4a13ff768a0d25). I think it makes the API and the tests clearer, what do you think?

src/queries/a11yHint.ts
```diff
@@ -0,0 +1,68 @@
...
+export type ByA11yHintQueries = {
+  getByA11yHint: GetByQuery<TextMatch>;
+  getAllByA11yHint: GetAllByQuery<TextMatch>;
+  queryByA11yHint: QueryByQuery<TextMatch>;
+  queryAllByA11yHint: QueryAllByQuery<TextMatch>;
+  findByA11yHint: FindByQuery<TextMatch>;
+  findAllByA11yHint: FindAllByQuery<TextMatch>;
+};
+
+export const bindByA11yHintQueries = (
```
mdjastrzebski, line 59: binder misses the aliased names `getByHintText`, `getByAccessibilityHint`. Please add them as well. 

It would be a good idea to have `getByA11yHint` === `getByAccessibilityHint`. You can easily achieve that by using intermediate variables, e.g.:
```
const getByHintText = getBy(instance);
//..

return {
  getByHintText,
  getByA11yHint: getByHintText,
  getByAccessibilityHint: getByHintText,
  // etc...
}
```

src/queries/a11yState.ts
```diff
@@ -0,0 +1,68 @@
...
+    a11yState
+  )}`;
+
+const { getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy } = makeQueries(
+  queryAllByA11yState,
+  getMissingError,
+  getMultipleError
+);
+
+export type ByA11yStateQueries = {
```
mdjastrzebski, line 50: Add alias names `getByAccessiblityState`, etc. See comments in `a11yHint.ts` for details.

src/queries/a11yStates.ts
```diff
@@ -0,0 +1,79 @@
...
+  )}`;
+
+const { getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy } = makeQueries(
+  queryAllByA11yStates,
+  getMissingError,
+  getMultipleError
+);
+
+export type ByA11yStatesQueries = {
+  getByA11yStates: GetByQuery<AccessibilityStateKeyQueryParam>;
```
mdjastrzebski, line 62: Add alias names `getByAccessiblityStates`, etc. See comments in `a11yHint.ts` for details.

src/queries/a11yStates.ts
```diff
@@ -0,0 +1,79 @@
...
+  FindAllByQuery,
+  FindByQuery,
+  GetAllByQuery,
+  GetByQuery,
+  QueryAllByQuery,
+  QueryByQuery,
+} from './makeQueries';
+
+type AccessibilityStateKey = keyof AccessibilityState;
+type AccessibilityStateKeyQueryParam =
```
mdjastrzebski, line 14:
```suggestion
type AccessibilityStateKeys =
```

src/queries/a11yStates.ts
```diff
@@ -0,0 +1,79 @@
...
+    return prop.includes(matcher);
+  }
+
+  return !matcher.some((e) => !prop.includes(e));
+}
+
+const queryAllByA11yStates = (
+  instance: ReactTestInstance
+): ((
+  accessibilityStateKey: AccessibilityStateKeyQueryParam
```
mdjastrzebski, line 36:
```suggestion
  accessibilityStates: AccessibilityStateKeys
```

src/queries/a11yValue.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export function matchObject<T extends Record<string, unknown>>(
+  prop: T | undefined,
+  matcher: T
+): boolean {
+  return prop
+    ? Object.keys(matcher).length !== 0 &&
+        Object.keys(prop).length !== 0 &&
+        !Object.keys(matcher).some((key) => prop[key] !== matcher[key])
```
mdjastrzebski, line 26: Using de Morgans law it seems easier to read. Is there any reason for favoring negated `some` over `every`?

```suggestion
        Object.keys(matcher).every((key) => prop[key] === matcher[key])
```

mdjastrzebski: BTW our docs could be updated with info that we require all params to matched in order for element to be matched.

src/queries/labelText.ts
```diff
@@ -0,0 +1,68 @@
...
+): boolean {
+  if (!prop) {
+    return false;
+  }
+
+  if (typeof matcher === 'string') {
+    return prop === matcher;
+  }
+
+  return Boolean(prop.match(matcher));
```
mdjastrzebski, line 25: We probably could use `Regexp.test()` here.

src/queries/labelText.ts
```diff
@@ -0,0 +1,68 @@
...
+}
+
+const queryAllByLabelText = (
+  instance: ReactTestInstance
+): ((displayValue: TextMatch) => Array<ReactTestInstance>) =>
```
mdjastrzebski, line 30:
```suggestion
): ((text: TextMatch) => Array<ReactTestInstance>) =>
```

src/queries/a11yHint.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (typeof matcher === 'string') {
+    return prop === matcher;
+  }
+
+  return Boolean(prop.match(matcher));
+}
+
+const queryAllByA11yHint = (
+  instance: ReactTestInstance
+): ((displayValue: TextMatch) => Array<ReactTestInstance>) =>
```
mdjastrzebski, line 30:
```suggestion
): ((hint: TextMatch) => Array<ReactTestInstance>) =>
```

src/queries/a11yHint.ts
```diff
@@ -0,0 +1,68 @@
...
+  function queryAllByDisplayValueFn(displayValue) {
+    return instance.findAll(
+      (node) =>
+        typeof node.type === 'string' &&
+        matchStringValue(node.props.accessibilityHint, displayValue)
+    );
+  };
+
+const getMultipleError = (a11yHint: TextMatch) =>
```
mdjastrzebski, line 39:
```suggestion
const getMultipleError = (hint: TextMatch) =>
```

src/queries/a11yHint.ts
```diff
@@ -0,0 +1,68 @@
...
+  `Found multiple elements with accessibilityHint: ${String(a11yHint)} `;
+const getMissingError = (a11yHint: TextMatch) =>
```
mdjastrzebski, line 41:
```suggestion
const getMissingError = (hint: TextMatch) =>
```

src/queries/a11yState.ts
```diff
@@ -0,0 +1,68 @@
...
+  return prop
+    ? Object.keys(matcher).length !== 0 &&
+        Object.keys(prop).length !== 0 &&
+        !Object.keys(matcher).some((key) => prop[key] !== matcher[key])
+    : false;
+}
+
+const queryAllByA11yState = (
+  instance: ReactTestInstance
+): ((accessibilityStateKey: AccessibilityState) => Array<ReactTestInstance>) =>
```
mdjastrzebski, line 26:
```suggestion
): ((state: AccessibilityState) => Array<ReactTestInstance>) =>
```

src/queries/a11yState.ts
```diff
@@ -0,0 +1,68 @@
...
+  function queryAllByDisplayValueFn(accessibilityStateKey) {
+    return instance.findAll(
+      (node) =>
+        typeof node.type === 'string' &&
+        matchObject(node.props.accessibilityState, accessibilityStateKey)
+    );
+  };
+
+const getMultipleError = (a11yState: AccessibilityState) =>
```
mdjastrzebski, line 35:
```suggestion
const getMultipleError = (state: AccessibilityState) =>
```

src/queries/a11yState.ts
```diff
@@ -0,0 +1,68 @@
...
+  `Found multiple elements with accessibilityState: ${JSON.stringify(
+    a11yState
+  )}`;
+const getMissingError = (a11yState: AccessibilityState) =>
```
mdjastrzebski, line 39:
```suggestion
const getMissingError = (state: AccessibilityState) =>
```

src/queries/role.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (typeof matcher === 'string') {
+    return prop === matcher;
+  }
+
+  return Boolean(prop.match(matcher));
+}
+
+const queryAllByRole = (
+  instance: ReactTestInstance
+): ((displayValue: TextMatch) => Array<ReactTestInstance>) =>
```
mdjastrzebski, line 30:
```suggestion
): ((role: TextMatch) => Array<ReactTestInstance>) =>
```

src/queries/a11yValue.ts
```diff
@@ -0,0 +1,74 @@
...
+): ((displayValue: A11yValue) => Array<ReactTestInstance>) =>
+  function queryAllByDisplayValueFn(displayValue) {
+    return instance.findAll(
+      (node) =>
+        typeof node.type === 'string' &&
+        matchObject(node.props.accessibilityValue, displayValue)
+    );
+  };
+
+const getMultipleError = (a11yValue: A11yValue) =>
```
mdjastrzebski, line 41:
```suggestion
const getMultipleError = (value: A11yValue) =>
```

src/queries/a11yValue.ts
```diff
@@ -0,0 +1,74 @@
...
+  `Found multiple elements with accessibilityValue: ${JSON.stringify(
+    a11yValue
+  )} `;
+const getMissingError = (a11yValue: A11yValue) =>
```
mdjastrzebski, line 45:
```suggestion
const getMissingError = (value: A11yValue) =>
```

src/queries/a11yState.ts
```diff
@@ -0,0 +1,68 @@
...
+} from './makeQueries';
+
+export function matchObject<T extends Record<string, unknown>>(
+  prop: T | undefined,
+  matcher: T
+): boolean {
+  return prop
+    ? Object.keys(matcher).length !== 0 &&
+        Object.keys(prop).length !== 0 &&
+        !Object.keys(matcher).some((key) => prop[key] !== matcher[key])
```
mdjastrzebski, line 20: Using de Morgans law it seems easier to read. Is there any reason for favoring negated some over every?

```suggestion
        Object.keys(matcher).every((key) => prop[key] === matcher[key])
```

src/within.ts
```diff
@@ -1,11 +1,16 @@
 import type { ReactTestInstance } from 'react-test-renderer';
-import { a11yAPI } from './queries/a11yAPI';
 import { bindByTextQueries } from './queries/text';
 import { bindByTestIdQueries } from './queries/testId';
 import { bindByDisplayValueQueries } from './queries/displayValue';
 import { bindByPlaceholderTextQueries } from './queries/placeholderText';
 import { bindUnsafeByTypeQueries } from './queries/unsafeType';
 import { bindUnsafeByPropsQueries } from './queries/unsafeProps';
+import { bindByLabelTextQueries } from './queries/labelText';
```
mdjastrzebski, line 8: nit: Can we move these imports above `unsafe` and have them sorted: label, role, a11yhint, a11y state, a11y states, a11y value?

src/within.ts
```diff
@@ -15,7 +20,12 @@ export function within(instance: ReactTestInstance) {
     ...bindByPlaceholderTextQueries(instance),
     ...bindUnsafeByTypeQueries(instance),
     ...bindUnsafeByPropsQueries(instance),
-    ...a11yAPI(instance),
+    ...bindByLabelTextQueries(instance),
```
mdjastrzebski, line 23: nit: Can we move these bind calls above `unsafe` and have them sorted: label, role, a11yhint, a11y state, a11y states, a11y value?

src/queries/__tests__/a11yHint.test.tsx
```diff
@@ -38,55 +38,95 @@ const Section = () => (
...
-  expect(() => getAllByA11yHint(NO_MATCHES_TEXT)).toThrow(
-    getNoInstancesFoundMessage('accessibilityHint')
-  );
-  expect(queryAllByA11yHint(NO_MATCHES_TEXT)).toEqual([]);
-
-  await expect(findAllByA11yHint(TEXT_HINT)).resolves.toHaveLength(2);
-  await expect(findAllByA11yHint(NO_MATCHES_TEXT)).rejects.toThrow(
-    getNoInstancesFoundMessage('accessibilityHint')
-  );
+describe('getAllByA11yHint, queryAllByA11yHint, findAllByA11yHint', () => {
```
mdjastrzebski, line 87: Pls skip the describe block, and just use top-level `test`, e.g. `byA11yHint finds the element by a11y hint`

MattAgn:
> Great job @MattAgn! This is exactly what I was looking for! 💯
> 
> There are some small missing parts like aliased exports, please check comments.
> 
> Re moving `matchString`, etc to helpers: I think it's a good idea. I'd suggest `helpers/machers.ts` or similar for that purpose.
> 
> Re re-organizing tests a bit: I would suggest doing it in a separate PR, in order to reduce change surface area. This PR should just split existing tests into separate files without modifications. You're welcome re-organise them a little bit more in the follow-up PR :-)

Thanks! 

Re moving `matchString`, etc to helpers: got it, will do so

Re re-organizing tests a bit: sure I'll do that :)

src/queries/__tests__/a11yHint.test.tsx
MattAgn: yes, I removed the commit that reorganized the queries by test case (with the describes) will do that in the follow up POUR

src/queries/a11yStates.ts
```diff
@@ -0,0 +1,79 @@
...
+): boolean {
+  if (!prop || matcher.length === 0) {
+    return false;
+  }
+
+  if (typeof matcher === 'string') {
+    return prop.includes(matcher);
+  }
+
+  return !matcher.some((e) => !prop.includes(e));
```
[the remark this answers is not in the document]
AugustinLF: I don't know what it is the desired behaviour, but I would expect to require all the states

mdjastrzebski:
@MattAgn let me know if you need any help with this PR.

MattAgn:
@mdjastrzebski sure thanks! for now I'm good, sorry I did not have much time to work on it the past 2 weeks, but I have some time this weekend. I'll let you know how it goes :)

src/queries/a11yHint.ts
```diff
@@ -0,0 +1,68 @@
...
+  `Unable to find an element with accessibilityHint: ${String(a11yHint)}`;
+
+const { getBy, getAllBy, queryBy, queryAllBy, findBy, findAllBy } = makeQueries(
+  queryAllByA11yHint,
+  getMissingError,
+  getMultipleError
+);
+
+export type ByA11yHintQueries = {
```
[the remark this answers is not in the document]
MattAgn: @mdjastrzebski  Oh indeed I missed that! But in the previous API, when I look at the [interface](https://github.com/callstack/react-native-testing-library/blob/597fa3c4f84b40cccfd28a2babeb90d98dbe7c93/typings/index.flow.js#L214) `A11yAPI`  , I only see `getByHintText` and `getByA11yHint` but not `getByAccessibilityHint`. Should I add it then? 

In the same way, I don't see `getByAccessiblityState` nor `getByState` in the former API nor any of the other aliases for states, role, label etc... (byAccessibilityXXX or getByXXX). Should I add them all? I agree that it would be better if every query followed the same pattern.
If I do change the API, should I do in this PR or in another one? (to keep this one a pure refactoring)

src/queries/a11yStates.ts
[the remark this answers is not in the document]
MattAgn: Indeed I don't know, though I did not change the logic, I copy pasted the `matchArrayValue` function from the previous a11yApi file. Maybe we should open an issue to discuss it? (as this PR should not change any behavior)

src/queries/a11yStates.ts
MattAgn: done

src/queries/a11yStates.ts
MattAgn: done

src/queries/a11yValue.ts
MattAgn: Agreed, your version is easier to read. I must admit that for the matchers, I purely copy pasted the existing code without trying to improve it

As for the docs, I can do it in a follow up PR, what do you think?

src/queries/labelText.ts
MattAgn: done

src/queries/a11yHint.ts
MattAgn: done

src/queries/a11yHint.ts
MattAgn: done

src/queries/a11yHint.ts
MattAgn: done

src/queries/a11yState.ts
MattAgn: done

src/queries/a11yState.ts
MattAgn: done

src/queries/a11yState.ts
MattAgn: I'm curious, why state over a11yState? the type says it's an accessibility state, it's more specific

src/queries/role.ts
MattAgn: done

src/queries/a11yValue.ts
MattAgn: done

src/queries/a11yValue.ts
MattAgn: done

src/queries/a11yStates.ts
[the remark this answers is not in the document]
MattAgn: Hmm I think it's ok in the end, because it checks that all the values in the matcher are included in the prop accessiblityStates, not the other way around
so : 
```js
matchArrayValue(["selected", "disabled"], [”selected"]) // returns true
matchArrayValue(["selected"], [”selected", "disabled"]) // returns false
```
You can check the tests I did [here](https://github.com/callstack/react-native-testing-library/pull/977/commits/24f5259ab85d7fb1514a9127a4725faac4a17978)

I tried to make the function more explicit in this [commit](https://github.com/callstack/react-native-testing-library/pull/977/commits/5c92cb5e9592fb7b2c3ee4e726917011871194c9)

src/queries/a11yState.ts
MattAgn: done

src/within.ts
MattAgn: done

src/within.ts
MattAgn: done

src/queries/a11yHint.ts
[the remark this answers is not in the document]
mdjastrzebski: Long story short about a11y prop names:
1. originally we had all a11y queries named `getByA11yXxx` & `getByAccessibilityXxx`. 
2. At some stage we decided to use RTL naming for these queries which had direct counterparts in RTL (iirc: `byRole`) and keep the existing `byA11y` name for all other which are not present in RTL. 
3. However we also kept double naming for `byHintText`/`byA11yHint` because it was present in alternative RNTL library which we merged with.

Generally I would not remove or rename any queries, because that would cause a breaking change. We might might fill in some missing gaps as that is not breaking.

mdjastrzebski (approved):
Looks good

mdjastrzebski:
Weird stuff tests work fine on my local machine but timeout on CI.

thymikee:
@mdjastrzebski Try increasing the timeout then

mdjastrzebski:
@thymikee actually is not ma matter of timeout, I've tried to increase that but this is not the cause. Normally tests run around 1,5 minutes (other PR runs checked in Circle CI), but for this particular PR they run most of the tests (but not all) and then Jest gets stuck, no output, etc. It will wait till Circle CI no output limit and then timeout after 10 minutes or so.

thymikee (approved):
Awesome work @MattAgn and @mdjastrzebski 🚀

MattAgn:
@mdjastrzebski  thanks for the finishing touches! :)

PR #979, opened 2022-05-19
docs: add FAQ page and native features question

mdjastrzebski:
### Summary

Adds FAQ page to the docs and provides answer to question about testing native features in RNTL.

### Test plan

None

### DIFF
--- a/website/docs/FAQ.md
+++ b/website/docs/FAQ.md
@@ -10,20 +10,24 @@ title: FAQ
 <p>Short answer: no.</p>
 
 React Native Testing Library does not provide a full React Native runtime since that would require running on physical device
-or iOS simulator/Android emulator to provide the underlying OS and platform APIs.
+or iOS simulator/Android emulator to provision the underlying OS and platform APIs.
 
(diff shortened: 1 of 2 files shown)


website/docs/FAQ.md
```diff
@@ -0,0 +1,29 @@
...
+
+- it allows testing most of the logic of regular React Native apps
+- it allows running test on any OS supported by Jest or other test runner, e.g. on CI
+- it uses much less resources than full runtime emulation
+- you can use Jest fake timers
+
+The the negative side:
+
+- you cannot test native features
+- certain JavaScript features might not be perfectly simulated by we are working on it
```
AugustinLF, line 27: I'd suggest adding a couple of examples (some people not well versed on the difference between native/js might struggle to figure out this barrier).

```
For instance, [react-native's ScrollView](https://reactnative.dev/docs/scrollview) has several props that depend on native calls. While you can trigger `onScroll` call with `fireEvent.scroll`, `onMomentumScrollBegin` is called from the native side and will therefore not be called.
```

AugustinLF (changes requested):

website/docs/FAQ.md
mdjastrzebski: Added the example you mentioned.

thymikee (approved):
🚀

PR #980, opened 2022-05-19
docs: table of contents for Queries and API pages

mdjastrzebski:
### Summary
Adds table of contents for Queries and APIs pages since they are quite long and hence hard to navigate.

### Test plan
None

### DIFF
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -3,6 +3,34 @@ id: api
 title: API
 ---
 
+### Table of contents:
+
+- [`render`](#render)
+  - [`update`](#update)
+  - [`unmount`](#unmount)
+  - [`debug`](#debug)
+  - [`toJSON`](#tojson)
+  - [`container`](#container)
+- [`cleanup`](#cleanup)
+- [`fireEvent`](#fireevent)
+- [`fireEvent[eventName]`](#fireeventeventname)
+  - [`fireEvent.press`](#fireeventpress)
+  - [`fireEvent.changeText`](#fireeventchangetext)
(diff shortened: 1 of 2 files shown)


website/docs/API.md
```diff
@@ -519,40 +559,34 @@ A `RenderHookOptions<Props>` object to modify the execution of the `callback` fu
...
-```ts
-function rerender(newProps?: Props): void;
-```
-
-A function to rerender the test component, causing any hooks to be recalculated. If `newProps` are passed, they will replace the `callback` function's `initialProps` for subsequent rerenders.  The `Props` type is determined by the type passed to or inferred by the `renderHook` call.
+A function to rerender the test component, causing any hooks to be recalculated. If `newProps` are passed, they will replace the `callback` function's `initialProps` for subsequent rerenders. The `Props` type is determined by the type passed to or inferred by the `renderHook` call.
 
 #### `unmount`
 
-```ts
```
AugustinLF, line 552: Why remove the API here while keeping it for `fireEvent.scroll: (element: ReactTestInstance, ...data: Array<any>) => void` for instance?

mdjastrzebski: This `rerender` refers to `RenderHookResult` type which is kinda its own section. So I've put the whole type definition just above sections for first of it (`result`). Do you think we should repeat it here?

AugustinLF: No you're right, it makes sense!

AugustinLF (approved):

thymikee (approved):
👍🏼

PR #984, opened 2022-05-24
fix: queryByText and getByText match newlines and tabs properly

### LINKED ISSUE #983: queryByText with newline \n doesn't find matching rendered text
## Describe the bug

New lines, that are rendered in RN as separate Lines, are wrongly normalised during testing.

```tsx
const { queryByText } = render(<Text>Line 1\nLine 2</Text>)
expect(queryByText("Line 1\nLine 2")).toBeDefined()
```

## Expected behavior

I expect the test to pass.
I have checked the queryByText and it seems it uses the same normaliser (Default) as the findTextByNode so this should actually work.

I know that in HTML whitespace is ignored and normalised by browsers to one space. But in RN it actually matters and is rendered (at least \n whitespace).

The test can be made to pass by this change:

```diff
-expect(queryByText("Line 1\nLine 2")).toBeDefined()
+expect(queryByText("Line 1\nLine 2".split("\n").join(" ")).toBeDefined()
```

which indicates to me, the queryByText does in fact not normalises the input query text correctly.

## Versions

 npmPackages:
    @testing-library/react-native: 9.0.0 => 9.0.0 
    react: 17.0.2 => 17.0.2 
    react-native: 0.66.3 => 0.66.3 
    react-test-renderer: 16.14.0 => 16.14.0

pke:
### Summary

This fixes #983.

New lines, that are rendered in RN as separate Lines, are wrongly normalised during testing.

```tsx
const { queryByText } = render(<Text>{"Line 1\nLine 2"}</Text>)
expect(queryByText("Line 1\nLine 2")).toBeDefined()
``` 

### Test plan

I have added 2 new tests to verify the fix is working.

### DIFF
--- a/src/queries/__tests__/text.test.tsx
+++ b/src/queries/__tests__/text.test.tsx
@@ -473,3 +473,17 @@ test('getByText and queryByText work properly with multiple nested fragments', (
   expect(getByText('Hello')).toBeTruthy();
   expect(queryByText('Hello')).not.toBeNull();
 });
+
+test('getByText and queryByText work with newlines', () => {
+  const textWithNewLines = 'Line 1\nLine 2';
+  const { getByText, queryByText } = render(<Text>{textWithNewLines}</Text>);
+  expect(getByText(textWithNewLines)).toBeTruthy();
+  expect(queryByText(textWithNewLines)).toBeTruthy();
+});
+
+test('getByText and queryByText work with tabs', () => {
+  const textWithTabs = 'Line 1\tLine 2';
+  const { getByText, queryByText } = render(<Text>{textWithTabs}</Text>);
+  expect(getByText(textWithTabs)).toBeTruthy();
+  expect(queryByText(textWithTabs)).toBeTruthy();
+});
--- a/src/matches.ts
+++ b/src/matches.ts
@@ -13,9 +13,10 @@ export function matches(
 
   const normalizedText = normalizer(text);
   if (typeof matcher === 'string') {
+    const normalizedMatcher = normalizer(matcher);
     return exact
-      ? normalizedText === matcher
-      : normalizedText.toLowerCase().includes(matcher.toLowerCase());
+      ? normalizedText === normalizedMatcher
+      : normalizedText.toLowerCase().includes(normalizedMatcher.toLowerCase());
   } else {
     return matcher.test(normalizedText);
   }


naabraz (approved):

AugustinLF (commented):
While I think this change makes sense, I believe it would get us away from `@testing-library/react`, which doesn't do such normalisation afaik. Do we think it's an ok tradeoff?

pke:
@AugustinLF That is because React Web an RN handle whitespace in JSX differently. In RW it uses HTML whitespace policy, that is, to ignore whitespace. However, RN respects whitespace inside `<Text>` tags.

src/queries/__tests__/text.test.tsx
```diff
@@ -473,3 +473,15 @@ test('getByText and queryByText work properly with multiple nested fragments', (
   expect(getByText('Hello')).toBeTruthy();
   expect(queryByText('Hello')).not.toBeNull();
 });
+
+test('getByText and queryByText work with newlines', () => {
+  const { getByText, queryByText } = render(<Text>{'Line 1\nLine 2'}</Text>);
```
mdjastrzebski, line 478:
```suggestion
  const textWithNewlines = 'Line 1\nLine 2';
  const { getByText, queryByText } = render(<Text>{textWithNewlines}</Text>);
```

src/queries/__tests__/text.test.tsx
```diff
@@ -473,3 +473,15 @@ test('getByText and queryByText work properly with multiple nested fragments', (
...
+  expect(getByText('Line 1\nLine 2')).toBeTruthy();
+  expect(queryByText('Line 1\nLine 2')).not.toBeNull();
```
mdjastrzebski, line 480:
```suggestion
  expect(queryByText('Line 1\nLine 2')).toBeTruthy();
```

mdjastrzebski (approved):
@AugustinLF Looking at the code it seems that we should match `Line1\nLine2' when `Text` component contains the same very same text content. Before this PR we normalised text but did not normalize the the text matcher string which caused the `normalizedText === matcher` check to fail.

mdjastrzebski:
As a general comment, regex text queries are more powerful and don't require perfect matching when using e.g. `\s`

src/queries/__tests__/text.test.tsx
pke: I was following the existing test code in the file:
https://github.com/callstack/react-native-testing-library/pull/984/files#diff-c98a36aa080f209baf1bdb04f2012f180f654c992c0d5c79dc69e21775ec15abR474

I have changed my code only now.

src/queries/__tests__/text.test.tsx
pke: I have replaced all inline test texts with references now. Thanks.

pke:
Rebased and applied suggestions from comments.

> As a general comment, regex text queries are more powerful and don't require perfect matching when using e.g. `\s`

I know but here its important, that the formatting is matched exactly, as it has direct visual consequences in RN Text elements unlike React Web.

mdjastrzebski:
@thymikee @AugustinLF We're looking for a second approval for this PR. Can you help?

thymikee (approved):

PR #994, opened 2022-06-23
feat: add `screen` API

mdjastrzebski:
### Summary

Simple `screen` implementation to match RTL feature. 

https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

### Test plan
* basic `screen` related tests added

### Steps
- [x] Implementation
- [x] Tests
- [x] Docs
- [x] Flow types

### DIFF
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -9,10 +9,12 @@ import { getDefaultNormalizer } from './matches';
 import { renderHook } from './renderHook';
 import { screen } from './screen';
 
+export type { RenderResult };
+
 export { act };
 export { cleanup };
 export { fireEvent };
-export { render, RenderResult };
+export { render };
 export { waitFor };
 export { waitForElementToBeRemoved };
 export { within, getQueriesForElement };
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -10,6 +10,7 @@ import { renderHook } from './renderHook';
 import { screen } from './screen';
 
 export type { RenderResult };
+export type RenderAPI = RenderResult;
 
 export { act };
 export { cleanup };
@@ -21,4 +22,3 @@ export { within, getQueriesForElement };
 export { getDefaultNormalizer };
 export { renderHook };
 export { screen };
-export type RenderAPI = ReturnType<typeof render>;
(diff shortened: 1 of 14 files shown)


src/pure.ts
```diff
@@ -1,19 +1,21 @@
...
 import { within, getQueriesForElement } from './within';
 import { getDefaultNormalizer } from './matches';
 import { renderHook } from './renderHook';
+import { screen } from './screen';
 
 export { act };
 export { cleanup };
 export { fireEvent };
-export { render };
+export { render, RenderResult };
```
thymikee, line 15: can we export this explicitly as a type?

website/docs/API.md
```diff
@@ -6,30 +6,41 @@ title: API
...
 - [`fireEvent[eventName]`](#fireeventeventname)
   - [`fireEvent.press`](#fireeventpress)
   - [`fireEvent.changeText`](#fireeventchangetext)
   - [`fireEvent.scroll`](#fireeventscroll)
+    - [On a `ScrollView`](#on-a-scrollview)
+    - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
 - [`waitForElementToBeRemoved`](#waitforelementtoberemoved)
-- [`within, getQueriesForElement`](#within-getqueriesforelement)
+- [`within`, `getQueriesForElement`](#within-getqueriesforelement)
```
thymikee, line 28: good catch :D

website/docs/API.md
```diff
@@ -154,13 +171,23 @@ container: ReactTestInstance;
...
 
+## `screen`
+
+```ts
+let screen: RenderResult;
+```
+
+Hold the value of latest render call for easier access to query and other functions returned by [`render`](#render). 
+
+It's value is automatically cleared after each test by calling [`cleanup`](#cleanup). If no `render` call has been made in a given test then it holds a special object that implements `RenderResult` but throws a helpful error on each property and method access.
```
thymikee, line 182:
```suggestion
Its value is automatically cleared after each test by calling [`cleanup`](#cleanup). If no `render` call has been made in a given test then it holds a special object that implements `RenderResult` but throws a helpful error on each property and method access.
```

thymikee (approved):
Whoa, this is good. Makes us way closer to a typical RTL usage now 🎉

PR #1000, opened 2022-06-24
fix: bring back RenderAPI type

### LINKED ISSUE #999: RenderAPI type not exported anymore?
## Ask your Question

I upgraded to @testing-library/react-native v10 and noticed that there were some breaking changes.

Like for example we were using the [RenderAPI](https://github.com/artsy/eigen/blob/47011c3be8b49d0fa6a4dd948a88f37dc41e57ac/src/app/Components/Bidding/Screens/Registration.tests.tsx#L57) type in our project and after the upgrade it is not exported, with what should we replace that?

thymikee:
### Summary

Fixes #999 

### Test plan

Added test

### DIFF
--- a/src/__tests__/render.test.tsx
+++ b/src/__tests__/render.test.tsx
@@ -7,7 +7,7 @@ import {
   SafeAreaView,
 } from 'react-native';
 import stripAnsi from 'strip-ansi';
-import { render, fireEvent } from '..';
+import { render, fireEvent, type RenderAPI } from '..';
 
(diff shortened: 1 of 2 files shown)


thymikee:
@mdjastrzebski we can introduce it along with `screen` PR :) there's nothing wrong with exporting it as an alias and deprecate in the future

gkartalis:
👏 👏 👏  Thanks for addressing that so quickly!

mdjastrzebski (approved):

PR #1015, opened 2022-07-20
refactor: remove `byA11yStates` queries

mdjastrzebski:
### Summary

Remove `byA11yStates` queries as they are not supported by RN from couple of releases. The supported alternative is `byA11yState`.

### Test plan

All checks and tests pass.


thymikee (approved):
🚀

PR #1025, opened 2022-07-27
feature: basic example app

mdjastrzebski:
### Summary

Basic modern example for using RNTL. The app is generated from RN CLI template so there is some iOS/Android boilerplate but the benefit here is that we work on a real RN app and not an empty shell of it.

I also plan to use it as a base for refresh of other example apps.

### Test plan

All checks and tests pass.


examples/basic/App.tsx
```diff
@@ -0,0 +1,24 @@
...
+      </SafeAreaView>
+    );
+  }
+
+  return (
+    <SafeAreaView>
+      <Home user={user} />
+    </SafeAreaView>
+  );
+};
```
thymikee, line 22: less code is less code
```suggestion
  return (
    <SafeAreaView>
      {user == null ? <LoginForm onLoginSuccess={setUser} /> : <Home user={user} />}
    </SafeAreaView>
  );
};
```

examples/basic/__tests__/App.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+  fireEvent.changeText(screen.getByLabelText('Password'), 'admin1');
+
+  // Hint: we can use `getByText` to find our button by its text.
+  fireEvent.press(screen.getByText('Sign In'));
+
+  // Idiom: since pressing button triggers async operation we need to use `findBy` query to wait
+  // for the action to complete.
+  // Hint: subsequent queries do not need to use `findBy`, because they are used after the async action
+  // already finished
+  expect(await screen.findByText('Welcome admin!')).toBeTruthy();
```
thymikee, line 43: There's no need to expect everything, right? This reads better as a scenario to me:
```suggestion
  await screen.findByText('Welcome admin!'));
```

thymikee (commented):
Looks good! For the ease of maintenance I'd strongly consider not committing the "android" and "ios" folders (we could generate those with `init` on demand) or use Expo

examples/basic/__tests__/App.test.tsx
mdjastrzebski: Yeah thats a good question, I'used that in a similar pattern as we suggest using `expect(getByText()).toBeTruthy()` to clarify the intent, despite the fact that simple `getByText` would do the same.

mdjastrzebski:
I've migrated the example to use Expo managed.

examples/basic/.prettierrc.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  arrowParens: 'avoid',
+  bracketSameLine: true,
+  bracketSpacing: true,
+  singleQuote: true,
+  trailingComma: 'all',
+  printWidth: 100,
+};
```
thymikee, line 8: let's remove this and apply the same prettier settings as the rest of the repo :)

mdjastrzebski: done

thymikee (approved):

PR #1034, opened 2022-08-03
Export options types and `renderHook` result type

evanwalsh:
### Summary

While writing helpers around the `render` and `renderHook` functions, I found myself wanting/needing the types for the options and the returns. This explicitly exports those types that weren't already exposed to avoid having to do something like the following:

```typescript
type RenderOptions = Parameters<typeof render>[1]
type RenderHookOptions = Parameters<typeof renderHook>[1]

// Copied/pasted from @testing-library/react-native
interface RenderHookResult<Result, Props> {
  result: {current: Result}
  rerender: (props: Props) => void
  unmount: () => void
}
```

These types are already documented, so this just fully exposes types that are already meant for users.

### Test plan

I'm not sure what additional tests would be needed for this. As long as type checking succeeds, these changes should be good.

### DIFF
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -1,15 +1,15 @@
 import act from './act';
 import cleanup from './cleanup';
 import fireEvent from './fireEvent';
-import render, { RenderResult } from './render';
+import render, { RenderResult, RenderOptions } from './render';
 import waitFor from './waitFor';
 import waitForElementToBeRemoved from './waitForElementToBeRemoved';
 import { within, getQueriesForElement } from './within';
 import { getDefaultNormalizer } from './matches';
-import { renderHook } from './renderHook';
+import { renderHook, RenderHookOptions, RenderHookResult } from './renderHook';
 import { screen } from './screen';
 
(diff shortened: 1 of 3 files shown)


mdjastrzebski (changes requested):
@evanwalsh thanks for submitting the PR. I guess it makes sens to export these types. The chosen names also seem correct.

You need to add `export` clauses to `pure.js` or otherwise users will not be able to directly import these types from `@testing-library/react-native`

src/pure.ts
```diff
@@ -20,5 +20,5 @@ export { waitFor };
 export { waitForElementToBeRemoved };
 export { within, getQueriesForElement };
 export { getDefaultNormalizer };
-export { renderHook };
+export { renderHook, RenderHookOptions, RenderHookResult };
```
mdjastrzebski, line 23: Export `RenderHookOptions`, `RenderHookResult` as `export type` next to other `export type` declarations.

mdjastrzebski (changes requested):
Almost there :-)

PR #1050, opened 2022-08-09
chore: remove release-it as it's unused

thymikee:
### Summary

I haven't used it in a good while, so why keeping it.

### Test plan

None


mdjastrzebski (approved):
Looks good! 👍🏻

PR #1059, opened 2022-08-10
chore: update deps to expo 46

mdjastrzebski:
### Summary

Update basic example app to use Expo 46. Should also fix running example apps due to "Jest not exited error"
 
### Test plan

All checks pass.


thymikee (approved):
Yes please!

PR #1064, opened 2022-08-16
feat: option for queries to respect accessibility

### LINKED ISSUE #970: Provide query option to respect accessibility
## Describe the Feature
This issue is based on #787 PR. That PR is stale but idea seems worth pursuing.

Currently all queries ignored accessibility ignored the fact that certain element, e.g. screen or modal is present in component tree but not really visible to the user.

This problem largely does not exist in web React world as previous screens are removed from component tree, except modals, but it is relevant in React Native world. We even suggest some work arounds for it with [`within` function](https://callstack.github.io/react-native-testing-library/docs/api#within-getqueriesforelement).

## Possible Implementations
* Add additional `respectAccessibility` query options for all queries. This query options would be common to all query verbs (`getBy`, etc) and predicates (`byText`, etc). 
* By default it should be disabled to preserve API stability
* If set explicitly to true it would ignore elements that are not accessible, by analysing e.g. `accessibilityElementsHidden`, `importantForAccessibility`, `accessibilityViewIsModal`, `style={{ display: 'none' }}`, etc
* Consider switching this option to true by default by introducing a breaking change in next major release, as having it on by default is probably what users would expect.
* We could salvage some a11y predicate code and possibly all tests from #787

## Related Issues
* #787 for initial idea

MattAgn:
# Summary

Resolves #970 

This PR aims to solve [this issue](https://github.com/callstack/react-native-testing-library/issues/970).
It adds a "respectAccessibility" option to all queries (false by default) that enables queries to ignore elements that would be hidden in an environment respecting accessibility props

# TODO

- [x] Code and test a function isReactElementVisibleToAccessibility that does all the checks
- [x] rename tests titles properly in accessibility.test.ts
- [x] rename the option `hidden` to comply to RTL naming
- [x] add the a11y check in all queries
- [x] add a test for each query
- [x] add documentation
- [x] update flow types

# Test plan

- ~~The function `isReactElementVisibleToAccessibility` is 100% tested~~ Using `isInaccessible()` which is already tested
- The queries all have one test to check the use of the option respectAccessibility

# Next steps

- Make another implementation of this feature by implementing a new `findAll` that would do the a11y check top down instead of bottom up as we do today
- ~~Make performance tests to choose the best implementation~~ (Matching DTL implementation (bottom up + cache))


mdjastrzebski:
I wonder about naming our option. RTL's `getByRole` has `hidden` option that kind matches what we are doing here: https://testing-library.com/docs/queries/byrole#hidden 

@MattAgn @thymikee @AugustinLF wdyt about proper name for this new option?

MattAgn:
Wow so quick @mdjastrzebski 😮

src/helpers/isReactElementVisibleToAccessibility.ts
```diff
@@ -0,0 +1,40 @@
+import { ReactTestInstance } from 'react-test-renderer';
+
+export function isReactElementVisibleToAccessibility(
```
mdjastrzebski, line 3: Seems like this might be expensive to compute, as `findAll` does top-down tree travelsal while this function is doing bottom-up travelsal to root on each parent. Maybe we should have a `findAll` that would exclude component tree branches at it performs top-down travelsal. We could start with `findAll` code, duplicate it and add our tweak there.

MattAgn: oh I thought you said the opposite on the previous PR that was opened

MattAgn: CleanShot 2022-08-16 at 15 56 09@2x

MattAgn:
@mdjastrzebski thanks a lot for the quick review!
I'm far from done though, the work I still have to do: 
- add the a11y check in all queries
- add a test for each query

src/helpers/isReactElementVisibleToAccessibility.ts
MattAgn: Maybe we should try both approaches and compare performances?

AugustinLF:
I like to have an API as close as possible from rtl web, so using `hidden` sounds like a great solution. Since their API lets user should the default value of hidden, we could release that in a minor with a default value of `true`, and turn it on by default in a major release (which is IMO better, and is also the default of RTL).

mdjastrzebski:
re default value: that would be the plan to have a minor with default behaving as is, and then in the next major release swap the switch to more reasonable value of `hidden: false`

src/helpers/isReactElementVisibleToAccessibility.ts
mdjastrzebski: Haha, nice catch 💯 
Recently I've learnt that some of our users have pretty large test bases that take longer time to run, so I try to avoid added too much additional run time for them. I agree that measurement is the best way of deciding which way to go. We could start with the simplest (which seems to be traversing up to check parents) and then have another PR with faster algorithm if it's indeed faster in any way.

src/helpers/accessibility.ts
```diff
@@ -0,0 +1,43 @@
+import { ReactTestInstance } from 'react-test-renderer';
```
mdjastrzebski, line 1: In TL/Jest-Native repo there is a related [PR for `toBeVisible` Jest matcher](https://github.com/testing-library/jest-native/pull/73/files). It seems that they are related and check the same concept of "being visible to the user". From the other PR I've noticed that they also check for `opacity: 0` style

src/helpers/isReactElementVisibleToAccessibility.ts
MattAgn: Totally agree with your idea! Seems like the most pragmatic to me. And if people find it too slow at first for their codebase, they can still just not activate the option until we push a more optimised version

As for slow test, I've seen many project at my company that had big codebases and slow tests with RNTL and jesr. We even had to make a guide on how to improve their speed. Our most used quick win is using [sucrase](https://github.com/alangpierce/sucrase) for our tests now, which drastically improves the speed of our tests

src/helpers/accessibility.ts
MattAgn: oh great catch! I'll add it to the list then

MattAgn:
> I like to have an API as close as possible from rtl web, so using `hidden` sounds like a great solution. Since their API lets user should the default value of hidden, we could release that in a minor with a default value of `true`, and turn it on by default in a major release (which is IMO better, and is also the default of RTL).

Yes sounds good!
What do you think of the clarity of the naming `hidden` though? I don't find it very explicit about its purpose

mdjastrzebski:
@MattAgn I'm on a fence here, on one hand it mimics RTL convention but on the other hand it's not very self explanatory. However I wonder if the other name (`respectAccessiblity`) is explanatory as well.

Seems like this is a concept that is less needed in RTL due to how web navigation works vs RN.

I think it will make more sense if/when we switch this feature on by default in a next major release, which btw will probably generate a lot of user commencts due to failing tests ;-)

MattAgn:
@mdjastrzebski I agree that respectAccessibility was not so clear either, especially since we also check styles, maybe we should split it into 2 options `respectA11yVisbility` & `respectStylingVisibility`?
A possible way to do it would be to open an issue on RTL web to discuss the issue with its mainteneurs?

> I think it will make more sense if/when we switch this feature on by default in a next major release, which btw will probably generate a lot of user commencts due to failing tests ;-)
Totally, a migration guide will be required 😄
I did not understand what you mean by "it will make more sense when we switch", what will make more sense? The naming we use?

mdjastrzebski:
@MattAgn I think that 'hidden' option name will make more sense when we will ignore hidden elements by default.

MattAgn:
@mdjastrzebski oh alright I understand!

MattAgn:
@mdjastrzebski I found 2 different ways to handle the a11y option and I can't choose between the 2: 

1. Handle the option in each query independently, tried in [this commit](https://github.com/callstack/react-native-testing-library/pull/1064/commits/59a143f7d60621a2db484674a09df9beb1739520). Maybe I could shorten it a bit by extracting some logic a helper but we would still have a lot of duplication
	- Pros: handling the filtering of the results based on the options seems 
	- Cons: lots of code duplication, a bit harder to maintain

2. Handle the option in makeQueries, in [this commit](https://github.com/callstack/react-native-testing-library/pull/1064/commits/8ddad242600d8d2d3d065e002a1e5398993db515)	
    - Pros: option handled in one place, no code duplication, easier to maintain. It also make sure all the queries have the a11y option automatically (I guess that's something but not 100% sure)
    - Cons: add a new responsibility to makeQueries, originally, its only responsibility was to create the queries, now it's also doing some logic. Might not be super intuitive for someone looking into the codebase for the first time

Imo, the first option seems the best, because the responsibility argument is the most important to me
What do you think?

mdjastrzebski:
Since `hidden`/`respectAccessiblity` would be an option for query not for `render` then each query should check for that option and use either regualar `findAll` or our accessiblity-respecting `findAll`. Sind `makeQueries` relies on feeding it with `queryAll` variant then it probably not the place we want to make the switch between `findAll` variants. That lead us to picking `findAll` variant separately in each of the base query functions, e.g. https://github.com/callstack/react-native-testing-library/blob/f5db461f806ef222ac6a21a91839556e415607a3/src/queries/text.ts#L87

mdjastrzebski:
@MattAgn not sure but your first example looks a bit too complex. Seems like you could just declare some common `findAll` function:
```
function findAll(instance: ReactTestInstance, predicate: (instance: ReactTestInstance) => boolean, hidden: boolean) {
  // Use regular findAll fucntions
  if (hidden) return instance.findAll(predicate) 

  //else use new accessibility respecting algorithm,
}
```

Then inside each query, e.g. `queryAllByText` replace:
```ts
  function queryAllByTextFn(text, options) {
    const results = instance.findAll((node) =>
      getNodeByText(node, text, options)
    );

    return results;
  };
```

with
```ts
  // our new `findAll`
  function queryAllByTextFn(text, options) {
    const results = findAll(instance, (node) =>
      getNodeByText(node, text, options), options.hidden ?? true
    );

    return results;
  };
```

mdjastrzebski:
@MattAgn I've added `isInaccessible` function from DTL to RNTL. In DTL it is used in implementation of `hidden` option. https://github.com/testing-library/dom-testing-library/blob/a9a8cf26992ff0f6b4257b7300939f461d04440d/src/queries/role.js#L193

Would you have time to resume working on this PR? Having the `isInaccessible` helper this should make the remaining work somewhat easier.

MattAgn:
sorry @mdjastrzebski i was away for a while, i'm ready to start working back on this
thanks for the isInaccessible function! Indeed should make it easier :)

MattAgn:
> @MattAgn not sure but your first example looks a bit too complex. Seems like you could just declare some common `findAll` function:
> 
> ```
> function findAll(instance: ReactTestInstance, predicate: (instance: ReactTestInstance) => boolean, hidden: boolean) {
>   // Use regular findAll fucntions
>   if (hidden) return instance.findAll(predicate) 
> 
>   //else use new accessibility respecting algorithm,
> }
> ```
> 
> Then inside each query, e.g. `queryAllByText` replace:
> 
> ```ts
>   function queryAllByTextFn(text, options) {
>     const results = instance.findAll((node) =>
>       getNodeByText(node, text, options)
>     );
> 
>     return results;
>   };
> ```
> 
> with
> 
> ```ts
>   // our new `findAll`
>   function queryAllByTextFn(text, options) {
>     const results = findAll(instance, (node) =>
>       getNodeByText(node, text, options), options.hidden ?? true
>     );
> 
>     return results;
>   };
> ```

Yes sounds good to me

mdjastrzebski:
@MattAgn let's start with simplest implementation as in DTL: 
```ts
  root.findAll(...).filter(element => {
      return hidden === false
        ? isInaccessible(element, {
            isSubtreeInaccessible: cachedIsSubtreeInaccessible,
          }) === false
        : true
    })
```

https://github.com/testing-library/dom-testing-library/blob/a9a8cf26992ff0f6b4257b7300939f461d04440d/src/queries/role.js#L193

MattAgn:
> @MattAgn great to read that you're back!
> 
> re `findAll`: let's start with simplest implementation as in DTL:
> 
> ```ts
>   root.findAll(...).filter(element => {
>       return hidden === false
>         ? isInaccessible(element, {
>             isSubtreeInaccessible: cachedIsSubtreeInaccessible,
>           }) === false
>         : true
>     })
> ```
> 
> [testing-library/dom-testing-library@`a9a8cf2`/src/queries/role.js#L193](https://github.com/testing-library/dom-testing-library/blob/a9a8cf26992ff0f6b4257b7300939f461d04440d/src/queries/role.js#L193)
> 
> It contains some optimization in form of cachedIsSubtreeInaccessible.

you mean I don't extract the findAll in another function yet but instead we add some caching to isInaccessible?

mdjastrzebski:
@MattAgn We could replicate DTL behavior of using `cachedIsSubtreeInaccessible` for better coherence with DTL. While caching should also address performance issues.

MattAgn:
> ```ts
> cachedIsSubtreeInaccessible
> ```

sure, I saw how it works, not too hard to implement. I would still be in favor of extracting the findAll function though, wdyt?
Otherwise it's gonna be quite a pain to maintain considering the number of queries we have

The findAll could look something like (I simply copy pasted the code from DTL for now) : 
```js
export function findAll<Options extends AccessibilityOption>(
  instance: ReactTestInstance,
  searchCallback: (node: ReactTestInstance) => boolean,
  options?: Options
) {
  const subtreeIsInaccessibleCache = new WeakMap();
  function cachedIsSubtreeInaccessible(element) {
    if (!subtreeIsInaccessibleCache.has(element)) {
      subtreeIsInaccessibleCache.set(element, isSubtreeInaccessible(element));
    }

    return subtreeIsInaccessibleCache.get(element);
  }

  const results = instance.findAll(searchCallback);

  if (options?.hidden === false) {
    return results;
  }

  return results.filter(
    (result) =>
      !isInaccessible(result, { isSubtreeInaccessible: cachedIsSubtreeInaccessible })
  );
}
```

mdjastrzebski:
Looks pretty good, we could rename `searchCallback` to `predicate` as more descriptive name and then tweak the body a bit. Pls also add `defaultHidden` option to `configure` as in DTL

```ts
 export function findAll<Options extends AccessibilityOption>(
  instance: ReactTestInstance,
  predicate: (node: ReactTestInstance) => boolean,
  options?: Options
) {
  const matchingElement = instance.findAll(predicate);

  const hidden = options?.hidden ?? getConfig().defaultHidden; // We will want to add `defaultHidden: boolean` option to `configure`  
  if (hidden) {
    return matchingElement;
  }

  const subtreeIsInaccessibleCache = new WeakMap();

  function cachedIsSubtreeInaccessible(element) {
    if (!subtreeIsInaccessibleCache.has(element)) {
      subtreeIsInaccessibleCache.set(element, isSubtreeInaccessible(element));
    }

    return subtreeIsInaccessibleCache.get(element);
  }

  return matchingElement.filter(
    (element) =>
      !isInaccessible(element, { isSubtreeInaccessible: cachedIsSubtreeInaccessible })
  );
}
```

@MattAgn wdyt?

MattAgn:
@mdjastrzebski sounds great! I'm also done, I'm mainly missing the tests

src/helpers/accessiblity.ts
```diff
@@ -2,7 +2,18 @@ import { StyleSheet } from 'react-native';
 import { ReactTestInstance } from 'react-test-renderer';
 import { getHostSiblings } from './component-tree';
 
-export function isInaccessible(element: ReactTestInstance | null): boolean {
+type IsAnaccessibleOptions = {
```
mdjastrzebski, line 5:
```suggestion
type IsInaccessibleOptions = {
```

src/helpers/findAll.ts
```diff
@@ -0,0 +1,39 @@
...
+export function findAll<Options extends AccessibilityOption>(
+  instance: ReactTestInstance,
+  predicate: (node: ReactTestInstance) => boolean,
+  options?: Options
+) {
+  const results = instance.findAll(predicate);
+
+  //TODO: implement getConfig
+  const defaultHidden = true;
+  const hidden = options?.hidden ?? defaultHidden; // We will want to add `defaultHidden: boolean` option to `configure`
```
mdjastrzebski, line 14:
```suggestion
  const hidden = options?.hidden ?? getConfig().defaultHidden; // We will want to add `defaultHidden: boolean` option to `configure`
```

src/helpers/findAll.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    return subtreeIsInaccessibleCache.get(element);
+  }
+
+  return results.filter(
+    (result) =>
+      !isInaccessible(result, {
+        isSubtreeInaccessible: cachedIsSubtreeInaccessible,
+      })
+  );
```
mdjastrzebski, line 38: I wonder if it wouldn't be a better API if instead of `isInaccessible` didnt take `isSubtreeInaccessible` but rather it would take a cache option `cache?: WeakMap<ReactTestInstance>`, and it would just pass that cache to `isSubtreeInaccessible` and we would also put the cache hit logic there. Wdyt @MattAgn?

mdjastrzebski: That would deviate from DTL's option but that option feels kinda internal anyway

mdjastrzebski (commented):
@MattAgn seems like this PR is nearly ready for merging. Seems like we're only missing `defaultHidden` config option and docs, and we're ready to go 🚀

I've suggested a slightly different internal API for handling `isInaccessible` caching, but that's optional.

MattAgn:
@mdjastrzebski yes almost done! (Sorry I'm so slow on this PR, I've got a lot of time this week end, I'll finish it by then)
I'm also missing tests on our new `findAll`, do you agree it should be tested? (I've started it but have an issue with the initial `findAll`, I'll push it this week end and you can tell me what you think)

Should I implement the new namings as well? Everybody seems to agree on [this issue](https://github.com/callstack/react-native-testing-library/issues/1194)

src/helpers/findAll.ts
MattAgn: done

src/helpers/findAll.ts
MattAgn: totally agree!

MattAgn: will do that

src/helpers/accessiblity.ts
```diff
@@ -12,14 +16,24 @@ export const accessibilityStateKeys: AccessibilityStateKey[] = [
   'expanded',
 ];
 
-export function isInaccessible(element: ReactTestInstance | null): boolean {
+export function isInaccessible(
+  element: ReactTestInstance | null,
+  { cache }: IsInaccessibleOptions = { cache: undefined }
```
mdjastrzebski, line 21:
```suggestion
  { cache }: IsInaccessibleOptions = {}
```

src/queries/accessibilityOption.ts
```diff
@@ -0,0 +1 @@
+export type AccessibilityOption = { hidden?: boolean };
```
mdjastrzebski, line 1: 1. How about putting this is `options.ts` file, and moving `TextMatchOptions` also into it. 
2. I would also suggest using `BaseOptions` name or something similar, as the idea here is that all queries accept that option.

src/helpers/findAll.ts
```diff
@@ -0,0 +1,27 @@
+import { ReactTestInstance } from 'react-test-renderer';
+import { getConfig } from '../config';
+import { AccessibilityOption } from '../queries/accessibilityOption';
+import { isInaccessible } from './accessiblity';
+
+export function findAll<Options extends AccessibilityOption>(
```
mdjastrzebski, line 6: Do we need generics here, seem like it should work fine with concrete `AccessibilityOption` type.
```suggestion
export function findAll<Options extends AccessibilityOption>(
```

src/helpers/findAll.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+
+  const subtreeIsInaccessibleCache = new WeakMap<ReactTestInstance>();
+
+  return results.filter(
+    (result) =>
+      !isInaccessible(result, {
+        cache: subtreeIsInaccessibleCache,
+      })
+  );
```
mdjastrzebski, line 26: nit: seems like cache role is easily readable for call context in the next line

```suggestion
  const cache = new WeakMap<ReactTestInstance>();
  return results.filter((element) => !isInaccessible(element, { cache }));
```

src/queries/text.ts
```diff
@@ -86,8 +88,12 @@ const queryAllByText = (
...
 
-    const results = baseInstance.findAll((node) =>
-      getNodeByText(node, text, options)
+    const results = findAll(
+      baseInstance,
+      (node) => getNodeByText(node, text, options),
+      {
+        hidden: options?.hidden,
+      }
     );
```
mdjastrzebski, line 97: Can we just foward options?
```suggestion
    const results = findAll(
      baseInstance,
      (node) => getNodeByText(node, text, options),
      options,
    );
```

src/queries/displayValue.ts
```diff
@@ -34,8 +35,11 @@ const queryAllByDisplayValue = (
   queryOptions?: TextMatchOptions
 ) => Array<ReactTestInstance>) =>
   function queryAllByDisplayValueFn(displayValue, queryOptions) {
-    return instance.findAll((node) =>
-      getTextInputNodeByDisplayValue(node, displayValue, queryOptions)
+    return findAll(
+      instance,
+      (node) =>
+        getTextInputNodeByDisplayValue(node, displayValue, queryOptions),
+      { hidden: queryOptions?.hidden }
```
mdjastrzebski, line 42:
```suggestion
    return findAll(
      instance,
      (node) =>
        getTextInputNodeByDisplayValue(node, displayValue, queryOptions),
      queryOptions
```

src/queries/a11yValue.ts
```diff
@@ -19,12 +21,17 @@ type A11yValue = {
...
+  queryOptions?: AccessibilityOption
+) => Array<ReactTestInstance>) =>
+  function queryAllByA11yValueFn(value, queryOptions) {
+    return findAll(
+      instance,
       (node) =>
         typeof node.type === 'string' &&
-        matchObjectProp(node.props.accessibilityValue, value)
+        matchObjectProp(node.props.accessibilityValue, value),
+      { hidden: queryOptions?.hidden }
```
mdjastrzebski, line 34:
```suggestion
      queryOptions
```

src/queries/a11yState.ts
```diff
@@ -11,14 +12,20 @@ import type {
...
+  matcher: AccessibilityState,
+  queryOptions?: AccessibilityOption
+) => Array<ReactTestInstance>) =>
+  function queryAllByA11yStateFn(matcher, queryOptions) {
+    return findAll(
+      instance,
       (node) =>
-        typeof node.type === 'string' && matchAccessibilityState(node, matcher)
+        typeof node.type === 'string' && matchAccessibilityState(node, matcher),
+      { hidden: queryOptions?.hidden }
```
mdjastrzebski, line 28:
```suggestion
      queryOptions
```

mdjastrzebski (commented):
Awesome work @MattAgn! Pls add relevant docs and let's 🚢 it :-)

mdjastrzebski:
Pls rebase due to #1193 getting merged.

src/queries/a11yState.ts
MattAgn: i wondered about that actually, that's what i did at first and then thought that it did not necessarily make sense to pass all options to findAll when all he needs is hidden, wdyt ?

src/queries/accessibilityOption.ts
MattAgn: done

src/helpers/findAll.ts
MattAgn: agreed, dont remembrer why i did that

src/helpers/findAll.ts
MattAgn: agreed

src/queries/a11yState.ts
mdjastrzebski: From typing perspective, passing whole `options` object seems fine and natural, as it implement the relevant interface, hence we will should only access the props defined in that interface.

src/helpers/findAll.ts
```diff
@@ -0,0 +1,21 @@
+import { ReactTestInstance } from 'react-test-renderer';
+import { getConfig } from '../config';
+import { BaseOptions } from '../queries/options';
+import { isInaccessible } from './accessiblity';
+
```
mdjastrzebski, line 5: For clarity I think we should define 
```ts
interface FindAllOptions {
  hidden?: boolean;
}
```
which would be used in `findAll(..., options: FindAllOptions)`.

This will intentionally overlap with BaseQueryOptions, so that TS will be happy.

src/helpers/findAll.ts
```diff
@@ -0,0 +1,21 @@
...
+export function findAll(
+  instance: ReactTestInstance,
+  predicate: (node: ReactTestInstance) => boolean,
+  options?: BaseOptions
+) {
+  const results = instance.findAll(predicate);
+
+  const hidden = options?.hidden ?? getConfig().hidden; // We will want to add `defaultHidden: boolean` option to `configure`
```
mdjastrzebski, line 13:
```suggestion
  const hidden = options?.hidden ?? getConfig().hidden;
```

src/queries/options.ts
```diff
@@ -0,0 +1,8 @@
+import { NormalizerFn } from '../matches';
+
+export type BaseOptions = { hidden?: boolean };
```
mdjastrzebski, line 3: naming: Sorry for back and forth with the naming, I think we should add `query` part to the type name, as `BaseOptions` used through the code seems too generic, and also `Common` seems more appropriate, as `Base` suggest a base type for all options, but in many cases this is also the actual type, so `Common` seems more appropriate.

```suggestion
export type CommonQueryOptions = { hidden?: boolean };
```

src/queries/role.ts
```diff
@@ -14,10 +15,12 @@ import type {
   QueryAllByQuery,
   QueryByQuery,
 } from './makeQueries';
+import { BaseOptions } from './options';
 
 type ByRoleOptions = {
   name?: TextMatch;
-} & AccessibilityState;
+} & AccessibilityState &
+  BaseOptions;
```
mdjastrzebski, line 23: nit, style: Could we go with:
```ts
type ByRoleOptions = BaseOptions & AccessibilityState & {
  name?: TextMatch;
};
```

src/helpers/findAll.ts
```diff
@@ -0,0 +1,21 @@
+import { ReactTestInstance } from 'react-test-renderer';
+import { getConfig } from '../config';
+import { BaseOptions } from '../queries/options';
+import { isInaccessible } from './accessiblity';
+
+export function findAll(
+  instance: ReactTestInstance,
```
mdjastrzebski, line 7: readability, nit:
```suggestion
  root: ReactTestInstance,
```

mdjastrzebski (changes requested):
@MattAgn I've added final ("Final 2 new" 😉) nit comments, otherwise I am really happy with the implementation you've built! ro🚀

BTW seems like we still need:
- [ ] Docs updates
- [x] Flow type updates

src/queries/options.ts
MattAgn: indeed makes sense

src/helpers/findAll.ts
MattAgn: agreed, but then findAll should be used in all queries by default, so maybe we should have CommonQueryOptions extends FindAllOptions right?

MattAgn:
> @MattAgn I've added final ("Final 2 new" 😉) nit comments, otherwise I am really happy with the implementation you've built! ro🚀
> 
> BTW seems like we still need:
> 
> * [ ]  Docs updates
> * [ ]  Flow type updates

Indeed completly forgot about flow types!
@mdjastrzebski For the documentation, should I implement the new namings first? Then I can write the documentation with the new namings, otherwise we'll rewrite part of it

mdjastrzebski:
@MattAgn let's write docs with the existing `hidden` naming and merge it as is. We will provide this naming anyway for RTL-compat, and the improved naming is just an alias and we can make that changes in a separate smaller PR. I would like to get this feature merged, as constantly rebasing is simply a waste.

MattAgn:
Good point, i'm on it (and I agree, rebasing all the time is a pain 😅)

MattAgn:
Done for the documentation, I took some inspiration from RTL and the TextMatch option doc in RNTL, wdyt?

website/docs/Queries.md
```diff
@@ -355,6 +359,57 @@ render(<Component />);
 const element = screen.getByA11yValue({ min: 40 });
 ```
 
+## Hidden accessibility option
```
mdjastrzebski, line 362:
```suggestion
## `hidden` query option
```

website/docs/Queries.md
```diff
@@ -355,6 +359,57 @@ render(<Component />);
...
+
+All queries have the `hidden` option which enables them to respect accessibility props on components when it is set to `false`. If you set `hidden` to `true`, elements that are normally excluded from the accessibility tree are considered for the query as well.  For now if no `hidden` option is defined, `hidden` is set to `true` by default in order to avoid breaking changes. In the near future, we plan to set it to `false` by default so that the default behavior of all queries is to respect accessibility.
```
mdjastrzebski, line 364:
```suggestion
All queries have the `hidden` option which enables them to respect accessibility props on components when it is set to `false`. If you set `hidden` to `true`, elements that are normally excluded from the accessibility tree are considered for the query as well. Currently `hidden` option is set `true` by default, which means that elements hidden from accessibility will be included by default. However, we plan to change the default value to `hidden: false` in the next major release.
```

website/docs/Queries.md
```diff
@@ -355,6 +359,57 @@ render(<Component />);
...
+
+An element will be considered inaccessible when it checks one of the following: 
+- it has a style with `display: "none"`
+- it has the prop [`importantForAccessibility='no-hide-descendants'`](https://reactnative.dev/docs/accessibility#importantforaccessibility-android)
+- it has the prop [`accessibilityElementsHidden={true}`](https://reactnative.dev/docs/accessibility#accessibilityelementshidden-ios) 
+- one of its siblings has the prop [`accessibilityViewIsModal={true}`](https://reactnative.dev/docs/accessibility#accessibilityviewismodal-ios)
+- one of its parents is inaccessible due to one of the checks above
```
mdjastrzebski, line 371:
```suggestion
An element are considered to be hidden from accessibility based on [`isInaccessible()`](./API.md#isinaccessible) function result.
```

website/docs/Queries.md
```diff
@@ -355,6 +359,57 @@ render(<Component />);
...
+Will **find a match**:
+```ts
+screen.getByText("I am inaccessible", { hidden: true});
+screen.getByText("I am inaccessible");
+```
+
+Will **NOT find a match**:
+```ts
+screen.getByText("I am inaccessible", { hidden: false }); 
+```
```
mdjastrzebski, line 392: I would suggest compacting the examples as below:

```suggestion
Given the following render: 
```tsx
render(<Text style={{display:"none"}}>I am inaccessible</Text>);

// Ignore hidden elements
expect(screen.queryByText("I am inaccessible", { hidden: false })).toBeFalsy();

// Match hidden elements
expect(screen.getByText("I am inaccessible")).toBeTruthy(); // Defaults to hidden: true for now
expect(screen.getByText("I am inaccessible", { hidden: true})).toBeTruthy();
```

mdjastrzebski: I think single example (display: none) should be fine for now.

website/docs/Queries.md
```diff
@@ -355,6 +359,57 @@ render(<Component />);
...
+
+An element will be considered inaccessible when it checks one of the following: 
+- it has a style with `display: "none"`
+- it has the prop [`importantForAccessibility='no-hide-descendants'`](https://reactnative.dev/docs/accessibility#importantforaccessibility-android)
+- it has the prop [`accessibilityElementsHidden={true}`](https://reactnative.dev/docs/accessibility#accessibilityelementshidden-ios) 
+- one of its siblings has the prop [`accessibilityViewIsModal={true}`](https://reactnative.dev/docs/accessibility#accessibilityviewismodal-ios)
+- one of its parents is inaccessible due to one of the checks above
+
+You can find more documentation about accessibility in React Native [here](https://reactnative.dev/docs/accessibility).
+
```
mdjastrzebski, line 374: Pls also mention that you can use `configure` to set this option globally

mdjastrzebski (changes requested):
I've suggested some changes to simplify the doc entry and improve its structure a bit. 
Pls add also documentation for `configure` option.

src/helpers/findAll.ts
mdjastrzebski: Yeah, makes sens for CommonQueryOptions to extend FindAllOptions

website/docs/Queries.md
MattAgn: oh indeed i did not see we had docs for isInaccessible

website/docs/API.md
```diff
@@ -782,7 +783,8 @@ it('should use context value', () => {
 ```ts
 type Config = {
   asyncUtilTimeout: number;
-  defaultDebugOptions: Partial<DebugOptions>
+  defaultDebugOptions: Partial<DebugOptions>;
+  hidden: boolean;
```
mdjastrzebski, line 787: Isn't the option called default hidden?
```suggestion
  defaultHidden: boolean;
```

website/docs/API.md
```diff
@@ -795,12 +797,17 @@ Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementTo
 
 Default [debug options](#debug) to be used when calling `debug()`. These default options will be overridden by the ones you specify directly when calling `debug()`.
 
+#### `hidden` option
```
mdjastrzebski, line 800:
```suggestion
#### `defaultHidden` option
```

website/docs/API.md
```diff
@@ -795,12 +797,17 @@ Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementTo
...
+
+Default [hidden option](Queries.md#hidden-option) to be used for all queries. This default option will be overridden by the one you specify directly when using your query (eg: `getByText("Hello world", { hidden: true })`).
```
mdjastrzebski, line 802:
```suggestion
Default [hidden](Queries.md#hidden-option) query option for all queries. This default option will be overridden by the one you specify directly when using your query.
```

website/docs/Queries.md
```diff
@@ -26,10 +26,9 @@ title: Queries
     - [Default state for: `disabled`, `selected`, and `busy` keys](#default-state-for-disabled-selected-and-busy-keys)
     - [Default state for: `checked` and `expanded` keys](#default-state-for-checked-and-expanded-keys)
   - [`ByA11Value`, `ByAccessibilityValue`](#bya11value-byaccessibilityvalue)
-- [Hidden accessibility option](#hidden-accessibility-option)
-  - [Examples](#examples)
-    - [`display: "none"`](#display-none)
-    - [`accessibilityElementsHidden`](#accessibilityelementshidden)
+- [Common query options](#common-query-options)
+  - [`hidden` option](#hidden-option)
+    - [Examples](#examples)
```
mdjastrzebski, line 31: Can we use `**Examples**` instead of `#### Examples` to avoid generating TOC entry for example?

website/docs/Queries.md
```diff
@@ -359,57 +358,31 @@ render(<Component />);
...
-```
+#### Examples
 
-Will **NOT find a match**:
-```ts
-screen.getByText("I am inaccessible", { hidden: false }); 
-```
-
-#### `accessibilityElementsHidden`
+Given the following render:
```
mdjastrzebski, line 373: This looks like a left over from previous version of the doc
```suggestion
```

website/docs/Queries.md
```diff
@@ -359,57 +358,31 @@ render(<Component />);
...
-Will **NOT find a match**:
-```ts
-screen.getByText("I am inaccessible", { hidden: false }); 
-```
-
-#### `accessibilityElementsHidden`
+Given the following render: 
 
-Given the following render:
 ```ts
```
mdjastrzebski, line 375: Due to usage of JSX
```suggestion
```tsx
```

website/docs/API.md
MattAgn: nope it's called hidden for now, I aligned with asyncUtilTimeout which did not have the prefix hidden 
but now we also have defaultDebugOptions...
I think we should have the same naming pattern for the three options

website/docs/Queries.md
MattAgn: sure

website/docs/Queries.md
MattAgn: looks like a git glitch, I have already deleted it

website/docs/API.md
mdjastrzebski: Can you switch it to `defaultHidden`. Both names, `asyncUtilTImeout` and `defaultHidden`, are taken from DOM testing library: https://testing-library.com/docs/dom-testing-library/api-configuration/#defaulthidden

MattAgn: alright!

mdjastrzebski (approved):
Awesome stuff 🚀🧨

Issue #1032, opened 2022-08-02
Feature: emulate "Text strings must be rendered within a component." error


mdjastrzebski:
## Describe the Feature
React Native renderer throws error when string is rendered outside of `Text` component, e.g. directly under `View` or other native component. It seems that React Test Renderer that we use does not throw a similar error in this situation, which may lead to missed errors that RNTL could capture in tests.

This feature would emulate that behaviour by throwing error whenever string value is rendered outside of `Text` component.

## Possible Implementations

As a first step we could analyse render tree after initial render and look for `string` `children` prop that are not children of `Text` component.

Then we would have to figure out a way to react to subsequent re-renders to trigger the validation again. Maybe a custom `act` wrapper that would invoke the validation.

## Related Issues
This feature request is based on #556

pierrezimmermannbam:
Hey @mdjastrzebski, I'd be happy to help, I'll send a pr soon

mdjastrzebski:
@pierrezimmermannbam awesome! Please do :-)

pierrezimmermannbam:
I've started to work on it, I've completed the first step but I'm having some difficulties for generalizing the behavior. 
The custom act wrapper would have been nice as it is supposed to wrap all updates but there is an issue which is that we need access to the instance in order to perform the check. I thought it could be accessed through the screen object but then again there are some issues : during render the screen object has not yet been updated and on unmount it is not defined anymore. I could ignore the error of screen being undefined in act and call the method manually in the render function but I'm not sure it's good, it seems pretty complicated and not easily understandable. 
My other idea would be to call the method in the rerender and the render method which are defined on the render where we have access to the instance and to use the screen object in the fireEvent method to make the check (I believe screen should always be defined when using fireEvent ?). This is not ideal as there are three distinct calls to the method but I think it's still better than the other solution and way easier to understand.
So far I haven't thought of a better solution, @mdjastrzebski what's your opinion ?

mdjastrzebski:
@pierrezimmermannbam Indeed finding a proper hook is rather difficult. Maybe we could wrap rendered user component inside 'React.Profiler' which has 'onRender' event that should be called after each render anywhere among its descendants. At this moment the component tree should probably be already updated, but we need to check that.

If this works we may also analyze how exactly 'React.Profiler' works  in order fine tune that solution.

pierrezimmermannbam:
@mdjastrzebski using a Profiler worked ! I opened a pr

PR #967, opened 2022-05-01
Fix translations text not being found by getByText

pierrezimmermannbam:
### Summary

Pr is still a draft. Fixes #937

PR #1070, opened 2022-08-20
Feat : emulate string outside text error

pierrezimmermannbam:
### Summary

Throw errors when strings are rendered outside Text to simulate native behaviour (#1032 )

On initial render we make a first check and then using a React.Profiler we perform a new check on each update by using the screen object to access the instance. The check at mount is done directly in the render method because the screen object is not updated yet with the render result. In the onRender prop of the React.Profile we verify that the phase is update before performing a check so it doesn't run on mount. 
I also had to refactor a bit the render function so that this behavior would only affect the rendering of components and not of hooks because in the case of hooks the instance is not defined in the screen object, which resulted in errors on update.
 

### Test plan

I added a test case for rendering, rerendering and updating with fireEvent and checked each time that it threw the expected error


mdjastrzebski (commented):
Added some initial comments.

src/render.tsx
```diff
@@ -57,6 +66,28 @@ export default function render<T>(
...
 
+export default function renderComponent<T>(
+  component: React.ReactElement<T>,
+  { wrapper: Wrapper, createNodeMock }: Options = {}
+) {
+  const assertStringsWithinTextOnUpdate: React.ProfilerProps['onRender'] = (
+    _,
+    phase
+  ) => {
+    if (phase === 'update') {
```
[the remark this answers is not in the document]
pierrezimmermannbam: On initial render the screen object has not yet been updated with the result of the render so trying to access screen.container results in an error being thrown

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,28 @@
+import React, { ReactElement } from 'react';
+import { ReactTestInstance } from 'react-test-renderer';
+import { filterNodeByType } from './filterNodeByType';
+
+export const assertStringsWithinText = (instance: ReactTestInstance): void => {
+  const nodesWithStringChild = instance.findAll(hasNodeStringChild);
+  nodesWithStringChild.forEach((node) => {
+    const isTextComponent = filterNodeByType(node, 'Text');
```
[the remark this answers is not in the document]
pierrezimmermannbam: This works but I get a typescript error saying this will always be false as types ElementType and string do not overlap but the typing must be incorrect since the type is a string for native component. It works when I use filterNodeByType because the node is either a ReactTestInstance or a ReactComponent, I could make the modification and add a ts-expect-error but I think it's better to use filterNodeByType and not have to explain why there is a ts error

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,28 @@
...
+    const isNativeComponent = typeof node.type === 'string';
+
+    if (!isTextComponent && isNativeComponent) {
+      throw new Error('Text strings must be rendered within a component.');
```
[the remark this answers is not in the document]
pierrezimmermannbam: Definitely better, thanks !

src/render.tsx
```diff
@@ -57,6 +66,28 @@ export default function render<T>(
   return result;
 }
 
+export default function renderComponent<T>(
+  component: React.ReactElement<T>,
+  { wrapper: Wrapper, createNodeMock }: Options = {}
+) {
+  const assertStringsWithinTextOnUpdate: React.ProfilerProps['onRender'] = (
```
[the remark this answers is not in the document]
pierrezimmermannbam: The function is indeed rather simple so a shorter name may be better, I changed it

src/__tests__/render.test.tsx
```diff
@@ -285,3 +285,17 @@ test('RenderAPI type', () => {
...
   expect(true).toBeTruthy();
 });
+
+test('should throw when rendering a string outside a text component', () => {
+  expect(() => render(<View>hello</View>)).toThrowError(
+    'Text strings must be rendered within a host Text component.'
+  );
+});
+
+test('should throw an error when rerendering with text outside of Text component', () => {
```
AugustinLF, line 295: Would you mind adding a test to check that `<Text><>My test</></Text>` doesn't throw?

src/render.tsx
```diff
@@ -22,19 +24,26 @@ export type RenderResult = ReturnType<typeof render>;
...
+      ? internalWrap(<Wrapper>{element}</Wrapper>)
+      : internalWrap(element);
 
   const renderer = renderWithAct(
     wrap(component),
     createNodeMock ? { createNodeMock } : undefined
   );
   const update = updateWithAct(renderer, wrap);
   const instance = renderer.root;
+  assertStringsWithinText(instance);
```
AugustinLF, line 46: Should we consider rolling that out behind an option to start?

AugustinLF (changes requested):

src/__tests__/render.test.tsx
```diff
@@ -285,3 +285,17 @@ test('RenderAPI type', () => {
...
+  const { rerender } = render(<View />);
+
+  expect(() => rerender(<View>hello</View>)).toThrowError(
+    'Text strings must be rendered within a host Text component.'
+  );
+});
```
mdjastrzebski, line 301: We probably need to cover some other supported/not supported cases:
1. Support for components returning `string` value, e.g. i18next `Trans`, etc
```
function ReturnsString() { return 'Translated Text' };

function NestedReturnsString() { return <ReturnsStrings />: }

test("Should be fine", () => {
  render(<Text><ReturnsString /></Text>)
  render(<Text><NestedReturnsString /></Text>)
})

test("Should catch this", () => {
  expect(render(<View><ReturnsString /></View>)).toThrowError();
  expect(render(<View><NestedReturnsString /></View>)).toThrowError();
})
```
2. How does RN handle rendering `number` inside/outside of `<Text/>`. If it fails then we should match that behaviour.

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,30 @@
+import React, { ReactElement } from 'react';
+import { ReactTestInstance } from 'react-test-renderer';
+import { filterNodeByType } from './filterNodeByType';
+
+export const assertStringsWithinText = (instance: ReactTestInstance): void => {
+  const nodesWithStringChild = instance.findAll(hasNodeStringChild);
+  nodesWithStringChild.forEach((node) => {
+    const isHostTextComponent = filterNodeByType(node, 'Text');
+    const isNativeComponent = typeof node.type === 'string';
```
mdjastrzebski, line 9: This second check seems no longer needed since `finderNodeByType` verifies that `node.type` is string equal to `Text`.
```suggestion
```

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,30 @@
...
+      );
+    }
+  });
+};
+
+const hasNodeStringChild = (node: ReactTestInstance) => {
+  const children: (string | number | ReactElement)[] = node.props.children;
+  let nodeHasStringChild = false;
+
+  React.Children.forEach(children, (child) => {
```
mdjastrzebski, line 23: This `forEach` can be replaced by `Array.some` method.

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (!isHostTextComponent && isNativeComponent) {
+      throw new Error(
+        'Text strings must be rendered within a host Text component.'
+      );
+    }
+  });
+};
+
+const hasNodeStringChild = (node: ReactTestInstance) => {
```
mdjastrzebski, line 19: The logic here is kinda permissive, as it allows to have a composite component, e.g. function/class component, that returns `string` value. That component would probably pass this check.

mdjastrzebski: I wonder if we would be even able to make check, as there as problems discussed in #967, which you probably can remember @pierrezimmermannbam ;-)

src/render.tsx
```diff
@@ -22,19 +24,26 @@ export type RenderResult = ReturnType<typeof render>;
  * Renders test component deeply using react-test-renderer and exposes helpers
  * to assert on the output.
  */
-export default function render<T>(
+export function render<T>(
```
mdjastrzebski, line 27: This should be wrapped behind an `render` option, which should be disabled by default at least until next major release as that be a breaking change. Let's figure out name of that option:
`validateRenderedStrings`, `validateStringsRenderedInText`, `validateStringsOutsideText`, etc, etc?

src/render.tsx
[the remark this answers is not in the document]
mdjastrzebski: Yeah that make sense.

mdjastrzebski (changes requested):
I've added a request for some addtional test,  I have a hunch that we won't be able to simulate that perfectly until we solve #967 .

src/__tests__/fireEvent.test.tsx
```diff
@@ -475,3 +476,29 @@ test('has only onMove', () => {
...
+    );
+  }
+
+  return <View>text rendered outside text component</View>;
+};
+
+test('should throw an error when strings are rendered outside Text', () => {
+  const { getByText } = render(<ErrorComponent />);
+
+  expect(() => fireEvent.press(getByText('Display text'))).toThrowError(
```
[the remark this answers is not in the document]
pierrezimmermannbam: Yes you're right I'll move it

src/render.tsx
pierrezimmermannbam: I think validateRenderedStrings is good, should I also rename the method assertStringsWithinText ? Maybe validateStringsRenderedInText since it's an internal name it could be a bit longer but I find it more explicit

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: We're checking !isHostTextComponent & isNativeComponent so it's not equivalent to !isHostComponent, the condition would be false for a component that has a type function or object, whereas with the check isNativeComponent it would be false

src/__tests__/render.test.tsx
pierrezimmermannbam: Rendering numbers also fails, I added a test case for it. Regarding the components returning string value you're right it will pass. I guess a solution would be a similar approach to the one I tried when trying to fix the same issue for queries meaning using the JSON representation in the assertStringsWithinText. This time there wouldn't be any change on the public api, the only issue is that there is no findAll method implemented for this type but it's not much of an issue

src/__tests__/render.test.tsx
pierrezimmermannbam: The validation didn't work when the child was rendered in a fragment, eg <View><>0</></View>, I fixed it and added another test case

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,49 @@
...
+export const assertStringsWithinText = (instance: ReactTestInstance): void => {
+  const nodesWithStringChild = instance.findAll((node) =>
+    isSomeChildStringOrNumber(node.props.children)
+  );
+  nodesWithStringChild.forEach((node) => {
+    const isHostTextComponent = filterNodeByType(node, 'Text');
+    const isNativeComponent = typeof node.type === 'string';
+
+    if (!isHostTextComponent && isNativeComponent) {
+      throw new Error(
```
AugustinLF, line 14: Would it be possible to get a better error, so the user can figure out where is the problem? Overall `@testing-library/react` does a good job with its errors to help the user.

In our case, I would expect at least the string rendered to be logged, ideally we'd get s snapshot of the part of the tree that poses problem (but I'm not sure RNTL offers that possibility)

AugustinLF (changes requested):

src/helpers/assertStringsWithinText.ts
mdjastrzebski: Logging part of the tree might be difficult without having `toJSON` being exported from React Test Renderer. But logging text string should work

src/render.tsx
mdjastrzebski: `validateRenderedStrings` seems ambiguous, there is no mention about them being outside of `Text`

src/helpers/assertStringsWithinText.ts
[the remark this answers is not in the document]
mdjastrzebski:
```suggestion
    const isHostTextComponent = filterNodeByType(node, 'Text');
```
And remove `isNativeComponent`

mdjastrzebski:
The PR seems to be almost ready. We still need to pick meaningful `render` option name for it. I would suggest to prefix it with `experimentalXxx`, where `xxx` is whatever name we will choose. That way we could release it to the users, inform them that it's kinda new so might still be some unexpected errors, and also give us ability to fix them without releasing major version, as typically in OSS experimental features breaking changes are not considered breaking change for the whole dependency. @pierrezimmermannbam @AugustinLF @thymikee Wdyt?

src/__tests__/render.test.tsx
mdjastrzebski: If we ship this as experiemental then we can live with imperfect but still usable check. Seem like missing `toJSON` would solve all our problems

src/helpers/assertStringsWithinText.ts
mdjastrzebski: Can we rename it as `isHostComponent` because seems like `host` and `native` refere to same thing, so it would make sense to use the same pattern as more readable.

src/render.tsx
```diff
@@ -57,6 +76,35 @@ export default function render<T>(
...
+  }
+
+  const handleRender: React.ProfilerProps['onRender'] = (_, phase) => {
+    if (phase === 'update') {
+      assertStringsWithinText(screen.container);
+    }
+  };
+
+  const wrap = (innerElement: React.ReactElement) => (
+    <Profiler id="renderProfiler" onRender={handleRender}>
```
mdjastrzebski, line 94: ❗️ @pierrezimmermannbam 
We would want to avoid wrapping in `<Profiler>` if the user does not enable this option. So for regular renders the behaviour stays exactly the same.

src/__tests__/render.test.tsx
mdjastrzebski: I would keep the explicit fragment check until `toJSON` is avialable though

src/render.tsx
pierrezimmermannbam: If the option is not provided we return render without wrapping with a Profiler, we have a check on the option both in the renderComponent function (for component updates) and in the render function (for check after mount)

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: Yes I think logging the string that is rendered is a good idea, I don't see how we could log a part of the tree though

src/__tests__/render.test.tsx
pierrezimmermannbam: In this case we wouldn't need the export of toJSON I think, we have access to it in the result of the render function as well as in the screen object so this should be doable

pierrezimmermannbam: I made the change, it now works with components returning strings and the implementation is way simpler

pierrezimmermannbam:
I changed the implementation to use toJSON, the validation should work fine for every case now, what remains to be done is deciding the name for the option and customizing the error message so that it's clearer. I don't know about logging a part of the tree but I'd say adding the string that is being rendered outside a Text component would be a good start. As for the name of the option, experimentalValidateStringsRenderedInText as suggested by @mdjastrzebski looks good to me

pierrezimmermannbam:
Also there is a small issue when runing tests, the React.Profiler logs errors so there are outputs of errors from the tests. This could be fixed by overriding console.error in the tests that trigger errors so that it doesn't do anything. It's not ideal though, maybe there is a better way to do this ? 

Capture d’écran 2022-08-30 à 19 17 36

AugustinLF:
> Also there is a small issue when running tests, the React.Profiler logs errors so there are outputs of errors from the tests. This could be fixed by overriding console.error in the tests that trigger errors so that it doesn't do anything. It's not ideal though, maybe there is a better way to do this ?

I'd say it's fair in this case to override console.error and add an explicit filter for the error containing `<Profiler> component` or something like that

pierrezimmermannbam:
> > Also there is a small issue when running tests, the React.Profiler logs errors so there are outputs of errors from the tests. This could be fixed by overriding console.error in the tests that trigger errors so that it doesn't do anything. It's not ideal though, maybe there is a better way to do this ?
> 
> I'd say it's fair in this case to override console.error and add an explicit filter for the error containing `<Profiler> component` or something like that

I have overriden console.error in the tests that logged errors so that it doesn't log errors from the Profiler

src/render.tsx
```diff
@@ -1,40 +1,59 @@
...
 import debugDeep from './helpers/debugDeep';
 import { getQueriesForElement } from './within';
-import { setRenderResult } from './screen';
+import { setRenderResult, screen } from './screen';
+import { assertStringsWithinText } from './helpers/assertStringsWithinText';
 
 type Options = {
   wrapper?: React.ComponentType<any>;
   createNodeMock?: (element: React.ReactElement) => any;
+  experimentalValidateStringsRenderedInText?: boolean;
```
mdjastrzebski, line 16: Seems like React Test Render is using `unstable_` prefix for experimental options, seems like a good convention to follow:
```suggestion
  unstable_validateStringsRenderedInText?: boolean;
```
See:  [unstable_isConcurrent: boolean,](https://github.com/facebook/react/blob/cae635054e17a6f107a39d328649137b83f25972/packages/react-test-renderer/src/ReactTestRenderer.js#L64)

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,28 @@
+import { ReactTestRendererNode } from 'react-test-renderer';
+
+export const assertStringsWithinText = (
+  rendererJSON: ReactTestRendererNode | null | ReactTestRendererNode[]
```
mdjastrzebski, line 4:
```suggestion
  node: ReactTestRendererNode | ReactTestRendererNode[] | null
```

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (typeof rendererJSON === 'string') {
+    return;
+  }
+
+  if (rendererJSON.type !== 'Text') {
+    if (rendererJSON.children?.some((child) => typeof child === 'string')) {
+      throw new Error(
+        'Text strings must be rendered within a host Text component.'
+      );
+    }
```
mdjastrzebski, line 22: What about `number` is that changed to `string` by `toJSON` method?

src/helpers/assertStringsWithinText.ts
mdjastrzebski: I wonder if we should accept just `ReactTestRendererNode | null` or even `ReactTestRendererNode`? Do we really need `ReactTestRendererNode[]`?

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (Array.isArray(rendererJSON)) {
+    rendererJSON.forEach(assertStringsWithinText);
+    return;
+  }
+
+  if (typeof rendererJSON === 'string') {
+    return;
+  }
+
+  if (rendererJSON.type !== 'Text') {
```
mdjastrzebski, line 17: Looks like having `<Text>` disables any descending down in the tree search. Does RN actually supports something like `<Text><View>AAA</View></Text>`? Is putting `<View> ` inside `<Text>` allowed in RN runtime?

src/__tests__/render.test.tsx
mdjastrzebski: Awesome that you managed to use `toJSON` 💯 , I thought that we would be limited by not being able to call `toJSON` on given node. It makes sense that we can do all the assertion on JSON tree instead of `ReactTestInstance` tree.

src/__tests__/render.test.tsx
```diff
@@ -285,3 +287,133 @@ test('RenderAPI type', () => {
...
+  console.error = (errorMessage: string) => {
+    if (
+      !errorMessage.includes(
+        'The above error occurred in the <Profiler> component'
+      )
+    ) {
+      originalConsoleError(errorMessage);
+    }
+  };
+  const { rerender } = render(<View />, {
```
mdjastrzebski, line 310: What will happen in user test? Will they also log `'The above error occurred in the <Profiler> component'`?

src/__tests__/render.test.tsx
```diff
@@ -285,3 +287,133 @@ test('RenderAPI type', () => {
...
+  // eslint-disable-next-line no-console
+  console.error = originalConsoleError;
+});
+
+const ErrorComponent = () => {
+  const [shouldDisplayText, setShouldDisplayText] = React.useState(false);
+
+  if (!shouldDisplayText) {
+    return (
+      <TouchableOpacity
```
mdjastrzebski, line 326: Pls use `<Pressable>` in tests, as recommended future-proof alternative to `TouchableXxx`

src/__tests__/render.test.tsx
```diff
@@ -285,3 +287,133 @@ test('RenderAPI type', () => {
...
+
+  return <View>text rendered outside text component</View>;
+};
+
+test('should throw an error when strings are rendered outside Text', () => {
+  // eslint-disable-next-line no-console
+  console.error = (errorMessage: string) => {
+    if (
+      !errorMessage.includes(
+        'The above error occurred in the <Profiler> component'
```
mdjastrzebski, line 344: Can we extract that error text to some const, it should limit the line wrapping

src/__tests__/render.test.tsx
```diff
@@ -285,3 +287,133 @@ test('RenderAPI type', () => {
...
+
+test('it should throw if a number is rendered outside a text', () => {
+  expect(() =>
+    render(<View>0</View>, { experimentalValidateStringsRenderedInText: true })
+  ).toThrowError('Text strings must be rendered within a host Text component.');
+});
+
+const Trans = ({ i18nKey }: { i18nKey: string }) => <>{i18nKey}</>;
+
+test('it should throw with components returning string value not rendered in Text', () => {
```
mdjastrzebski, line 410: Pls add test for
```ts
render(
      <Text>
        <Trans i18nKey="hello" />
      </Text>,
      { experimentalValidateStringsRenderedInText: true }
    )
```

not throwing

mdjastrzebski (commented):
Awesome work @pierrezimmermannbam ! Seems like your nearing to the end. I really like you were able to use JSON tree to run the assertion code. 

On major thing that brings my attention is the complexity around `render`/`renderComponent` function. I have suggest an approach that would allow us to use the original `render` while having a separete`renderWithValidateStrings`, instead of current more complex flow.

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: The return type of the toJson function is TestRenderer.ReactTestRendererJSON | TestRenderer.ReactTestRendererJSON[] | null, so we need it, I refactor the assertStringsWithinText a bit to break it in two so the first one handles all types one the second one only accepts a ReactTestRendererNode, I think it's clearer that way, is it ok for you ?

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: Yes it is, there is a test case to ensure rendering numbers outside text components does throw

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: Oh I thought it was allowed but it actually isn't, good catch ! I'll fix it and add a test case for it

pierrezimmermannbam: fixed

src/render.tsx
```diff
@@ -1,40 +1,59 @@
...
  * Renders test component deeply using react-test-renderer and exposes helpers
  * to assert on the output.
  */
-export default function render<T>(
-  component: React.ReactElement<T>,
-  { wrapper: Wrapper, createNodeMock }: Options = {}
-) {
-  const wrap = (innerElement: React.ReactElement) =>
-    Wrapper ? <Wrapper>{innerElement}</Wrapper> : innerElement;
+export function render<T>({
```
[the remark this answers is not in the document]
pierrezimmermannbam: I agree the exports are confusing, I initially didn't want to change the public api of the render function but since there is now an option it's not an issue anymore, I changed it so that there is a single exported function and extracted what was shared in another function, there is still a bit of duplicated code but not that much and it is indeed way clearer imo

src/__tests__/render.test.tsx
```diff
@@ -285,3 +281,160 @@ test('RenderAPI type', () => {
...
+    if (!errorMessage.includes(profilerErrorMessage)) {
+      originalConsoleError(errorMessage);
+    }
+  };
+  const { getByText } = render(<ErrorComponent />, {
+    unstable_validateStringsRenderedInText: true,
+  });
+
+  expect(() => fireEvent.press(getByText('Display text'))).toThrowError(
+    'Text strings must be rendered within a host Text component.'
```
AugustinLF, line 346: Why does this message doesn't contain string + node type like the other tests (`"Text strings must be rendered within a host Text component. Text 0 was rendered in a View"`)?

src/__tests__/render.test.tsx
```diff
@@ -285,3 +281,160 @@ test('RenderAPI type', () => {
...
+  expect(() =>
+    render(
+      <View>
+        <Text>hello</Text>
+        hello
+      </View>,
+      { unstable_validateStringsRenderedInText: true }
+    )
+  ).toThrowErrorMatchingInlineSnapshot(
+    `"Text strings must be rendered within a host Text component. Text hello was rendered in a View"`
```
AugustinLF, line 379: Lovely error message.

I'm curious btw, what does the stacktrace looks like?

src/__tests__/render.test.tsx
pierrezimmermannbam: Oh I forgot to replace it with an inline snapshot, toThrowErrror actually works when the tested string is included in the error message so this works even though the actual message is more complete

src/__tests__/render.test.tsx
pierrezimmermannbam: The stack trace looks like this :

Capture d’écran 2022-09-13 à 20 14 37

And when the error occurs during a rerender you get this additional warning that an error occurred in the React Profiler

Capture d’écran 2022-09-13 à 20 18 39

AugustinLF: Thanks!

AugustinLF (approved):
Stellar work! Really looking forward using that.

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,34 @@
...
+const assertStringsWithinTextForNode = (node: ReactTestRendererNode) => {
+  if (typeof node === 'string') {
+    return;
+  }
+
+  if (node.type !== 'Text') {
+    node.children?.forEach((child) => {
+      if (typeof child === 'string') {
+        throw new Error(
+          `Text strings must be rendered within a host Text component. Text ${child} was rendered in a ${node.type}`
```
mdjastrzebski, line 25:
```suggestion
          `Text strings must be rendered within a host Text component. Text "${child}" was rendered in a ${node.type}`
```

mdjastrzebski: Adding quotes should improve readability of the message.

mdjastrzebski: Maybe we should use the original wording of RN error: `Invariant Violation: Text strings must be rendered within a <Text> component`? and append a bit more details so that the final error message could be like:

```
Invariant Violation: Text strings must be rendered within a <Text> component.

Detected attempt to render "${child}" string within a <${node.type}> component.
```
Wdyt?

src/helpers/assertStringsWithinText.ts
```diff
@@ -0,0 +1,34 @@
+import { ReactTestRendererNode } from 'react-test-renderer';
+
+export const assertStringsWithinText = (
```
mdjastrzebski, line 3: The current structure consists of two recursive functions calling both themselves and each other.

The issue here results probably from the fact that there is no `parent` property on `toJSON` outputed nodes (unlike `ReactTestInstances`), as a workaround we could pass both parent and child to assertion function:

```ts
export function assertStringsWithinText(root: ReactTestRendererNode | Array<ReactTestRendererNode> | null) {
  if (!root) return;

  if (Array.isArray(root)) {
    root.forEach(child => _assertStringsWithinTextInternal(root, null));
    return;
  }

  _assertStringsWithinTextInternal(root, null);
}

function _assertStringsWithinTextInternal(node: ReactTestRendererNode, parent: ReactTestRendererNode) {
  if (typeof node === 'string' && parent.type !== "Text") {
    throw Error("Invariant violation: Text .... ");
  } 

  if (node.children) {
    node.children.forEach(child => _assertStringsWithinTextInternal(child, node));
  }
}
```

In the above setup only the internal function is recursive (to itself), the export functions servers as entry point to recursive one, but it only called once. 

Note that there is a slight difference in behaviour, as the above code would throw for root string value, which I think is reasonable, as that string value is not rendered inside `<Text>` component.

Wdyt?

src/render.tsx
```diff
@@ -25,17 +28,65 @@ export type RenderResult = ReturnType<typeof render>;
...
+
+  const wrap = (element: React.ReactElement) =>
+    Wrapper ? <Wrapper>{element}</Wrapper> : element;
 
   const renderer = renderWithAct(
     wrap(component),
     createNodeMock ? { createNodeMock } : undefined
   );
+
+  return enrichRenderResult(renderer, wrap);
```
mdjastrzebski, line 52: `enrich` naming here is too generic, maybe something along the line of:
```suggestion
  return buildRenderResult(renderer, wrap);
```

src/render.tsx
```diff
@@ -25,17 +28,65 @@ export type RenderResult = ReturnType<typeof render>;
...
+    if (phase === 'update') {
+      assertStringsWithinText(screen.toJSON());
+    }
+  };
+
+  const wrap = (innerElement: React.ReactElement) => (
+    <Profiler id="renderProfiler" onRender={handleRender}>
+      {Wrapper ? <Wrapper>{innerElement}</Wrapper> : innerElement}
+    </Profiler>
+  );
```
mdjastrzebski, line 72: Slight tweak to match `render`
```suggestion
  const wrap = (element: React.ReactElement) => (
    <Profiler id="renderProfiler" onRender={handleRender}>
      {Wrapper ? <Wrapper>{element}</Wrapper> : innerElement}
    </Profiler>
  );
```

src/helpers/assertStringsWithinText.ts
mdjastrzebski: nit: BTW since we use `validate` in both option name `unstable_validateStringsRenderedInText` as well as `renderWithStringValidation` then maybe we should call this:

```
function validateStringsWithinText(...)
```

Wdyt?

mdjastrzebski, line 3: nit: lets match naming with our option name and `renderWithStringValidation`.
```
function validateStringsRenderedInText()
// or 
function validateStringsWithinText()
// or 
function validateStringsRenderedWithinText()
```

src/render.tsx
```diff
@@ -1,16 +1,19 @@
...
 import debugDeep from './helpers/debugDeep';
 import { getQueriesForElement } from './within';
-import { setRenderResult } from './screen';
+import { setRenderResult, screen } from './screen';
+import { assertStringsWithinText } from './helpers/assertStringsWithinText';
 
 export type RenderOptions = {
   wrapper?: React.ComponentType<any>;
   createNodeMock?: (element: React.ReactElement) => any;
+  unstable_validateStringsRenderedInText?: boolean;
```
mdjastrzebski, line 16: Seems like React Natives uses `within` in this context for the original error message: `Invariant Violation: Text strings must be rendered within a <Text> component`

```suggestion
  unstable_validateStringsRenderedWithinText?: boolean;
```

src/__tests__/render.test.tsx
mdjastrzebski: I guess we cannot do about the `<Profiler>` part being logged in case of error.

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: I totally agree, I changed the error message and it looks very clear now, thanks ! 

Capture d’écran 2022-09-18 à 13 10 21

However the multiline inline snapshots don't fit well in vs code so I replaced them with regular snapshots

src/helpers/assertStringsWithinText.ts
pierrezimmermannbam: Oh good catch, actually it's just a mistake, the second function is intented to call only itself I just mixed up and didn't notice since it worked. Calling directly assertStringsWithinTextForNode instead of assertStringsWithinText will make no difference since calling assertStringsWithinText with a ReactTestRendererNode calls  assertStringsWithinTextForNode, so I just made the replacement

pierrezimmermannbam: Good idea ! I renamed it validateStringsRenderedWithinText

src/__tests__/render.test.tsx
```diff
@@ -283,3 +279,154 @@ test('RenderAPI type', () => {
...
+  ).not.toThrow();
+});
+
+test('it should not throw if option validateRenderedString is false', () => {
+  expect(() => render(<View>hello</View>)).not.toThrow();
+});
+
+test(`it should throw 
+      - when one of the children is a text
+      - and the parent is not a Text component`, () => {
```
mdjastrzebski, line 365:
```suggestion
test(`it should throw when one of the children is a text and the parent is not a Text component`, () => {
```

src/__tests__/render.test.tsx
```diff
@@ -283,3 +279,154 @@ test('RenderAPI type', () => {
...
+        <Text>hello</Text>
+        hello
+      </View>,
+      { unstable_validateStringsRenderedWithinText: true }
+    )
+  ).toThrowErrorMatchingSnapshot();
+});
+
+test(`it should throw 
+    - when a string is rendered within a fragment rendered outside a Text`, () => {
```
mdjastrzebski, line 378: ```suggestion when a string is rendered within a fragment rendered outside a Text`, () => {
```

mdjastrzebski:
I've extracted the tests into separate `render-stringValidation.test.tsx` file, as the original `render.test.tsx` got pretty long. I've also replaced error snapshots/inline snapshots with cleaner `toThrow` checks.

mdjastrzebski:
Awesome work @pierrezimmermannbam! Merging it 🚀

mdjastrzebski (approved):

mdjastrzebski:

PR #1101, opened 2022-09-08
chore: optional jest peer dep

mdjastrzebski:
### Summary

Make Jest 28 an optional peer dep.

### Test plan

All checks should pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,13 @@
   "peerDependencies": {
     "react": ">=16.0.0",
     "react-native": ">=0.59",
-    "react-test-renderer": ">=16.0.0"
+    "react-test-renderer": ">=16.0.0",
+    "jest": ">=28.0.0"
+  },
+  "peerDependenciesMeta": {
+    "jest": {
+      "optional": true
+    }
   },
   "scripts": {
     "clean": "del build",


package.json
```diff
@@ -40,28 +40,34 @@
...
   "peerDependencies": {
     "react": ">=16.0.0",
     "react-native": ">=0.59",
-    "react-test-renderer": ">=16.0.0"
+    "react-test-renderer": ">=16.0.0",
+    "jest": ">=28.0.0"
+  },
+  "peerDependenciesMeta": {
+    "jest": {
+      "optional": true
```
thymikee, line 69: does this work with npm too? otherwise jest will be installed as a part of npm install

mdjastrzebski: Seems like yes: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta

thymikee: FYI. npm 8 is available since node 16. node 14 and 15 has npm 7

mdjastrzebski: Npm v7 docs also mentions support for `peerDepsMeta` + `optional`: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

thymikee (approved):

friederbluemle:
Is there anything in RNTL that *requires* Jest 28 or higher? Or does it work with Jest 27 and 26 as well? Asking because only RN 0.71 and later come with Jest 29. Prior to that the RN template shipped with Jest 26.
If there is no strict dependency, can this be loosened (preferred)? Of if there is, do we have a RN-RNTL compatibility matrix?

mdjastrzebski:
@friederbluemle You should try updating to Jest 29 regardless of your RN version, as afaik these are quite independent and older RN version should work the latest Jest, despite the fact that RN template ships given versions of Jest by default.

If that does not work for you, then afair we require Jest 28 when you use fake timers. You'll find more info in our [v11 migration guide](https://callstack.github.io/react-native-testing-library/docs/migration-v11). If you do not use any fake timers, then Jest will **probably** work fine.

PR #1114, opened 2022-09-13
chore: add test to ensure it is possible to trigger native event

AugustinLF:
This PR adds tests for a behaviour we support (triggering event that don't exist on the JS side but are handled natively). It is however not tested, so I want to add test to make sure we don't lose this feature by mistake (see discussions in https://github.com/callstack/react-native-testing-library/pull/1080#issuecomment-1241207946).

### DIFF
--- a/src/__tests__/fireEvent.test.tsx
+++ b/src/__tests__/fireEvent.test.tsx
@@ -475,3 +475,53 @@ test('has only onMove', () => {
   });
   expect(handleDrag).toHaveBeenCalled();
 });
+
+// Those events ideally should be triggered through `fireEvent.scroll`, but they are handled at the
+// native level, so we need to support manually triggering them
+describe('native events', () => {
+  test('triggers onScrollBeginDrag', () => {
+    const onScrollBeginDragSpy = jest.fn();
+    const { getByTestId } = render(
+      <ScrollView testID="test-id" onScrollBeginDrag={onScrollBeginDragSpy} />
+    );
+
+    fireEvent(getByTestId('test-id'), 'onScrollBeginDrag');
+    expect(onScrollBeginDragSpy).toHaveBeenCalled();
+  });
+
+  test('triggers onScrollEndDrag', () => {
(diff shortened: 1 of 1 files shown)


src/__tests__/fireEvent.test.tsx
```diff
@@ -475,3 +475,53 @@ test('has only onMove', () => {
   });
   expect(handleDrag).toHaveBeenCalled();
 });
+
+// Those events ideally should be triggered through `fireEvent.scroll`, but they are handled at the
+// native level, so we need to support manually triggering them
+describe('native events', () => {
```
mdjastrzebski, line 481:
```suggestion
describe('native event support', () => {
```

src/__tests__/fireEvent.test.tsx
```diff
@@ -475,3 +475,53 @@ test('has only onMove', () => {
...
+      <ScrollView
+        testID="test-id"
+        onMomentumScrollEnd={onMomentumScrollEndSpy}
+      />
+    );
+
+    fireEvent(getByTestId('test-id'), 'onMomentumScrollEnd');
+    expect(onMomentumScrollEndSpy).toHaveBeenCalled();
+  });
+});
```
mdjastrzebski, line 527: Should we need to add some other native events to this list?

src/__tests__/fireEvent.test.tsx
mdjastrzebski: I wonder if we should actually use `host component event support` instead, as by design we do not support "native" events.

src/__tests__/fireEvent.test.tsx
AugustinLF: We could? I'm unsure why ones are relevant, I know that those ones are events that I use. The goal for this PR was to add some coverage to the feature, so didn't go further than that.

If we want to be exhaustive, we could also add something like `fireEvent(element, 'eventNameThatDoesNotExist)`, because such event would be triggered.

mdjastrzebski (approved):
Looks good!

Issue #1093, opened 2022-09-04
Unexpected failing test when mixing promises and useEffect


AugustinLF:
## Describe the bug

I've written a small test case demonstrating issues I've had when combining `user event (press) -> promise -> setState -> useEffect`. The test case fails for RNTL but passes on `@testing-library/react`.

```tsx
test.only('async event with fireEvent', async () => {
  const Comp = ({ onPress }: { onPress: () => void }) => {
    const [state, setState] = React.useState(false);

    React.useEffect(() => {
      console.log('useEffect');
      if (state) {
        console.log('onPress');
        onPress();
      }
    }, [state, onPress]);

    return (
      <Pressable
        onPress={async () => {
          console.log('waiting');
          await Promise.resolve();
          console.log('setting');
          setState(true);
        }}
      >
        <Text>Trigger</Text>
      </Pressable>
    );
  };

  const spy = jest.fn();
  const { getByText } = render(<Comp onPress={spy} />);

  fireEvent.press(getByText('Trigger'));

  await waitFor(() => {
    console.log('checking');
    expect(spy).toHaveBeenCalled();
  });
});
```

Logs go:
```
useEffect
waiting
checking
setting
checking
... a bunch of other `checking`
checking
useEffect
onPress
```

Note that the final `onPress` call (the one that we're waiting for) happens after the last check.

## Expected behavior

It works as expected on web, with those logs
```
useEffect
waiting
checking
setting
useEffect
onPress
checking
```

I'll do more investigation to figure out what's happening, but if anyone manage to fix it, feel free to do so :) I kind of expect this issue to be related to a lot of the ones we've had around problems with `waitFor`, or `async act ...`.

## Versions

Latest `main`, commit of `cb46ae2689cc661879e8609766041bf93f1d3137`.

AugustinLF:
Changing the way we handle async update to the way it's handled in the newer versions of `@testing-library/react`, i.e. dropping the `await act` and changing the global `IS_REACT_ACT_ENVIRONMENT` variable works (not changing the global variable still works but leads to some `console.error`.

https://github.com/testing-library/react-testing-library/pull/1031/files#diff-2ef28f1bd92d5dcd1f2a04d56814d3adaee10cc939b4a7d7c861af3a3cbbccb7R22

I'll need to investigate a bit more on what are the consequences of this change. @mdjastrzebski we officially support `react@>16`, right? Have we considered running both `react@17` and `react@18` in the CI?

[...]

PR #1115, opened 2022-09-13
Add support for testing with react@17

AugustinLF:
### Summary

Follow-up on the discussion that happened in https://github.com/callstack/react-native-testing-library/issues/1093#issuecomment-1240852061. If we want to change the way we work with act and be more aligned with RTL, we either need to drop support to `react@17` (which I'd like to avoid doing), or we need to increase our confidence that we're not breaking older versions.

I don't think we really need to cover for `react@<=16` because tests mostly behave the same way, but when trying to work on bumping react in the codebase I work on, _a lot_ of tests break for subtle reasons.

### Approach

I went with a fairly simple approach where I copy the whole test/code base to be able to run any PR changes on both platforms (which prevents us from using the `examples/` approach that reaches to a published version of RNTL).

Since different versions of `react` also implies different versions of `react-native`, we have a problem where some tests can't work in both cases. Currently the issue seems to be only with snapshots. I don't have any brilliant idea, a solution would be to do runtime check for `React.version` and have a snapshot by version. We could replace those specific snapshots with something that works in both versions, but I think we'll run in this problem regardless.


mdjastrzebski:
@AugustinLF I like the idea of this PR, so that we are able to test our compatibility with React 17/older React Native. 

I wonder however if copying files is the right approach, maybe we could have a (shell/sed/etc) script that would update `react` + `rest-test-renderder` + `react-native` to specified version in place, then run `yarn install` & `yarn test`. Not sure how feasible is that but if so, we could even have a list of of supported R/RTR/RN versions. 

Also can we use the latest Jest with older versions of R/RN? I think yes, but we'll need to check that.

yarn.lock
```diff
@@ -1581,48 +1614,48 @@
...
+    "@react-native-community/cli-tools" "^9.1.0"
     chalk "^4.1.2"
     hermes-profile-transformer "^0.0.6"
     ip "^1.1.5"
 
-"@react-native-community/cli-platform-android@^9.0.0":
-  version "9.0.0"
-  resolved "https://registry.yarnpkg.com/@react-native-community/cli-platform-android/-/cli-platform-android-9.0.0.tgz#c21b26f456c568687c0e58a6e42ba8b11b607b8a"
-  integrity sha512-4Rp5OUZW/7Qc9hyCd+ZEikuu2k9dW3ssu6KzWygbVc9wY80i4GQmvjfsiUi21o3uPDvL4KUMANmnQqoTOIcVMA==
+"@react-native-community/cli-platform-android@^9.0.0", "@react-native-community/cli-platform-android@^9.1.0":
```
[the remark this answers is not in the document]
AugustinLF: (left overs from initial debugging, will remove)

AugustinLF:
@mdjastrzebski went with your suggestion, let me know how you feel about this solution :)

scripts/test_react_17
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+cp yarn.lock tmp.yarn.lock
+cp package.json tmp.package.json
+
+yarn add react@17.0.2 react-test-renderer@17.0.2 react-native@0.68.1 --ignore-scripts
```
AugustinLF, line 6: I hardcoded dependencies, but I assume we should be able to test other combination if we want, like different jest versions

scripts/test_react_17
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
+
+cp yarn.lock tmp.yarn.lock
+cp package.json tmp.package.json
```
mdjastrzebski, line 4:
```suggestion
cp yarn.lock yarn.lock.backup
cp package.json package.json.backup
```

src/__tests__/render.test.tsx
```diff
@@ -205,7 +205,7 @@ test('debug changing component', () => {
 test('renders options.wrapper around node', () => {
   type WrapperComponentProps = { children: React.ReactNode };
   const WrapperComponent = ({ children }: WrapperComponentProps) => (
-    <SafeAreaView testID="wrapper">{children}</SafeAreaView>
+    <View testID="wrapper">{children}</View>
```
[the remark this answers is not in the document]
mdjastrzebski: Yeah snapshots can create potential problems, as different versions of react might generate different output. For react 17 tests we could run tests using `jest --updateSnapshot` option. It's not ideal, but changing the existing rendered JSX could result failures in the future, if render output for `View` changes even slightly.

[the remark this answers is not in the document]
mdjastrzebski: I guess that proper solution here is to replace snapshots with explicit `expect` assertions that would check what the test author really had on his mind. Although I would suggest moving that in a separate PR, and just updating snapshots for R17 in this PR.

scripts/test_react_17
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh -e
```
mdjastrzebski, line 1: env seems to be recommended approach
```suggestion
#!/usr/bin/env bash
```

package.json
```diff
@@ -82,7 +82,8 @@
     "build:ts": "tsc --build tsconfig.release.json",
     "build:ts:watch": "yarn build:ts --watch --preserveWatchOutput",
     "build": "yarn clean && yarn build:js && yarn build:ts && yarn copy-flowtypes",
-    "prepare": "yarn build"
+    "prepare": "yarn build",
+    "test:react:17": "scripts/test_react_17"
```
mdjastrzebski, line 86: Since we are RN TL maybe we should rather name our test script with `React Native version, e.g. `test-react-native-0.68.3`, @AugustinLF wdyt?

scripts/test_react_17
mdjastrzebski, line 6: lets use the latest patch release from 0.68.x
```suggestion
yarn add react@17.0.2 react-test-renderer@17.0.2 react-native@0.68.3 --ignore-scripts
```

mdjastrzebski (commented):
I like the current approach with modifying package.json and restoring changes afterwards. The current solution looks pretty solid. I've added some suggestions for improvint the reliability

package.json
mdjastrzebski, line 86: Let's put this `test:xx` next to other `test:xx` scripts

mdjastrzebski: BTW since we're invoking `test:ci` underneath, so maybe `test:ci:react-native-0.68.3`? Wdyt?

src/__tests__/render.test.tsx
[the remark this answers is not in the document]
AugustinLF: Yes I don't think `--updateSnapshot` is safe. Ideally we'd want less snapshot, even though those are pretty useful for the `debug()` tests.

scripts/test_react_17
AugustinLF: > Will this overwrite existing versions of react?

Yes.

> BTW these should be devDeps as currently we put them as such

Will fix that

package.json
AugustinLF: We are RN TL, but I would argue that here we are actually testing the react differences, not as much the react-native ones (and we're still using react, I'd argue the problem is that `@testing-library/react` should be renamed `@testing-library/react-dom`^^')

AugustinLF:
@mdjastrzebski I implemented all your suggestions (or answered in the comments), let me know if you need anything else. If not I think we can merge this so I can start working on the `act` related changes for react@18 :)

scripts/test_react_17
```diff
@@ -0,0 +1,12 @@
...
+cp yarn.lock yarn.lock.backup
+cp package.json package.json.backup
+
+yarn add -D react@17.0.2 react-test-renderer@17.0.2 react-native@0.68.3 --ignore-scripts
+yarn test:ci
+
+mv package.json.backup package.json
+mv yarn.lock.backup yarn.lock
+
+yarn --ignore-scripts
```
mdjastrzebski, line 12:
```suggestion
yarn --ignore-scripts

```

mdjastrzebski (approved):

mdjastrzebski:

PR #1127, opened 2022-09-18
Add support for ByRole with name

### LINKED ISSUE #827: getByRole should accept a second argument to refine query as in react-testing-library
## Describe the Feature
Should be able to query a Button element as follows:

`getByRole('button', {name: /submit/i})`

At the present time, this is impossible so I have to add redundant `accessiblityLabel` props, for example:

`<Button accessibilityLabel="submit-button">Submit</Button>`

And then, use `queryByA11yLabel('submit-button')`

AugustinLF:
### Summary

Folllow up of #875. I started from scratch given the amount of conflicts.

Closes #827

If merged as is, don't forget to mention @kiranjd as part of the release.

Kudo to @MattAgn's work in #977 which made that a breeze. Look at the diff between this PR and the previously opened one, the code reorg was def the right call.

### DIFF
--- a/src/queries/__tests__/role.test.tsx
+++ b/src/queries/__tests__/role.test.tsx
@@ -79,8 +79,8 @@ test('getAllByRole, queryAllByRole, findAllByRole', async () => {
   );
 });
 
-describe('*ByRole with a name', () => {
-  test('Find an element that has the corresponding role and a children with the name', () => {
+describe('supports name option', () => {
+  test('returns an element that has the corresponding role and a children with the name', () => {
     const { getByRole } = render(
       <TouchableOpacity accessibilityRole="button" testID="target-button">
         <Text>Save</Text>
@@ -93,7 +93,21 @@ describe('*ByRole with a name', () => {
     );
   });
 
-  test('Find an element that has the corresponding role and a children with a matching accessibilityLabel', () => {
+  test('returns an element that has the corresponding role when several children include the name', () => {
+    const { getByRole } = render(
+      <TouchableOpacity accessibilityRole="button" testID="target-button">
+        <Text>Save</Text>
+        <Text>Save</Text>
+      </TouchableOpacity>
+    );
+
+    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
+    expect(getByRole('button', { name: 'Save' }).props.testID).toBe(
+      'target-button'
+    );
+  });
+
+  test('returns an element that has the corresponding role and a children with a matching accessibilityLabel', () => {
     const { getByRole } = render(
       <TouchableOpacity accessibilityRole="button" testID="target-button">
         <Text accessibilityLabel="Save" />
@@ -106,7 +120,7 @@ describe('*ByRole with a name', () => {
     );
   });
 
-  test('Find an element that has the corresponding role and a matching accessibilityLabel', () => {
+  test('returns an element that has the corresponding role and a matching accessibilityLabel', () => {
     const { getByRole } = render(
       <TouchableOpacity
         accessibilityRole="button"
@@ -117,21 +131,6 @@ describe('*ByRole with a name', () => {
 
     // assert on the testId to be sure that the returned element is the one with the accessibilityRole
     expect(getByRole('button', { name: 'Save' }).props.testID).toBe(
-      'target-button'
-    );
-  });
-
-  test('Can find by name using a regex', () => {
-    const { getByRole } = render(
-      <TouchableOpacity
-        accessibilityRole="button"
-        testID="target-button"
-        accessibilityLabel="Save"
-      ></TouchableOpacity>
-    );
-
-    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
-    expect(getByRole('button', { name: /Save/ }).props.testID).toBe(
       'target-button'
     );
   });
--- a/src/queries/__tests__/role.test.tsx
+++ b/src/queries/__tests__/role.test.tsx
@@ -42,6 +42,7 @@ test('getByRole, queryByRole, findByRole', async () => {
   expect(() => getByRole(NO_MATCHES_TEXT)).toThrow(
     getNoInstancesFoundMessage(NO_MATCHES_TEXT)
   );
+
   expect(queryByRole(NO_MATCHES_TEXT)).toBeNull();
 
   expect(() => getByRole('link')).toThrow(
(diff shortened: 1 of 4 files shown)


src/queries/__tests__/role.test.tsx
```diff
@@ -77,3 +78,61 @@ test('getAllByRole, queryAllByRole, findAllByRole', async () => {
     getNoInstancesFoundMessage(NO_MATCHES_TEXT)
   );
 });
+
+describe('*ByRole with a name', () => {
```
mdjastrzebski, line 82:
```suggestion
describe('supports name option', () => {
```

website/docs/Queries.md
```diff
@@ -189,6 +189,10 @@ render(<MyComponent />);
 const element = screen.getByRole('button');
 ```
 
+#### Options
+
+`name`: Finds an element with given `accessibilityRole` and an accessible name (equivalent to `ByText` or `queryByLabelText`).
```
mdjastrzebski, line 194:
```suggestion
`name`: Finds an element with given `accessibilityRole` and an accessible name (equivalent to `byText` or `byLabelText` query).
```

typings/index.flow.js
```diff
@@ -202,6 +202,11 @@ interface UnsafeByPropsQueries {
     | Array<ReactTestInstance>
     | [];
 }
+
+interface ByRoleOption {
+  name: string;
```
mdjastrzebski, line 207: The option should be optional

```suggestion
  name?: string;
```

typings/index.flow.js
```diff
@@ -202,6 +202,11 @@ interface UnsafeByPropsQueries {
     | Array<ReactTestInstance>
     | [];
 }
+
+interface ByRoleOption {
```
mdjastrzebski, line 206: Plural `options` in order to match type parameter naming

```suggestion
interface ByRoleOptions {
```

typings/index.flow.js
```diff
@@ -244,16 +249,24 @@ interface A11yAPI {
...
   findByRole: (
     matcher: A11yRole | RegExp,
+    role?: ByRoleOption,
     waitForOptions?: WaitForOptions
   ) => FindReturn;
   findAllByRole: (
     matcher: A11yRole | RegExp,
+    role?: ByRoleOption,
     waitForOptions?: WaitForOptions
   ) => FindAllReturn;
```
mdjastrzebski, line 271:
```suggestion
  getByRole: (matcher: A11yRole | RegExp, options?: ByRoleOptions) => GetReturn;
  getAllByRole: (
    matcher: A11yRole | RegExp,
    options?: ByRoleOptions
  ) => GetAllReturn;
  queryByRole: (matcher: A11yRole | RegExp, options?: ByRoleOptions) => QueryReturn;
  queryAllByRole: (
    matcher: A11yRole | RegExp,
    options?: ByRoleOptions
  ) => QueryAllReturn;
  findByRole: (
    matcher: A11yRole | RegExp,
    options?: ByRoleOptions,
    waitForOptions?: WaitForOptions
  ) => FindReturn;
  findAllByRole: (
    matcher: A11yRole | RegExp,
    options?: ByRoleOptions,
    waitForOptions?: WaitForOptions
  ) => FindAllReturn;
```

src/queries/role.ts
```diff
@@ -11,20 +13,37 @@ import type {
...
 
+type Role = AccessibilityRole | TextMatch;
+
+type ByRoleOptions = {
+  name?: TextMatch;
+};
+
+const filterByAccessibleName = (
+  node: ReactTestInstance,
+  name: TextMatch | undefined
```
mdjastrzebski, line 24:
```suggestion
  name?: TextMatch
```

src/queries/role.ts
```diff
@@ -11,20 +13,37 @@ import type {
...
+) => {
+  if (!name) return true;
+
+  const { queryByText, queryByLabelText } = getQueriesForElement(node);
+  return Boolean(queryByText(name) || queryByLabelText(name));
```
mdjastrzebski, line 29: I think we should use the most permissive query here: `queryAll` as `query` might throw in case it finds two matching child elements:

```suggestion
  const nodeQueries = getQueriesForElement(node);
  return nodeQueries.queryAllByText(name).length > 0 || nodeQueries.queryAllByLabelText(name)).length > 0;
```

src/queries/role.ts
```diff
@@ -11,20 +13,37 @@ import type {
   QueryByQuery,
 } from './makeQueries';
 
+type Role = AccessibilityRole | TextMatch;
+
+type ByRoleOptions = {
+  name?: TextMatch;
+};
+
+const filterByAccessibleName = (
```
mdjastrzebski, line 22: Since the method is not an array filter method but rather a predicate let's tweak the name a bit to reduce confusion:

```suggestion
const matchAccessibleName = (
```

src/queries/role.ts
```diff
@@ -11,20 +13,37 @@ import type {
...
+  node: ReactTestInstance,
+  name: TextMatch | undefined
+) => {
+  if (!name) return true;
```
mdjastrzebski, line 26: Previous check was also triggered by empty string.

```suggestion
  if (name == null) return true;
```

src/queries/__tests__/role.test.tsx
```diff
@@ -77,3 +78,61 @@ test('getAllByRole, queryAllByRole, findAllByRole', async () => {
...
+  test('Find an element that has the corresponding role and a children with the name', () => {
```
mdjastrzebski, line 83: nit,style:
Most of the tests for RNTL are named starting with lowercase and using `it()`-like naming, e.g `test('returns ...')`, `test('waits ...')`, etc. Pls adjust the test naming here

src/queries/__tests__/role.test.tsx
```diff
@@ -77,3 +78,61 @@ test('getAllByRole, queryAllByRole, findAllByRole', async () => {
...
+    const { getByRole } = render(
+      <TouchableOpacity accessibilityRole="button" testID="target-button">
```
mdjastrzebski, line 85: All of the following tests rely on verifying explicitly set `button` role on `TouchableOpacity`. It would be good to add some tests that:
1. Test of implicitly set roles by RN (I think it should do such thing, but I'm not sure it does)
2. Test various other roles beside buttons for more real life coverage.

src/queries/__tests__/role.test.tsx
```diff
@@ -77,3 +78,61 @@ test('getAllByRole, queryAllByRole, findAllByRole', async () => {
...
+        <Text>Save</Text>
```
mdjastrzebski, line 86: duplicate the test, but duplicate the <Text> inside the <TO>, like this:

```ts
<TouchableOpacity accessibilityRole="button" testID="target-button">
  <Text>Save</Text>
  <Text>Save</Text>
</TouchableOpacity>
```

This way we can check that it does not throw on finding two matching `name` instances inside component with riven role

src/queries/role.ts
```diff
@@ -11,20 +13,37 @@ import type {
   QueryByQuery,
 } from './makeQueries';
 
+type Role = AccessibilityRole | TextMatch;
```
mdjastrzebski, line 16: Currently the effective type of `Role` is `TextMatch` === `string | RegExp`, as `AccessibilityRole` role is string literal union and this gets merged into `string` from `TextMatch`. So I think we should remove `AccessibilityRole` reference here, as it does not change anything but can confuse the reader.

Alternatively, we could allow for only `AccessibilityRole` values, but that would be a breaking change and also would go against RTL API which accepts wide type of `string | RegExp`.

mdjastrzebski (changes requested):
Awesome work @AugustinLF!

The code is indeed much simpler after @MattAgn's #977.

I've added some minor suggests and tweaks to address before we merge this.

src/queries/role.ts
AugustinLF: Good catch!

src/queries/__tests__/role.test.tsx
AugustinLF: Woopsy, I'm used to either going `it('should...` or using a fully descriptive name with `test('...`.

src/queries/role.ts
AugustinLF: Yeah that's something I wanted to discus, forgot to raise in the PR. Another solution is to use [this hack](https://github.com/testing-library/dom-testing-library/blob/main/types/matches.d.ts#L11) to get autocomplete on the name, which I think would be a nice feature.

mdjastrzebski: Hmmm, that's an interesting approach. From the thread you linked I conjured this locally:
```ts
type HintString<T extends string> = T | Omit<T, T>;
type Role = HintString<AccessibilityRole> | RegExp;
```

I like the autocompletion but the solution feels a hacky and type information in hover popups is actually harder to understand. All in all, I think we just should go with simple `TextMatch`.

src/queries/__tests__/role.test.tsx
AugustinLF: > Test of implicitly set roles by RN (I think it should do such thing, but I'm not sure it does)

Surprise surprise, it not something that works^^' Event `getByRole('image')`  for `render(<Image />)` doesn't work, you need to set `accessibilityRole`. That's something that we definitely should support, but I'd say it's not tied to this PR (since I don't change the role matching). I created the issue #1130 to track that.

mdjastrzebski: How come RN not implement this by default 🤦🏻‍♂️

mdjastrzebski: So the actual screen reader should know the a11y role of `Image`, but JS is not informed.

I wonder about things like `TouchableOpacity` which afaik are implemented as composite components that renders host `View`, so they probably should have a `button` role set in JS, because it's the JS part which knows that `TouchableOpacity`'s host `View` is meant to be a button.

AugustinLF: > So the actual screen reader should know the a11y role of `Image`, but JS is not informed.

Yes exactly. So for `TouchableOpacity`, I'm not sure what a screen-reader would read (`button` might be the default, or not), but that's what we should rely upon for our implementation.

src/queries/role.ts
```diff
@@ -11,14 +12,28 @@ import type {
   QueryByQuery,
 } from './makeQueries';
 
+type ByRoleOptions = {
+  name?: TextMatch;
+};
+
+const matchAccessibleName = (node: ReactTestInstance, name?: TextMatch) => {
```
mdjastrzebski, line 19:
```suggestion
const matchAccessibleNameIfNeeded = (node: ReactTestInstance, name?: TextMatch) => {
```

mdjastrzebski (changes requested):
Requested one small minor readability change.

AugustinLF:
@mdjastrzebski we should be good!

mdjastrzebski (approved):
Great job @AugustinLF 🚀

mdjastrzebski:

PR #1128, opened 2022-09-19
feat: `isInaccessible` API form RTL/DTL

mdjastrzebski:
### Summary

Implement [`isInaccessible`](https://testing-library.com/docs/dom-testing-library/api-accessibility/#isinaccessible) API from DTL (re-exported from RTL).

This should be a useful building block for features as:
* #1064
* [feat: add toBeVisible matcher #73 from Jest Native](https://github.com/testing-library/jest-native/pull/73)

### To Do
- [x] Implement & export basic API
- [x] Add style-based checks (`display: none` & `opacity: 0`) + tests
- [x] Add iOS `accessibilityElementsHidden` with tests
- [x] Add Android `importantForAccessibility` with tests
- [x] Add iOS: `accessibilityViewIsModal` with tests
- [x] Write docs

### Test plan

All checks should pass.

### DIFF
--- a/src/helpers/accessiblity.ts
+++ b/src/helpers/accessiblity.ts
@@ -7,10 +7,6 @@ export function isInaccessible(element: ReactTestInstance | null): boolean {
     return true;
   }
 
-  // Android: importantForAccessibility
-  // See: https://reactnative.dev/docs/accessibility#importantforaccessibility-android
-  if (element.props.importantForAccessibility === 'no') return true;
-
   let current: ReactTestInstance | null = element;
   while (current) {
     if (isSubtreeInaccessible(current)) {
--- a/src/helpers/accessiblity.ts
+++ b/src/helpers/accessiblity.ts
@@ -44,7 +44,7 @@ function isSubtreeInaccessible(
   if (instance.props.importantForAccessibility === 'no-hide-descendants')
     return true;
 
-  const flatStyle = StyleSheet.flatten(instance.props.style);
+  const flatStyle = StyleSheet.flatten(instance.props.style) ?? {};
   if (flatStyle.display === 'none') return true;
   if (flatStyle.opacity === 0) return true;
 
--- a/src/helpers/accessiblity.ts
+++ b/src/helpers/accessiblity.ts
@@ -2,16 +2,16 @@ import { StyleSheet } from 'react-native';
 import { ReactTestInstance } from 'react-test-renderer';
 import { getHostSiblings } from './component-tree';
 
-export function isInaccessible(node: ReactTestInstance | null): boolean {
-  if (node == null) {
+export function isInaccessible(element: ReactTestInstance | null): boolean {
+  if (element == null) {
     return true;
   }
 
   // Android: importantForAccessibility
   // See: https://reactnative.dev/docs/accessibility#importantforaccessibility-android
-  if (node.props.importantForAccessibility === 'no') return true;
+  if (element.props.importantForAccessibility === 'no') return true;
 
-  let current: ReactTestInstance | null = node;
+  let current: ReactTestInstance | null = element;
   while (current) {
     if (isSubtreeInaccessible(current)) {
       return true;
@@ -23,30 +23,30 @@ export function isInaccessible(node: ReactTestInstance | null): boolean {
   return false;
 }
 
-function isSubtreeInaccessible(node: ReactTestInstance | null): boolean {
-  if (node == null) {
+function isSubtreeInaccessible(element: ReactTestInstance | null): boolean {
+  if (element == null) {
     return true;
   }
 
   // iOS: accessibilityElementsHidden
   // See: https://reactnative.dev/docs/accessibility#accessibilityelementshidden-ios
-  if (node.props.accessibilityElementsHidden) {
+  if (element.props.accessibilityElementsHidden) {
     return true;
   }
 
   // Android: importantForAccessibility
   // See: https://reactnative.dev/docs/accessibility#importantforaccessibility-android
-  if (node.props.importantForAccessibility === 'no-hide-descendants') {
+  if (element.props.importantForAccessibility === 'no-hide-descendants') {
     return true;
   }
 
   // Note that `opacity: 0` is not threated as inassessible on iOS
-  const flatStyle = StyleSheet.flatten(node.props.style) ?? {};
+  const flatStyle = StyleSheet.flatten(element.props.style) ?? {};
   if (flatStyle.display === 'none') return true;
 
   // iOS: accessibilityViewIsModal
   // See: https://reactnative.dev/docs/accessibility#accessibilityviewismodal-ios
-  const hostSiblings = getHostSiblings(node);
+  const hostSiblings = getHostSiblings(element);
   if (hostSiblings.some((sibling) => sibling.props.accessibilityViewIsModal)) {
     return true;
   }
--- a/src/helpers/accessiblity.ts
+++ b/src/helpers/accessiblity.ts
@@ -1,8 +1,9 @@
 import { StyleSheet } from 'react-native';
 import { ReactTestInstance } from 'react-test-renderer';
+import { getHostChildren } from './component-tree';
 
 export function isInaccessible(instance: ReactTestInstance | null): boolean {
-  if (!instance) {
+  if (instance == null) {
     return true;
   }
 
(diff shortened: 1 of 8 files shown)


src/helpers/accessiblity.ts
```diff
@@ -0,0 +1,52 @@
+import { StyleSheet } from 'react-native';
+import { ReactTestInstance } from 'react-test-renderer';
+
+export function isInaccessible(instance: ReactTestInstance | null): boolean {
+  if (!instance) {
```
thymikee, line 5: explicit checks are micro-faster
```suggestion
  if (instance === null) {
```

src/helpers/accessiblity.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (instance.props.accessibilityElementsHidden) return true;
+
+  // Android: importantForAccessibility
+  // See: https://reactnative.dev/docs/accessibility#importantforaccessibility-android
+  if (instance.props.importantForAccessibility === 'no-hide-descendants')
+    return true;
+
+  const flatStyle = StyleSheet.flatten(instance.props.style) ?? {};
+  if (flatStyle.display === 'none') return true;
+  if (flatStyle.opacity === 0) return true;
```
thymikee, line 49: Are zero-opacity items inaccessible? Would an item with opacity 0 would be available for screen readers?

mdjastrzebski: I had not found any clear statement in docs that supports, or rejects, being read by VoiceOver screen reader. On the web it is reportedly not ignored by a screen reader, so I will need to conduct an actual experiment to check that.

#### DTL comparison

Seems like the DTL implementation does not check for `opacity: 0`, and uses only the following:
* `element.hidden === true` (incl. subtree)
* `element.getAttribute('aria-hidden') === 'true'` (incl. subtree)
* `window.getComputedStyle(element).display === 'none'` (incl. subtree)
* `window.getComputedStyle(element).visibility === 'hidden'` (only directly on element)

See: https://github.com/testing-library/dom-testing-library/blob/0226aea74e873ba96dae414edc533c33b1e51867/src/role-helpers.js#L15

Also `hidden` does not seem to be affected by `opacity` style.

thymikee: Let's remove the opacity check then :)

thymikee: not at the moment. Let's watch this space

website/docs/API.md
```diff
@@ -669,3 +680,23 @@ it('should use context value', () => {
...
+
+```ts
+function isInaccessible(
+  element: ReactTestInstance | null
+): boolean {}
+```
+
+Checks if given element should be excluded from the accessiblity e.g. by screen readers. 
+
+Element is considered inaccessbile when one of the follwing applies:
```
AugustinLF, line 696:
```suggestion
Element is considered inaccessible when one of the following applies:
```

website/docs/API.md
```diff
@@ -669,3 +680,23 @@ it('should use context value', () => {
...
+* element has `display: none` style
```
AugustinLF, line 697: Should we also exclude `accessible={false}`?

src/helpers/component-tree.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  const hostChildren: ReactTestInstance[] = [];
+
+  element.children.forEach((child) => {
+    if (typeof child !== 'object') {
+      return;
+    }
+
+    if (isHostElement(child)) {
```
AugustinLF, line 52: If I understand correctly, if a child host element itself has host children, only this child host element will be returned. Is that on purpose?

AugustinLF (changes requested):

website/docs/API.md
mdjastrzebski: As far as I understand the documentation around [`accessible`](https://reactnative.dev/docs/accessibility#accessible) prop, it is not about marking element as accessible or not, but rather specifying that given element is an "accessibility element": "When a view is an accessibility element, it groups its children into a single selectable component. By default, all touchable elements are accessible."

So setting it to `false` does not mean that element is not accessible, but rather that it's not an accessibility element.

src/helpers/component-tree.ts
mdjastrzebski: Yes, as this method returns host children, i.e. direct host descendants, not all host descendents

website/docs/API.md
AugustinLF: `accessible={false}` implies that a component will not be read by a screen-reader as a touchable if I'm . A common use case is nested touchables, where you don't want the top one to be read.

mdjastrzebski: Your comments make sense, seems like this might be more complex functionality then I thought. I will convert this PR back to draft and read more about accessibility. Thanks for pointing out holes in my implementation.

mdjastrzebski: @AugustinLF I've one my research on a11y and here are my findings:

### Accessibility tree:
1. Accessibility tree is a parallel component tree to host component tree that consists only elements with which assistive technology can interact with (aka accessible objects): https://www.w3.org/TR/wai-aria-1.2/#terms 
2. Accessibility tree does not contain options that are purely presentational (`presentational`/`none` `role`), as they are not helpful to assistive technologies. So our regular `Views` that are not `buttons`, etc., will fit here.
3. Elements can be excluded from accessibility tree for [two reasons](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion):
  a. they are not to be displayed (`display: none`, `hidden` prop, `aria-hidden` prop, etc) - this condition applies to all of their children down the element tree
  b. they are presentational (`role="presentational"|"none"` - btw these are synonyms)

### [`isInaccessible`](https://testing-library.com/docs/dom-testing-library/api-accessibility) API from DTL:
1. It returns `true` only for elements that are not to be displayed (`display: none`, etc) but not for presentational elements. 

> This function will compute if the given element should be excluded from the accessibility API by the browser. It implements every MUST criteria from the [Excluding Elements from the Accessibility Tree](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion) section in WAI-ARIA 1.2 with the exception of checking the role attribute.

The `role` part is exactly about about `presentational/none` role not being excluded. [The function implementation](https://github.com/testing-library/dom-testing-library/blob/a9a8cf26992ff0f6b4257b7300939f461d04440d/src/role-helpers.js#L15) confirms this.

### Conclusions
1. There are three types of elements with respect o accessiblity:
 a. accessible elements - elements with which assistive technology can interact with
 b. presentational elements - elements which server layout/presentation purpose but are ignored by assistive tech
 c. hidden elements - elements which are explicitly excluded from assitive tech

2. DTL `isInaccessible` returns `true` for accessible & presentational elements, but `false` for hidden elements
3. RN's `accessible` would correspond to controlling whether element is accessible or presentational, so should not have an effect on our `isInaccessible` API. Similarly setting `importantForAccessibility` to `no` would only affect assignment to accessible or presentational groups.

website/docs/API.md
```diff
@@ -669,3 +680,23 @@ it('should use context value', () => {
...
+
+### `isInaccessible`
+
+```ts
+function isInaccessible(
+  element: ReactTestInstance | null
+): boolean {}
+```
+
+Checks if given element should be excluded from the accessiblity e.g. by screen readers.
```
thymikee, line 694: "from the accessiblity" what? "from the accessibility checks" maybe?

mdjastrzebski: The text is based on DTLs "This function will compute if the given element should be excluded from the accessibility API by the browser. It implements every MUST criteria from the [Excluding Elements from the Accessibility Tree](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion) section in WAI-ARIA 1.2 with the exception of checking the role attribute."

However the first part of the original text is confusing, as presentational elements are excluded from accessibility API but the function will say that they aren't excluded. I'll update the description to be more accurate.

website/docs/API.md
AugustinLF: Great write up, you're right! I was not familiar enough with what actually `isInaccessible` entails, and what it want to cover. It does make sense when you come back to the root reason, which was detecting "hidden" elements, and in the case, `accessible:false` should indeed return that.

website/docs/API.md
mdjastrzebski: I've updated the doc, and explained the behavior of the functions and it's relation to DTL's counterpart and ARIA in the simpliest terms I could :-)

mdjastrzebski:
@AugustinLF I'd like to include this in the upcoming release, could you review it again?

AugustinLF (approved):
🚢

mdjastrzebski:

PR #1131, opened 2022-09-20
Switch to IS_REACT_ACT_ENVIRONMENT instead of act when needed when using react 18

### LINKED ISSUE #1093: Unexpected failing test when mixing promises and useEffect

AugustinLF:
Closes #1093. Implementation is copied from https://github.com/testing-library/react-testing-library/pull/1031

Without the react 17/18 branches (i.e only using the new API), tests pass but are full of `console.error` related to `act` and `async act`. With his new implementation, everything passes (including the new tests that used to fail on react 18), without calls to `console.error`.


src/__tests__/waitFor.test.tsx
```diff
@@ -78,6 +78,38 @@ test('waits for element with custom interval', async () => {
...
+      if (state) {
+        onPress();
+      }
+    }, [state, onPress]);
+
+    return (
+      <Pressable
+        onPress={async () => {
+          await Promise.resolve();
+          setState(true);
```
mdjastrzebski, line 95: Would it matter if we invoked `onPress` here after `await Promise` instead of in `useEffect`?

src/checkReactVersionAtLeast.ts
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+
+export function checkReactVersionAtLeast(
```
mdjastrzebski, line 3: Can we rename file to `react-version.ts` or something simplier then the current name?

src/index.ts
```diff
@@ -14,4 +15,21 @@ if (typeof afterEach === 'function' && !process.env.RNTL_SKIP_AUTO_CLEANUP) {
   });
 }
 
+if (
+  typeof beforeAll === 'function' &&
+  typeof afterAll === 'function' &&
+  !process.env.RNTL_SKIP_AUTO_CLEANUP
```
mdjastrzebski, line 21: Should we check `RNTL_SKIP_AUTO_CLEANUP` flag? This does not look like it's strictly related to autocleanup feature.

src/waitFor.ts
```diff
@@ -198,6 +191,17 @@ export default async function waitFor<T>(
     return waitForInternal(expectation, optionsWithStackTrace);
   }
 
+  if (checkReactVersionAtLeast(18, 0)) {
```
mdjastrzebski, line 194: nit:style:

Pls put this `if` above the `if (!checkReactVersionAtLeast(16, 9)) {` in order to improve readability

src/act.ts
```diff
@@ -1,7 +1,101 @@
...
+import { checkReactVersionAtLeast } from './checkReactVersionAtLeast';
 
 const actMock = (callback: () => void) => {
   callback();
 };
 
-export default act || actMock;
+type GlobalWithReactActEnvironment = {
+  IS_REACT_ACT_ENVIRONMENT?: boolean;
+} & typeof globalThis;
```
mdjastrzebski, line 10:
```suggestion
} & typeof globalThis;

```

src/act.ts
```diff
@@ -1,7 +1,101 @@
-import { act } from 'react-test-renderer';
+import { act as reactTestRendererAct } from 'react-test-renderer';
+import { checkReactVersionAtLeast } from './checkReactVersionAtLeast';
 
 const actMock = (callback: () => void) => {
   callback();
 };
 
-export default act || actMock;
+type GlobalWithReactActEnvironment = {
```
mdjastrzebski, line 8: Can we use TS `declare` to extend global object? https://bobbyhadz.com/blog/typescript-declare-global-variable#declare-global-variables-in-typescript

src/act.ts
```diff
@@ -1,7 +1,101 @@
...
+  IS_REACT_ACT_ENVIRONMENT?: boolean;
+} & typeof globalThis;
+function getGlobalThis(): GlobalWithReactActEnvironment {
+  // eslint-disable-next-line no-restricted-globals
```
mdjastrzebski, line 12: This code seems to come from RTL, but the RTL starts with following check:

```ts
  /* istanbul ignore else */
  if (typeof globalThis !== 'undefined') {
    return globalThis
  }
```

Why we don't have it?

src/act.ts
```diff
@@ -1,7 +1,101 @@
...
+
+function setIsReactActEnvironment(isReactActEnvironment: boolean | undefined) {
+  getGlobalThis().IS_REACT_ACT_ENVIRONMENT = isReactActEnvironment;
+}
+
+function getIsReactActEnvironment() {
+  return getGlobalThis().IS_REACT_ACT_ENVIRONMENT;
+}
+
+type Act = typeof reactTestRendererAct;
```
mdjastrzebski, line 35:
```suggestion
type Act = typeof reactTestRendererAct;

```

src/act.ts
```diff
@@ -1,7 +1,101 @@
...
+    try {
+      // The return value of `act` is always a thenable.
+      let callbackNeedsToBeAwaited = false;
+      const actResult = actImplementation(() => {
+        const result = callback();
+        if (
+          result !== null &&
+          typeof result === 'object' &&
+          // eslint-disable-next-line promise/prefer-await-to-then
+          typeof (result as any).then === 'function'
```
mdjastrzebski, line 51: Maybe that will work?
Or 
```
// @ts-expect-error this should be a promise or thenable
```
```suggestion
          typeof result['then'] === 'function'
```

src/act.ts
```diff
@@ -1,7 +1,101 @@
...
+    } catch (error) {
+      // Can't be a `finally {}` block since we don't know if we have to immediately restore IS_REACT_ACT_ENVIRONMENT
+      // or if we have to await the callback first.
+      setIsReactActEnvironment(previousActEnvironment);
+      throw error;
+    }
+  };
+}
+
+const act = reactTestRendererAct
```
mdjastrzebski, line 91: Ah the nested ternary. Pls replace by simple js function
```
function getAct() {
  if (!reactTestRendererAct) { 
    return actMock;
  }

  return checkReactVersionAtLeast(18, 0)
    ? withGlobalActEnvironment(reactTestRendererAct)
    : return reactTestRendererAct;
}
```

mdjastrzebski (changes requested):
@AugustinLF big WOW for creating this PR 🚀 

I've done some cursory review and left few comments, I plan to dig a bit deeper into that, as the PR is quite complex.

src/act.ts
thymikee: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis it's available since Node v12.0.0, so we don't need to check any of these and use `globalThis` directly

src/index.ts
AugustinLF: If you don't do that, you have nested `beforeAll` in the test that tests for cleanup^^' and that breaks it. Current implementation is the same as in RTL. And alternative would be to break out the `index.ts` in something that re-exports pure, + two files once for the clean up, once for the ACT thingy, and in the test file, only import the other one. Or add another custom global just to ignore it in the tests.

src/act.ts
AugustinLF: > Why we don't have it?

It got added in a follow up PR so I missed it.

But as @thymikee said, but given the current versions of node we support, we can switch to globalThis.

src/act.ts
AugustinLF: First one doesn't work, but expecting the error is indeed good

src/act.ts
```diff
@@ -1,7 +1,101 @@
...
+          callbackNeedsToBeAwaited = true;
+        }
+        return result;
+      });
+      if (callbackNeedsToBeAwaited) {
+        const thenable = actResult;
+        return {
+          then: (
+            resolve: (value: never) => never,
+            reject: (value: never) => never
```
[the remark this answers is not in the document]
AugustinLF: Nope, that's not the types of `act`. You'll end up with
```sh
(parameter) returnValue: never
Argument of type '(returnValue: never) => void' is not assignable to parameter of type '(value: never) => never'.
  Type 'void' is not assignable to type 'never'.
```

AugustinLF:
@mdjastrzebski I've updated the PR with your feedback, let me know if you need anything else :)

src/__tests__/waitFor.test.tsx
```diff
@@ -78,6 +78,38 @@ test('waits for element with custom interval', async () => {
   expect(mockFn).toHaveBeenCalledTimes(2);
 });
 
+test('waits for async event with fireEvent', async () => {
+  const Comp = ({ onPress }: { onPress: () => void }) => {
```
mdjastrzebski, line 82: readability: Can we move the `Comp` outside of `test` block?

src/__tests__/waitFor.test.tsx
mdjastrzebski: Ok, I understand. Pls add some comments that describes this, as without that knowledge the code looks very unnatural and badly written event handling

src/index.ts
mdjastrzebski: Can we structure it more like RTL, so that it's easier to sync code in the future:

```
if (typeof process === 'undefined' || !process.env?.RTL_SKIP_AUTO_CLEANUP) {
  // If we're running in a test runner that supports afterEach
  // then we'll automatically run cleanup afterEach test
  // this ensures that tests run in isolation from each other
  // if you don't like this then either import the `pure` module
  // or set the RNTL_SKIP_AUTO_CLEANUP env variable to 'true'.
  if (typeof afterEach === 'function') {
    // eslint-disable-next-line no-undef
    afterEach(async () => {
      await flushMicroTasks();
      cleanup();
    });
  }

  if (typeof beforeAll === 'function' && typeof afterAll === 'function') {
    // This matches the behavior of React < 18.
    let previousIsReactActEnvironment = getIsReactActEnvironment();
    beforeAll(() => {
      previousIsReactActEnvironment = getIsReactActEnvironment();
      setReactActEnvironment(true);
    });
  
    afterAll(() => {
      setReactActEnvironment(previousIsReactActEnvironment);
    });
  }
}
```

mdjastrzebski (changes requested):
Almost there 🚀 , looking forward to merging that :-)

src/__tests__/waitFor.test.tsx
AugustinLF: Sure, I prefer colocation and I find it easier to read (since you can fold the `test()` block), but since we're doing that, I guess consistency is also worth!

AugustinLF:
We're god @mdjastrzebski :)

src/__tests__/waitFor.test.tsx
```diff
@@ -78,6 +78,40 @@ test('waits for element with custom interval', async () => {
...
+    <Pressable
+      onPress={async () => {
+        await Promise.resolve();
+        setState(true);
+      }}
+    >
+      <Text>Trigger</Text>
+    </Pressable>
+  );
+};
```
mdjastrzebski, line 103:
```suggestion
};

```

mdjastrzebski:
@AugustinLF It seems I've broken build by pushing a simple commit that adds one line for formatting. Can you add some minimal commit so that the build is triggered as your user?

mdjastrzebski (approved):
Looks great! Though I've broke the build by adding an empty line for formatting purpose.

AugustinLF:
@mdjastrzebski fixed, added an empty commit!

thymikee:
I wish React team documented this better. This horrific experience for testing library authors.

AugustinLF:
> I wish React team documented this better. This horrific experience for testing library authors.

Yeah, the documentation was there in the WG, but not in the main release notes, so a bit sketchy.

@mdjastrzebski what do you think of creating an RC release of that? I'm currently working on our react 18 upgrade and that'd be some good battle testing.

mdjastrzebski:
Going through RC sounds good, for extra safety. 

From our analysis of the code from both RTL and React 18, we came to conclusion that `IS_REACT_ACT_ENVIRONMENT` option affects only logging of sync `act` error messages.

In short:
`IS_REACT_ACT_ENVIRONMENT  === false` 
  => [isReactActEnvironmentGlobal === false`](https://github.com/facebook/react/blob/346c7d4c43a0717302d446da9e7423a8e28d8996/packages/react-reconciler/src/ReactFiberAct.new.js#L25) 
  => [!isLegacyActEnvironment](https://github.com/facebook/react/blob/346c7d4c43a0717302d446da9e7423a8e28d8996/packages/react-reconciler/src/ReactFiberAct.new.js#L34) 
  => [disable sync act warning in React](https://github.com/facebook/react/blob/112d0498c8e05100f0a1fa0c4c0cc11534ae5353/packages/react-reconciler/src/ReactFiberWorkLoop.new.js#L3576)

AugustinLF:
Yes. But on top of that we updated the implementation of waitFor:
```
// from
let result: T;
await act(async () => {
  result = await waitForInternal(expectation, optionsWithStackTrace);
});

// to
await waitForInternal(expectation, optionsWithStackTrace);
```

mdjastrzebski:
@AugustinLF we will make the release after the weekend, as I need @thymikee for npm publish access rights.

mdjastrzebski:

Issue #1130, opened 2022-09-20
byRole queries don't find element that don't have an explicitly set accessibilityRole prop


AugustinLF:
## Describe the bug

`react-native` built-in elements are not queriable with `byRole` queries if you don't set the `accessibilityRole` prop.

## Expected behaviour

`byRole` queries should be able to match some elements without `accessibilityRole`.

We however need to be conscious of what roles should be used to match what elements. My suggestion is checking each element with a voice-over and see what matches:

- should `<Pressable />` be `button` by default?
- `<TextInput />` will read `Input` on ios voiceover, but that is not an  `AccessibilityRole` as exported by `react-native`.

We might need to think about types. The point raised in https://github.com/callstack/react-native-testing-library/pull/1127#discussion_r974160967 that suggests using `TextMatch` for the role will make it harder to discover those cases.

## Steps to Reproduce

```shell
    Unable to find an element with accessibilityRole: image

      140 |     const { getByRole } = render(<Image />);
      141 |
    > 142 |     getByRole('image');
          |     ^
      143 |   });
      144 | });
      145 |
```
## Versions

Latest.

mdjastrzebski:
This is a questions whether we want to emulate that behavior or should React Native actually set these roles by itself.

I wonder why React Native is not supporting more of [ARIA roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)? It even misses things like `textbox` or equivalent role, which seems a huge omission, that will somehow limit the usefulness of `byRole + name`  query.

mdjastrzebski:
Regarding getting implicit roles there is another relevant public API: [`getRoles()`](https://testing-library.com/docs/dom-testing-library/api-accessibility#getroles) which we should take into account. 

Looks like it's responsible for calculating both explicit and implicit roles for given element: https://github.com/testing-library/dom-testing-library/blob/a9a8cf26992ff0f6b4257b7300939f461d04440d/src/role-helpers.js#L155

mdjastrzebski:
I think we also need to think whether to support some ARIA roles that are important but not present in RN [`AccessibilityRole`](https://reactnative.dev/docs/accessibility#accessibilityrole) definition, especially: `textbox`, maybe something more.

mdjastrzebski:
Do we know which implicit roles we want to support for which components?

Initially we could start with:

* `text` for `Text`
[...]

PR #1139, opened 2022-09-26
fix: *ByRole queries to match their own text text content

### LINKED ISSUE #1140: ByRole queries don't find node with name when the node with role has the name as a direct string child
## Describe the bug

`getByRole(role, {name})` doesn't work when the target node doesn't have components as a child.

```tsx
test('returns an element when the direct child is text', () => {
    const { getByRole } = render(
      <Text accessibilityRole="header" testID="target-header">
        About
      </Text>
    );

    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
    expect(getByRole('header', { name: 'About' }).props.testID).toBe(
      'target-header'
    );
  });
```

Issue open for better tracking, see #1139 for discussions.

## Versions
RNTL 11.2

AugustinLF:
### Summary

The following test case doesn't work with our new `getByRole(role, {name})` API,

```tsx
 test('returns an element when the direct child is text', () => {
    const { getByRole } = render(
      <Text accessibilityRole="header" testID="target-header">
        About
      </Text>
    );

    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
    expect(getByRole('header', { name: 'About' }).props.testID).toBe(
      'target-header'
    );
  });
```

Would probably warrant a patch to `11.2`.

Closes #1140


src/queries/text.ts
```diff
@@ -60,21 +54,18 @@ const getNodeByText = (
...
-    const { Text } = require('react-native');
-    const isTextComponent = filterNodeByType(node, Text);
-    if (isTextComponent) {
-      const textChildren = getChildrenAsText(node.props.children, Text);
-      if (textChildren) {
-        const textToTest = textChildren.join('');
-        const { exact, normalizer } = options;
-        return matches(text, textToTest, normalizer, exact);
-      }
+  const isTextComponent = filterNodeByType(node, 'Text');
```
AugustinLF, line 57: The trick is to work with host components, instead of composite. Running `instance.findAll()` on a component without component children means that all passed instances to the `ByText` functions will be host components, meaning we won't even search the children.

The test suite passes. This shouldn't be a problem, since we filter by type in the different queries.

thymikee: The reason we picked `const { Text } = require('react-native');` was to make sure we pick the correct component in case the internal displayName changes. I think it used to be "RNText" or "forwardRef(Text)" at some point

thymikee: Another reason, which is not valid anymore, was to support different Text node implementations. So, I'm fine with having this as "Text" as long as we have some kind of guarantee this won't change too often in the framework due to adding refs, memo or whatnot

AugustinLF: I'd assume RNTL's own test suite would break with any new version of RN which'd change that. But if we want to add it as a runtime check (for instance if someone uses a version that we don't explicitly test), we could still add a check somewhere to make sure that the matching "string name" for `<Text />` is `Text`.

mdjastrzebski:
If I understand correctly, the root cause of the issue is that when inside `queryAllByRole` we process only host nodes, because part of your filter is `typeof node.type === 'string'`. However `getByText` is looking for composite `Text` exported from RN, which is actually a parent of our host `Text`.

This problem is not a general `*ByText` problem, as in regular cases there always are both composite and host `Text` pair. But in our case we start with queries bound to the host `Text`.

As @thymikee matching by `instance.type === 'Text'` is fragile, as different versions of RN used different host component name. This seems fair as the exact name of host element is an implementation detail, and we want to avoid relying on these if possible.

Perhaps a more robust workaround, but still a workaround, would be to pass `node.parent` to `queryAllByText` in `matchAccessibleNameIfNeeded`. Like this:
```ts
const matchAccessibleNameIfNeeded = (
  node: ReactTestInstance,
  name?: TextMatch
) => {
  if (name == null) return true;

  // Note: due to queryAllByText matching composite `Text` element, we have to start at parent level
  const nodeParentQueries = getQueriesForElement(node.parent);
  const matchesText = queryAllByText(name).length > 0;

  const nodeQueries = getQueriesForElement(node);
  const matchesLabel = queryAllByLabelText(name).length > 0;
 
  return matchesText || matchesLabel;
};
```

This is also a bit fragile as it assumes that composite `Text` exported from RN will be direct parent of host `Text`. 

@thymikee, @AugustinLF  wdyt?

AugustinLF:
The proposition to switch to our own internal model that has been discussed a few times already (see #967) would fix this host vs composite component, right? If that's the case, should we then consider a "hack" to solve the problem at hand, while we look into a more long term solution as part of the rewrite?

Tested your implementation @mdjastrzebski which works (with a slight fix to make it work with TS)
```tsx
const matchAccessibleNameIfNeeded = (
  node: ReactTestInstance,
  name?: TextMatch
) => {
  if (name == null) return true;

  // Note: due to queryAllByText matching composite `Text` element, we have to start at parent level
  const nodeParentQueries = getQueriesForElement(node.parent ?? node);
  const matchesText = nodeParentQueries.queryAllByText(name).length > 0;

  const nodeQueries = getQueriesForElement(node);
  const matchesLabel = nodeQueries.queryAllByLabelText(name).length > 0;

  return matchesText || matchesLabel;
};
```

We could also consider climbing up the parent tree util we find a host component, which would make it less brittle for future changes.

As discussed this is not ideal because it might pose a problem with future RN versions, but given that we now have support for running our test suite against multiple versions, if this ends up changing, it should be fairly easy for us to make sure RNTL works with all the versions we support.

mdjastrzebski:
@AugustinLF I think we could tweak the solution a bit, by climbing up, but only through composite components to the exported `Text` component, since we know that we are doing that for `*ByText` query. That would make our workaround a bit more robust as now `Text` could render some intermediate composite components between exported component and final host component.

I think the proper way to add this, is to make `queryAllByText` have a special check, that would allow it to work when bound to host component instead of adding this to `queryAllByRole`. So essentially we want following test to pass:

```ts
const view = render(<Text testID="subject">Hello</Text>)`
const hostTextQueries = getQueriesForElement(view.getByTestId("subject"));
expect(hostTextQueries.getByText("Hello")).toBeTruthy();
```

We could build some helper to check for being host `Text`:
```ts
function isHostTextElement(element: ReactTestInstance) {
  // Not a host element
  if (typeof element.type !== "string") return false;
  
  let current = element.parent;
  while(typeof current.type !== 'string') {
    if (current.type === RN.Text) { 
      return true;
    }
    current = current.parent;
  }
  
  return false;
}
```

We should be wary not to climb too high the component tree, and then make down queries, as otherwise we might actually catch `Text` in the siblings on component with `role` which would be an implementation error. We might want to add some test to cover for that case.

src/queries/text.ts
```diff
@@ -58,37 +59,65 @@ const getChildrenAsText = (
...
   options?: TextMatchOptions
 ) => Array<ReactTestInstance>) =>
   function queryAllByTextFn(text, options) {
-    const results = instance.findAll((node) =>
-      getNodeByText(node, text, options)
-    );
+    try {
+      const { Text } = require('react-native');
+      const rootInstance = isHostElement(instance)
+        ? getCompositeParent(instance, Text) ?? instance
```
AugustinLF, line 109: The issue here is that we'll get into this line of the ternary also for non `Text` host components. For instance, when searching `getByRole('button', { name })`, it'll try to find a Text higher in the tree and fail. So we need to default to `?? instance`.

A better (more performant) solution would be to do `const rootInstance = filterNodeByType(instance, 'Text')`, so we'll hit this path only when needed. But we fall back to the hardcoded host string `Text`.

src/queries/text.ts
```diff
@@ -58,37 +59,65 @@ const getChildrenAsText = (
...
+  if (!isHostElement(element)) return null;
+
+  let current = element.parent;
+  while (!isHostElement(current)) {
+    // We're at the top of the tree
+    if (!current) {
+      return null;
+    }
+
+    if (filterNodeByType(current, compositeType)) {
```
mdjastrzebski, line 90: Should be simplier
```suggestion
    if (current.type === compositeType) {
```

src/queries/text.ts
mdjastrzebski: What if instead of `isHostElement` you used `isHostTextInstance`?:

```ts
function isHostTextElement(element: ReactTestInstance) {
  // Not a host element
  if (typeof element.type !== "string") return false;
  
  let current = element.parent;
  while(typeof current.type !== 'string') {
    if (current.type === RN.Text) { 
      return true;
    }
    current = current.parent;
  }
  
  return false;
}
```

src/queries/text.ts
```diff
@@ -58,37 +59,65 @@ const getChildrenAsText = (
...
-  } catch (error) {
-    throw createLibraryNotSupportedError(error);
   }
+  return false;
 };
 
+function getCompositeParent(
+  element: ReactTestInstance,
+  compositeType: React.ComponentType
+) {
```
mdjastrzebski, line 80:
```suggestion
function getCompositeParentOfType(
  element: ReactTestInstance,
  type: React.ComponentType
) {
```

src/queries/text.ts
```diff
@@ -58,37 +59,65 @@ const getChildrenAsText = (
...
+  if (!isHostElement(element)) return null;
```
mdjastrzebski, line 81: IMO we can get rid of this check, as it's not needed in the code and also it's hard to guess such check exists based on function name.
```suggestion
  if (!isHostElement(element)) return null;
```

src/queries/text.ts
```diff
@@ -58,37 +59,65 @@ const getChildrenAsText = (
...
+      return matches(text, textToTest, normalizer, exact);
     }
-    return false;
-  } catch (error) {
-    throw createLibraryNotSupportedError(error);
   }
+  return false;
 };
 
+function getCompositeParent(
```
mdjastrzebski, line 77: Pls move this function to `component-tree.ts`, seem like it will fit other tree walking functions

src/__tests__/within.test.tsx
```diff
@@ -94,3 +94,11 @@ test('within() exposes a11y queries', async () => {
 test('getQueriesForElement is alias to within', () => {
   expect(getQueriesForElement).toBe(within);
 });
+
+test('within allows searching for text within a composite component', () => {
+  const view = render(<Text testID="subject">Hello</Text>);
+  // view.getByTestId('subject') returns a composite component, contrary to most queries returning host component
```
mdjastrzebski, line 100: This comment is not accurate s `getByTestId` returns host component only

src/queries/text.ts
AugustinLF: Would it be possible for a host instance that is not `Text` to have a composite parent that is `Text`? Because if that's the case we'd want to avoid going higher up the tree (and take the risk of going to a sibling).

mdjastrzebski: Assuming that when traversing up in the component tree there are intermittent host ancestors along the way, only composite ones, then I think we're pretty safe. 

If non-`Text` host instance would have composite ancestor of exported `Text` type (assuming no host ancestors along the way), then it would mean that exported `Text` renders not only host `Text` but also some other host component.

It seems that the above check (`isHostTextElement`) is atm the most reliable method for checking that given host component is a host `Text` that we have.

@thymikee maybe you have some thoughts on this...

mdjastrzebski: BTW it seems that this whole problem exists only because of the way we collect component text content. If we would have https://github.com/facebook/react/pull/25329 merged, then we would have a better alternative for getting element text content.

thymikee: I'd appreciate adding a heavy suite of tests with lots of configurations of Text, and non-Text components with all possible configurations and going with @mdjastrzebski solution for now. We'll be able to move to a simple string check once the name of the component is a public API of RN. Which is something open for a discussion in the https://github.com/react-native-community/discussions-and-proposals repo. Anybody would like to create a proposal for at least Text and View host components? :)

src/__tests__/within.test.tsx
AugustinLF: Mixed it up^^'

src/queries/text.ts
thymikee: https://github.com/facebook/react/pull/25329 would be for React 18 only though

mdjastrzebski: @thymikee I've submitted PR for such tests https://github.com/callstack/react-native-testing-library/pull/1143

AugustinLF: Just repushed with a carpet bombing approach on tests. Let me know if you feel I should keep/remove them (didn't see your competing PR)

src/helpers/__tests__/component-tree.test.tsx
```diff
@@ -200,3 +202,16 @@ test('returns host siblings for composite component', () => {
...
 });
+
+test('getCompositeParentOfType', () => {
+  const view = render(<View testID="test"></View>);
+  const hostComponent = view.getByTestId('test');
+
+  const compositeComponent = getCompositeParentOfType(hostComponent, View);
+
+  // We get the corresponding composite component (same testID), but not the host
+  expect(hostComponent.props.testID).toBe(compositeComponent?.props.testID);
```
mdjastrzebski, line 213:
```suggestion
  expect(compositeComponent?.type).toBe(View);
  expect(compositeComponent?.props.testID).toBe(hostComponent.props.testID);
```

src/helpers/__tests__/component-tree.test.tsx
```diff
@@ -200,3 +202,16 @@ test('returns host siblings for composite component', () => {
...
+
+  expect(hostComponent).not.toBe(compositeComponent);
```
mdjastrzebski, line 215:
```suggestion
  expect(compositeComponent).not.toBe(hostComponent);
```

src/helpers/component-tree.ts
```diff
@@ -87,3 +87,24 @@ export function getHostSiblings(
...
+  while (!isHostElement(current)) {
+    // We're at the root of the tree
+    if (!current) {
+      return null;
+    }
+
+    if (current.type === type) {
+      return current;
+    }
+    current = current.parent ?? null;
```
mdjastrzebski, line 106: Is that null coalescing necessary? According to @types/react then type of parent is `ReactTestInstance | null`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2b7c710f4890784cd537c82bdcd2c03e98ec09db/types/react-test-renderer/index.d.ts#L31

src/queries/__tests__/role.test.tsx
```diff
@@ -134,4 +134,43 @@ describe('supports name option', () => {
...
+    const { getByRole } = render(
+      <Text accessibilityRole="header" testID="target-header">
+        About
+      </Text>
+    );
+
+    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
+    expect(getByRole('header', { name: 'About' }).props.testID).toBe(
+      'target-header'
+    );
```
mdjastrzebski, line 148: Using `getByTestId` to fetch the host element seems to be a bit stronger assumption on getRoleBehavior. But it makes assumption about `getByTestId` being correctly implemented.
```suggestion
    expect(getByRole('header', { name: 'About' })).toBe(getByTestId('target-header'));
    expect(getByRole('header', { name: 'About' }).props.testID).toBe(
      'target-header'
    );
```

src/queries/__tests__/role.test.tsx
```diff
@@ -134,4 +134,43 @@ describe('supports name option', () => {
...
+
+  test('returns an element with nested Text as children', () => {
+    const { getByRole } = render(
+      <Text accessibilityRole="header" testID="parent">
+        <Text testID="child">About</Text>
+      </Text>
+    );
+
+    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
+    expect(getByRole('header', { name: 'About' }).props.testID).toBe('parent');
```
mdjastrzebski, line 159:
```suggestion
    expect(getByRole('header', { name: 'About' })).toBe(getByTestId('parent'));
    expect(getByRole('header', { name: 'About' }).props.testID).toBe('parent');
```

src/queries/__tests__/role.test.tsx
```diff
@@ -134,4 +134,43 @@ describe('supports name option', () => {
...
+    const { getByRole } = render(
+      <Text
+        accessibilityRole="header"
+        testID="target-header"
+        accessibilityLabel="About"
+      ></Text>
+    );
+
+    // assert on the testId to be sure that the returned element is the one with the accessibilityRole
+    expect(getByRole('header', { name: 'About' }).props.testID).toBe(
```
mdjastrzebski, line 172:
```suggestion
   expect(getByRole('header', { name: 'About' })).toBe(getByTestId('target-header'));
    expect(getByRole('header', { name: 'About' }).props.testID).toBe(
```

src/queries/text.ts
```diff
@@ -58,37 +62,66 @@ const getChildrenAsText = (
...
+        ? getCompositeParentOfType(instance, Text)
+        : instance;
 
-    return results;
+      if (!baseInstance) return [];
+      const results = baseInstance.findAll((node) =>
+        getNodeByText(node, text, Text, options)
+      );
+
+      return results;
```
mdjastrzebski, line 121: nit:
```suggestion
      if (!baseInstance) {
        return [];
      };
      
      return baseInstance.findAll((node) =>
        getNodeByText(node, text, Text, options)
      );
```

src/queries/text.ts
```diff
@@ -58,37 +62,66 @@ const getChildrenAsText = (
...
+      return matches(text, textToTest, normalizer, exact);
     }
-    return false;
-  } catch (error) {
-    throw createLibraryNotSupportedError(error);
   }
+  return false;
 };
 
+function isHostTextElement(
```
mdjastrzebski, line 80: Let's move that to `component-tree.ts` or `react-native-api.ts` (not existis yet but we have `react-native-api.test.tsx`)

mdjastrzebski: Looking at the body of the fn this could be called `isHostElementForType`, as the current function can be called with like `isHostElementForType(element, View)` and still makes sense and is useful.

BTW it seems that we could reuse `getCompositeParentOfType` for implementation:
`isHostElementForType(element, Component) {
    return getCompositeParentOfType(element, Component) != null;
 }

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Our queries interact differently with composite and host elements, and some specific cases require us
+ * to crawl up the tree to a Text composite element to be able to traverse it down again. Going up the tree
+ * is a dangerous behaviour because we could take the risk of then traversing a sibling node to the original one.
+ * This test suite is designed to be able to test as many different combinations, as a safety net.
+ * Specific cases should still be tested within the relevant file (for instance an edge case with `within` should have
+ * an explicit test in the within test suite)
+ */
+describe('nested text handling', () => {
```
mdjastrzebski, line 13: I see couple of points similar across tests in this files:
1. All `getByRole` tests seem to be more relevant as a check for `*ByRole` queries then text handling, Consider moving them to `role.test.tsx`
2. Ditto `*ByText` queries look like a good candidate to be moved to `text.test.tsx`
3. Using assertions like `expect(view.getByText('Save').props.testID).toBe('text');` seem to be a weaker version of using `expect(view.getByRole('button', { name: 'Save' }).props.testID).toBe(getByTestId('pressable'));` assertion, as the latter would pinpoint exactly to host element. We could use both for the `getByRole` queries as they too return host components

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+        <Text testID="text">Save</Text>
+      </View>
+    );
+
+    expect(view.getByTestId('text').props.testID).toBe('text');
+  });
+
+  test('Text within view within text', () => {
+    const view = render(
+      <Text>
```
mdjastrzebski, line 176: Is this legal? I think you can only nest `Text` inside `Text`.

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+        <View>
+          <Text testID="text">Save</Text>
+        </View>
+      </Text>
+    );
+
+    expect(view.getByTestId('text').props.testID).toBe('text');
```
mdjastrzebski, line 183: Otherwise the test seems to be testing `getByTestId` itself.

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    );
+
+    expect(view.getByRole('button', { name: 'Save' }).props.testID).toBe(
+      'pressable'
+    );
+  });
+
+  test('pressable within View within Text, with text child within view', () => {
+    const view = render(
+      <Text>
```
mdjastrzebski, line 116: Nesting non-Text inside `Text` seems to be illegal in RN.

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    );
+
+    expect(view.getByRole('button', { name: 'Save' }).props.testID).toBe(
+      'pressable'
+    );
+  });
+
+  test('pressable within text, with text child', () => {
+    const view = render(
+      <Text>
```
mdjastrzebski, line 57: Nesting non-Text inside Text seems to be illegal in RN.

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,185 @@
...
+    );
+
+    expect(view.getByRole('header', { name: 'About' }).props.testID).toBe(
+      'child'
+    );
+  });
+
+  test('pressable within text with label', () => {
+    const view = render(
+      <Text>
```
mdjastrzebski, line 41: Nesting non-Text inside Text seems to be illegal in RN.

mdjastrzebski (changes requested):
Good work @AugustinLF That PR looks pretty solid. I've found some minor things to address but look forward to having fully functional `getByRole`!

src/__tests__/host-text-nesting.test.tsx
AugustinLF: My take on this is most of those behaviours are already tested in the different test files (`text.test`, `role.test.tsx`, etc.), but usually without the extra parent/child intermediate components.

So this file is more about adding an additional safety nest, is we care about grouping things, this file doesn't make sense anymore. I'm fine with this, but then we would probably want to create a more exhaustive list of all the test cases we want, see what is missing for the different queries, and adding tests for host vs composite elements a bit everywhere.

I feel this would make our test suites a bit hard to grasp, and rough for contributors to join and contribute (most new features shouldn't have to care about host vs composite).

src/__tests__/host-text-nesting.test.tsx
AugustinLF: Yeah I don't think you can, just went through RN docs and didn't see anything about that, but you're right, I'll drop those tests (hopefully it's a behaviour we should try to catch with our own tests at some points).

src/__tests__/host-text-nesting.test.tsx
AugustinLF: Yep this is true, but as raised [here](https://github.com/callstack/react-native-testing-library/pull/1139#discussion_r982614865), most cases should be covered in the relevant queries test suites.

src/helpers/__tests__/component-tree.test.tsx
```diff
@@ -200,3 +202,17 @@ test('returns host siblings for composite component', () => {
     view.getByTestId('siblingAfter'),
   ]);
 });
+
+test('getCompositeParentOfType', () => {
+  const view = render(<View testID="test"></View>);
```
mdjastrzebski, line 207:
```suggestion
  const view = render(<View testID="test" />);
```

src/helpers/__tests__/component-tree.test.tsx
```diff
@@ -200,3 +202,17 @@ test('returns host siblings for composite component', () => {
...
+  const hostComponent = view.getByTestId('test');
+
+  const compositeComponent = getCompositeParentOfType(hostComponent, View);
+
+  // We get the corresponding composite component (same testID), but not the host
+  expect(compositeComponent?.type).toBe(View);
+  expect(compositeComponent?.props.testID).toBe(hostComponent.props.testID);
+
+  expect(compositeComponent).not.toBe(hostComponent);
+  expect(isHostElement(compositeComponent)).toBe(false);
```
mdjastrzebski, line 217: This checks seems redundant (and weaker) when we already have `expect(compositeComponent?.type).toBe(View);`

src/helpers/component-tree.ts
```diff
@@ -87,3 +87,34 @@ export function getHostSiblings(
...
+    if (current.type === type) {
+      return current;
+    }
+    current = current.parent;
+  }
+
+  return null;
+}
+
+export function isHostElementOfType(
```
mdjastrzebski, line 112: Seems like `for` should be more accurate then `of` here:

as in "this view is host element for type `Text` (exported from RN)", 

```suggestion
export function isHostElementForType(
```

src/queries/text.ts
```diff
@@ -84,11 +84,24 @@ const queryAllByText = (
   options?: TextMatchOptions
 ) => Array<ReactTestInstance>) =>
   function queryAllByTextFn(text, options) {
-    const results = instance.findAll((node) =>
-      getNodeByText(node, text, options)
-    );
+    try {
+      const { Text } = require('react-native');
```
mdjastrzebski, line 88: Maybe we could extract that `try/catch` as separate function:
```ts
function getTextComponentFromReactNative() {
  try {
    const { Text } = require('react-native');
    return Text;
  } catch (error) {
     throw createLibraryNotSupportedError(error);
  } 
}
```
good place to put it would be `react-native-api.ts` file

This way we would not have to bloat various places with `try/catch/throw` that make the code blocks more visually cluttered.

AugustinLF: We actually use it only once, the other times are `TextInput`. Should I extract both of them?

mdjastrzebski: I think that soon it will be useful also in different places, e.g. for implementing implicit roles for #1130

AugustinLF: I did extract both. Ideally we should also do it for StyleSheet (which is imported in `isInaccessible`, but that poses some weird types problem, so I'll skip this for now.

src/helpers/react-native-api.ts
```diff
@@ -0,0 +1,19 @@
+import { createLibraryNotSupportedError } from './errors';
+
+export function importTextFromReactNative() {
+  try {
+    const { Text } = require('react-native');
```
mdjastrzebski, line 5: @thymikee do you know why are we importing RN's `Text`, etc using dynamic `require` syntax, instead of regular ESM `import` at the top of the file?

thymikee: This was to throw a descriptive error when react-native is not a dependency, and to open the door for using Text outside of `react-native`. Can be moved to import now

AugustinLF: TBH it's not a bad thing. We have a repo mixing web and native, and we've had people importing RNTL by mistake on web for some dedicated things that doesn't pose a problem (we have runtime check for web vs native). Until 11.2, it was fine, now we have a problem because of the `StyleSheet` import in `'isInaccessible` now breaks just importing it.

But I do understand that it's an edge case that probably doesn't matter for most people.

mdjastrzebski: I think that having React Native as a peer dep is enough. We do not do such dynamic checks for other peer deps like React or React Test Renderer.

If we would really want to have a check like this then we could add it insider `index.ts` where we hook into test runner `before` and `after` hooks or `render` as it's always used. Both of these seem to be a better place than `*ByText` queries.

mdjastrzebski:
@AugustinLF could you refactor the code to use regular `Text`/`TestInput` import from React Native instead of current dynamic one. This should simplify the codebase.

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+        <View>
+          <Text testID="text">Save</Text>
+        </View>
+      </Pressable>
+    );
+
+    expect(view.getByText('Save').props.testID).toBe('text');
+  });
+
+  test('View with text child', () => {
```
mdjastrzebski, line 91: This test verifies `getByTestId` basic case which is probably covered by existing test cases. It's not really related to text nesting. Either remove that or if you feel it's work keeping then it would make sense to move it to `testId.test.tsx`.

src/__tests__/host-text-nesting.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const view = render(
+      <View testID="view">
+        <Text>Save</Text>
+      </View>
+    );
+
+    expect(view.getByTestId('view').props.testID).toBe('view');
+  });
+
+  test('Text within view', () => {
```
mdjastrzebski, line 101: This test verifies `getByTestId` basic case which is probably covered by existing test cases. It's not really related to text nesting. Either remove that or if you feel it's work keeping then it would make sense to move it to `testId.test.tsx`.

src/queries/__tests__/role.test.tsx
```diff
@@ -134,4 +134,50 @@ describe('supports name option', () => {
...
+    expect(getByRole('header', { name: 'About' }).props.testID).toBe('parent');
+  });
+
+  test('returns a header with an accessibilityLabel', () => {
+    const { getByRole, getByTestId } = render(
+      <Text
+        accessibilityRole="header"
+        testID="target-header"
+        accessibilityLabel="About"
+      ></Text>
```
mdjastrzebski, line 172:
```suggestion
      <Text
        accessibilityRole="header"
        testID="target-header"
        accessibilityLabel="About"
      />
```

src/helpers/component-tree.ts
```diff
@@ -87,3 +87,34 @@ export function getHostSiblings(
...
+    if (current.type === type) {
+      return current;
+    }
+    current = current.parent;
+  }
+
+  return null;
+}
+
+export function isHostElementForType(
```
mdjastrzebski, line 112:
```suggestion
/**
 * Note: this function should be generally used for core React Native types like `View`, `Text`, `TextInput`, etc. 
 */
export function isHostElementForType(
```

src/helpers/__tests__/component-tree.test.tsx
```diff
@@ -200,3 +201,14 @@ test('returns host siblings for composite component', () => {
...
+test('getCompositeParentOfType', () => {
+  const view = render(<View testID="test" />);
+  const hostComponent = view.getByTestId('test');
+
+  const compositeComponent = getCompositeParentOfType(hostComponent, View);
+
+  // We get the corresponding composite component (same testID), but not the host
+  expect(compositeComponent?.type).toBe(View);
+  expect(compositeComponent?.props.testID).toBe(hostComponent.props.testID);
+});
```
mdjastrzebski, line 214:
```suggestion
test('getCompositeParentOfType', () => {
  const root = render(<View testID="view"><Text testID="test" /></View>);
  const hostView= root.getByTestId('view');
  const hostText= root.getByTestId('text');
  
  const compositeView = getCompositeParentOfType(hostView, View);
  // We get the corresponding composite component (same testID), but not the host
  expect(compositeView?.type).toBe(View);
  expect(compositeView?.props.testID).toBe("view");

  const compositeText = getCompositeParentOfType(hostText, Text);
  expect(compositeText?.type).toBe(View);
  expect(compositeComponent?.props.testID).toBe("text");

  expect(getCompositeParentOfType(hostText, View)).toBeNull();  
  expect(getCompositeParentOfType(hostView, Text)).toBeNull();
});

test('isHostElementForType', () => {
  const view = render(<View testID="test" />);
  const hostComponent = view.getByTestId('test');
  const compositeComponent = getCompositeParentOfType(hostComponent, View);

  expect(isHostElementForType(hostComponent, View)).toBe(true);
  expect(isHostElementForType(hostComponent, Text)).toBe(false);
  expect(isHostElementForType(compositeComponent, View)).toBe(false);
});
```

mdjastrzebski (commented):
@AugustinLF Looks good! Recommended some final tweaks to the tests.

mdjastrzebski (approved):
Awesome work on these difficult subjects @AugustinLF !

I guess we have all learned a lot about accessibility and host/composite elements tree 😆

PR #1141, opened 2022-09-27
feature: configure API from DTL

mdjastrzebski:
### Summary

Implement `configure` API based on [similar API in DTL](https://testing-library.com/docs/dom-testing-library/api-configuration).

Added initial option for [`asyncWaitTimeout`](https://testing-library.com/docs/dom-testing-library/api-configuration/#asyncutiltimeout).

### Test plan

* Added tests for `config` module functions
* Added tests for `waitFor` consuming `asyncWaitTimeout` option

### DIFF
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -16,9 +16,11 @@ export type {
   RenderResult as RenderAPI,
 } from './render';
 export type { RenderHookOptions, RenderHookResult } from './renderHook';
+export type { Config } from './config';
 
 export { act };
 export { cleanup };
+export { configure, resetToDefault } from './config';
 export { fireEvent };
 export { render };
 export { waitFor };
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -20,7 +20,7 @@ export type { Config } from './config';
 
 export { act };
 export { cleanup };
-export { configure, resetToDefault } from './config';
+export { configure, resetToDefaults } from './config';
 export { fireEvent };
 export { render };
 export { waitFor };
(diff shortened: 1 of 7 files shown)


AugustinLF (approved):

src/pure.ts
```diff
@@ -16,9 +16,11 @@ export type {
   RenderResult as RenderAPI,
 } from './render';
 export type { RenderHookOptions, RenderHookResult } from './renderHook';
+export type { Config } from './config';
 
 export { act };
 export { cleanup };
+export { configure, resetToDefaults } from './config';
```
thymikee, line 23: Is `resetToDefaults` exported from DTL? Couldn't find it on their docs

thymikee (approved):
We could also implement `defaultHidden`, right? Looks good though

src/pure.ts
mdjastrzebski: No it's not, only `configure` is, but I think it's good to have that for testinging convenience. Wdyt?

mdjastrzebski: Re `defaultHidden`: we first need to have https://github.com/callstack/react-native-testing-library/pull/1064 merged as I belive that `defaultHidden` is the default value of `hidden` query option, which we are yet to implement.

mdjastrzebski:
re `defaultHidden`: we first need to merge https://github.com/callstack/react-native-testing-library/pull/1064, as it contains the `hidden` query options, and afaik `defaultHidden` is just the default value of that option.

thymikee:
> re defaultHidden: we first need to merge https://github.com/callstack/react-native-testing-library/pull/1064, as it contains the hidden query options, and afaik defaultHidden is just the default value of that option.

Huh, it's opened for so long I thought it's already in 🙈

mdjastrzebski:
It should be much easier done now when we have `isInaccessible` #1128  , which is the core logic behind `hidden` flag.

thymikee:
Definitely! And please don't get me wrong, I'm not expecting to rush things here :)

Issue #1135, opened 2022-09-23
Debug util that would only print displayed text


pierrezimmermannbam:
## Describe the Feature

When debugging tests on large components, i.e. pages, snapshots can be very long and may not be convenient for debugging purposes. Very often I'm mostly interested in seing what text / testID is displayed on the screen and don't really care about the design so I think that a function that would log only the displayed text and maybe also the testIDs could come in pretty handy. 

## Possible Implementations

We'd have a function to transform the json representation into a string with just text and maybe also testIDs displayed. I implemented such a function on my current project and it looks like this : 

```typescript
export const toText = (node: ReactTestRendererNode | null, textSeparator = '\n'): string | null => {
  if (!node) return '';
  if (typeof node === 'string') return node;
  if (Array.isArray(node)) return toText({ type: 'Fragment', props: {}, children: node });

  const texts = [
    getImageSource(node),
    ...(node.children?.map(child => toText(child, textSeparator)) || []),
  ].filter(truthy);

  return texts.length > 0
    ? texts.join(textSeparator) +
        // Arbitrarily add a separator if more than 1 child, just to add some space in the snap
        (texts.length > 1 ? textSeparator : '')
    : null;
};
```

Then the api would need to be modified to add this functionality. I'm thinking of two ways to do this : 

- Add a debugText function to RenderResult (not sure about the naming yet, this could be something else)
- Add options to the current debug function to enable customization of the output

Second option would allow more customization in the future and we may want to combine several options (for instance text and testID are visible or just text or just testID) but it also makes the api more complex and more difficult to use. Also people may not notice that this feature exists. Alternatively both approach could be combined to have an accessible and easy way of printing just text and also the ability to custom output according to one's needs.

## Related Issues

No related issues as far as I know

@mdjastrzebski  @AugustinLF  what are your thoughts on this ?

AugustinLF:
I agree with the idea of filtering output. For instance, filtering out `style` props would be useful for me since I never use it.

I do think we need customisable output. Because if you look at the preferred query discussion we're having (#1133), role is going to become a priority, so it'd need to be in the output. But if you don't care about it, it could be useful to skip it.

I also think we should allow `debug()` to take an instance (i.e. a subset of the tree). I need quite often to comment out part of the layout that just pollutes my logs...

mdjastrzebski:
@pierrezimmermannbam I think that your use case is closely related to `debug` function, so let's first consider integrating the narrowing down of its input. 

[...]

PR #1160, opened 2022-10-06
feat: debug options

pierrezimmermannbam:
### Summary

Adds optional filterProps option for debug util as discussed in #1135

### Test plan

I added some tests using the option and filtering through propName, propValue or node
Additionally I checked that filtering props in debug doesn't mutate initial instance

### DIFF
--- a/src/helpers/format.ts
+++ b/src/helpers/format.ts
@@ -23,11 +23,12 @@ const getCustomPlugin = (mapProps?: MapPropsFunction): NewPlugin => {
   return {
     test: (val) => plugins.ReactTestComponent.test(val),
     serialize: (val, config, indentation, depth, refs, printer) => {
+      let newVal = val;
       if (mapProps && val.props) {
-        val.props = mapProps(val.props, val);
+        newVal = { ...val, props: mapProps(val.props, val) };
       }
       return plugins.ReactTestComponent.serialize(
-        val,
+        newVal,
         config,
         indentation,
         depth,
--- a/src/render.tsx
+++ b/src/render.tsx
@@ -159,7 +159,6 @@ function debug(
     }
 
     const json = renderer.toJSON();
-
     if (json) {
       return debugDeep(json, debugOptions);
     }


src/__tests__/render.test.tsx
```diff
@@ -191,11 +195,34 @@ test('debug changing component', () => {
...
     'bananaFresh button message should now be "fresh"'
   );
 });
 
+test('debug with filterProps option', () => {
+  jest.spyOn(console, 'log').mockImplementation((x) => x);
+
+  const { debug, getByText } = render(<Banana />);
+
+  debug({ filterProps: (propName) => propName === 'children' });
```
AugustinLF, line 208: I'd either break those in different tests (to have better snapshot tests), and/or inline the snapshots. The snapshot are very "isolated" right now

src/render.tsx
```diff
@@ -135,18 +135,18 @@ function updateWithAct(
 }
 
 interface DebugFunction {
-  (message?: string): void;
+  (options?: string | DebugOptions): void;
```
AugustinLF, line 138: Please update flow types (I wish we could drop them^^')

AugustinLF (changes requested):
Looks pretty good overall (even though I'm not familiar with the way `pretty-format` works, two additional points:
- can we get `mapNode` also as an option?
- can you try adding this option to the global `configure` options? I believe there is a lot of values in setting a value for a whole project (I'll definitely filter out all my `style` props)

src/render.tsx
pierrezimmermannbam: Oh I forgot about these, I updated them

pierrezimmermannbam:
@AugustinLF I added the option to configure the debugOptions globally and treated your other comments. Also I have updated the website docs with the new api. Regarding the mapNode option I was considering adding it in another pr if that's ok with you, so that the review process is shorter and also because I think this option by itself makes sense so the faster it is available the better

package.json
```diff
@@ -97,6 +97,7 @@
     "testTimeout": 60000,
     "transformIgnorePatterns": [
       "/node_modules/(?!(@react-native|react-native)/).*/"
-    ]
+    ],
+    "clearMocks": true
```
mdjastrzebski, line 101: Is this change related to his PR? Why is it needed?

src/config.ts
```diff
@@ -1,6 +1,9 @@
+import { DebugOptions } from './helpers/debugDeep';
+
 export type Config = {
   /** Default timeout, in ms, for `waitFor` and `findBy*` queries. */
   asyncUtilTimeout: number;
+  debugOptions?: Partial<DebugOptions>;
```
mdjastrzebski, line 6: RTL/DTL convention seems to be [`defaultXxx`](https://testing-library.com/docs/dom-testing-library/api-configuration#defaulthidden)
```suggestion
  defaultDebugOptions?: Partial<DebugOptions>;
```

src/helpers/format.ts
```diff
@@ -1,10 +1,58 @@
...
+  input: ReactTestRendererJSON | ReactTestRendererJSON[],
+  options: FormatOptions = {}
+) =>
   prettyFormat(input, {
-    plugins: [plugins.ReactTestComponent, plugins.ReactElement],
+    plugins: [getCustomPlugin(options.filterProps)],
     highlight: true,
   });
 
+const getCustomPlugin = (filterProps?: FilterNodeByTypeFunction): NewPlugin => {
```
mdjastrzebski, line 23:
```suggestion
const reactTestComponentPlugin = (filterProps?: FilterNodeByTypeFunction): NewPlugin => {
```

src/helpers/format.ts
```diff
@@ -1,10 +1,58 @@
...
+  filterProps?: FilterNodeByTypeFunction;
+};
+
+const format = (
+  input: ReactTestRendererJSON | ReactTestRendererJSON[],
+  options: FormatOptions = {}
+) =>
   prettyFormat(input, {
-    plugins: [plugins.ReactTestComponent, plugins.ReactElement],
+    plugins: [getCustomPlugin(options.filterProps)],
```
mdjastrzebski, line 19: Seems like we might as well use the orginal ReactElement plugin and just tweak the ReactTestComponent
```suggestion
    plugins: [reactTestComponentPlugin(options.filterProps), plugins.ReactElement],
```

src/helpers/format.ts
```diff
@@ -1,10 +1,58 @@
...
     highlight: true,
   });
 
+const getCustomPlugin = (filterProps?: FilterNodeByTypeFunction): NewPlugin => {
+  return {
+    test: (val) =>
+      plugins.ReactTestComponent.test(val) || plugins.ReactElement.test(val),
+    serialize: (val, config, indentation, depth, refs, printer) => {
```
mdjastrzebski, line 27: As mentioned above you should be able to make this impl. easier by just focusing on `ReactTestComponent` part and using `plugins.ReactElement` as-is.

typings/index.flow.js
```diff
@@ -306,8 +306,22 @@ interface Thenable {
...
+  node: ReactTestRendererJSON
+) => boolean;
+
+type FormatOptions = {
+  filterProps?: FilterNodeByTypeFunction,
+};
+
+type DebugOptions = {
+  message?: string,
+} & FormatOptions;
```
mdjastrzebski, line 321: nit: I would go with interface here, but this is purely personal style pref.

```suggestion
interface FormatOptions {
  filterProps?: FilterNodeByTypeFunction,
};

interface DebugOptions extends FormatOptions {
  message?: string,
}
```

typings/index.flow.js
```diff
@@ -306,8 +306,22 @@ interface Thenable {
...
+
 type Debug = {
-  (message?: string): void,
+  (options?: string | DebugOptions): void,
```
mdjastrzebski, line 324:
```suggestion
  (options?: DebugOptions | string): void,
```

website/docs/API.md
```diff
@@ -167,15 +168,24 @@ Usually you should not need to call `unmount` as it is done automatically if you
...
 
 ```ts
-debug(message?: string): void
+
+interface DebugOptions { 
+  message?: string;
+  filterProps?: FilterNodeByTypeFunction;
+}
+
+debug(options?: string | DebugOptions): void
```
mdjastrzebski, line 177:
```suggestion
debug(options?: DebugOptions | string): void
```

website/docs/API.md
```diff
@@ -167,15 +168,24 @@ Usually you should not need to call `unmount` as it is done automatically if you
...
 
-Pretty prints deeply rendered component passed to `render` with optional message on top.
+Pretty prints deeply rendered component passed to `render`. 
+
+You can provide a message that will be printed on top either by calling debug with a string or using the message property in debugOptions
 
 ```jsx
 render(<Component />);
 
 screen.debug('optional message');
```
mdjastrzebski, line 187: Let's suggest them only the modern way.
```suggestion
```

website/docs/API.md
```diff
@@ -167,15 +168,24 @@ Usually you should not need to call `unmount` as it is done automatically if you
...
+screen.debug({message: 'optional message'});
```
mdjastrzebski, line 188: formatting:
```suggestion
screen.debug({ message: 'optional message' });
```

website/docs/API.md
```diff
@@ -190,6 +200,16 @@ optional message
...
 
+You can use the filterProps option to filter the props that will be printed : 
+
+```jsx
+render(<View style={{backgroundColor: 'red'}}/>);
+
+debug({filterProps : (propName, propValue, node) => propName !== 'style'})
+```
+
+This will log the rendered JSX without style props
```
mdjastrzebski, line 211:
```suggestion
This will log the rendered JSX without the `style` prop.
```

website/docs/API.md
```diff
@@ -735,6 +755,9 @@ function configure(options: Partial<Config>)  {}
 
 Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementToBeRemoved`) and `findBy*` queries. Defaults to 1000 ms.
 
+#### `debugOptions` option
+
+Default debugOptions to be used when calling debug. These default options will be overridden by the ones you specify directly when calling debug
```
mdjastrzebski, line 760:
```suggestion
Default `debugOptions` to be used when calling `debug()`. These default options will be overridden by the ones you specify directly when calling `debug()`.
```

mdjastrzebski (commented):
I've done initial review on implementation and left some comments there. Overall I like the code so far, looks pretty solid. 

The point I want to rise is the desired API we want to have. 

The current API is:
```
export type FormatOptions = {
  filterProps?: (
    propName: string,
    propValue: unknown,
    node: ReactTestRendererJSON
  ) => boolean;
};
```
While it's quite flexible, I think we could investigate alternative approaches, based on the use cases we want to support.

### Idea 1: allowlist/blocklist
As far as I understand the primary use case is to print only some relevant props and/or exclude some props that are cluttering the output. We are rarely interested in inspecting the prop value. We could achieve that by somewhat easier API:

```ts
type FormatOptions = {
  allowedProps?: string[];
  ignoredProps?: string[];
}
```

If specified `allowedProps` will contain the names props that will be printed, no other props would be printed by `format`. Otherwise, all props will be printed except the ones specified in `ignoredProps`.

Pros: simpler API for majority use case
Cons: less flexible

### Idea 2: ability to transform props
Opposite to idea 1, this makes the API more flexible by allow prop transformation before printing. This could allow for both removal of props that are not interesting to as, or applying some transformation to their value that would improve the readability:

Version 1 - per prop
```ts
type FormatOptions = {
  mapProp?: (
    propName: string,
    propValue: unknown,
    node: ReactTestRendererJSON
  ) => [name, value];
}
```

Version 2 - per node
```ts
type FormatOptions = {
  mapProps?: (props: Record<string, unknown>) => Record<string, unknown>; // output props to be displayed
}
```

### Examples
Original:
```ts
debug(<JSX />, { filterProps: (name, value, node) => name === 'style' });
debug(<JSX />, { filterProps: (name, value, node) => ['style', 'testID'].includes(name) });
```

Idea 1:
```ts
debug(<JSX />, { allowedProps: ['style'] });
debug(<JSX />, { allowedProps: ['style', 'testID'] });
```

Idea 2 - per prop:
```ts
debug(<JSX />, { mapProp: (name, value, node) => name === 'style' ? [name, value] : undefined });
debug(<JSX />, { mapProp: (name, value, node) => ['style', 'testID'].includes(name ? [name, value] : undefined });
```

Idea 2 - per node:
```ts
debug(<JSX />, { mapProps: ({ style }) => ({ style }) });
debug(<JSX />, { mapProps: ({ style, testID }) => ({ style, testID }) });
```

@pierrezimmermannbam could you provide typical examples you envision this option will handle so that we can figure out which API options is the best to address it.

src/__tests__/__snapshots__/render.test.tsx.snap
```diff
@@ -124,6 +124,224 @@ exports[`debug changing component: bananaFresh button message should now be "fre
...
+  <Text>
+    hello
+  </Text>
+</View>"
+`;
+
+exports[`debug with all props filtered 1`] = `
+"<View>
+  <Text>
+    Is the banana fresh?
```
AugustinLF, line 152: Shouldn't `children` be filtered also?

AugustinLF (approved):
~Looks good to me, my feedback has been addressed, I'll let you address @mdjastrzebski comments~

AugustinLF (commented):
Good for me, I'll let you address @mdjastrzebski's comment meanwhile. I won't hit the `Approve` button to be sure it doesn't get merged by mistake :)

pierrezimmermannbam:
@mdjastrzebski you're right we should consider alternative approaches. I plan to use this feature in the following ways : 

- remove globally some props that aren't useful for debugging with configure api (style or path for instance)
- create custom debug utils that would be used for debugging each query using screen (for instance debugText would filter all props except children)
- use a more custom version in tests if the need arises but this should be pretty rare

Overall I'd mostly use this either with configure or have a set of utils that I would write only once so I like the idea of having a rather flexible api. However It should be accessible to users too so maybe we could document it a bit more with some use cases that we find relevant.

AllowedProps and ignoredProps is probably simpler than filterProps so I think it's interesting. I'm  not sure however that it's that much simpler, plus i'm afraid that there may be some conflicts between ignored and allowed props especially if one has been set through configure. Also i was thinking of globally filtering props whose value is a function which you cannot do without access to the propValue. I also like to have a single option for filtering props, because we may add other options later 
and the more there are the more it's going to be complicated to understand all of them.

mapProps is very interesting though, being able to flatten styles in the debug function for instance could be a very useful option. Are there any other use cases though ? Maybe directly having a flattenStyle option would be more convenient if that's not the case. What I'm afraid of is that it seems less obvious to me why you'd want to map props than why you'd want to filter props but it would indeed allow way more flexibility.  I think I wouldn't mind having both the filterProps and the mapProps option though. 

What do you think would be the best api based on the examples I provided ?

mdjastrzebski:
@pierrezimmermannbam not sure if you're participating in Hacktoberfest but if so I've added `hacktopberfest-accept`, I plan to review this again in the following days.

src/__tests__/__snapshots__/render.test.tsx.snap
mdjastrzebski: I think this is due to `children` prop being extracted from `props` in both TestInstance and JSON. Imo we should keep this as is, as a reasonable behaviour for most cases.

mdjastrzebski:
@pierrezimmermannbam 

Imo based on your use cases I think `mapProps` (per node version) makes most sense and is also the most flexible approach. It allows for both filtering & transforming props (e.g. flatten styles), while also allowing for basic output on a prop combination, e.g. output `style` prop only if `disabled` props is not set, etc. I'm curious about @AugustinLF take on the API.

re accept list/ignore list priority: I think a reasonable assumption is that `accept` list takes priority over `ignore` list. If you have `accept` list then you do not use `ignore` list at all.

re ease of use: we should have some documentation examples for common use cases, that should clarify the usage.

re having both `filterProp` and `mapProps` that would make the API a bit confusing as we would have two functions doing very similar things, and also it's not clear how should they interact. Which should be used if both are present, etc

I would be wary of introducing niche options like `flattenStyles` as it can be done with `mapProps` and imo is not worth to complicating our API.

AugustinLF:
I think it's fair to go with `mapProps`. It is an option for advanced users, so we should aim it to make it complete. And yes, for things like `flattenStyles` we could offer it as a recipe. It doesn't have to be in our API. Or if highly requested, we could expose it as an export, that could then be passed to the `mapProps`.

pierrezimmermannbam:
I also agree, I'll change the api and add some examples of usage in the documentation

package.json
pierrezimmermannbam: It's not really related to this pr but I noticed that mocks weren't automatically cleared because I had to add tests on debug and since the calls to console.log were not cleared, my tests did not pass after I ran them with the other ones. I think tests should all pass when they are run by themselves and should not depend on other tests and automatically clearing mocks call history helps achieving that

src/__tests__/__snapshots__/render.test.tsx.snap
pierrezimmermannbam: I was surprised by this as well but I actually think that this is fine, we'll want to keep the children prop in most cases and not having to explicitly include it is nice

pierrezimmermannbam:
@mdjastrzebski  I changed the api to use mapProps instead and updated the website docs to include more examples of the option usage. I also treated the comments you left. 

There is one last thing I'm not sure about though, should we keep the compatibility with previous versions for debugOptions ? The debug util is not meant to stay in the code, it's just a debugging tool so there should be little to no issues regarding the migration. I feel like it adds unnecessary complexity in the RNTL codebase while providing very little value to users. I think this could be merged as is and be included in a release before next major version but I would remove retro compatibility with the next major version, wdyt ?

src/render.tsx
```diff
@@ -135,18 +136,31 @@ function updateWithAct(
...
     const json = renderer.toJSON();
+    const debugOptions =
+      typeof options === 'string'
+        ? { ...defaultDebugOptions, message: options }
+        : { ...defaultDebugOptions, ...options };
+
+    if (typeof options === 'string') {
+      // eslint-disable-next-line no-console
+      console.warn(
+        'Please use debug({message; "message"}) instead of debug("message")'
```
mdjastrzebski, line 158:
```suggestion
        'Using debug("message") is deprecated and will be removed in future release, please use debug({ message; "message" }) instead.'
```

website/docs/API.md
```diff
@@ -737,6 +782,9 @@ function configure(options: Partial<Config>)  {}
 
 Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementToBeRemoved`) and `findBy*` queries. Defaults to 1000 ms.
 
+#### `defaultDebugOptions` option
+
+Default `debugOptions` to be used when calling `debug()`. These default options will be overridden by the ones you specify directly when calling `debug()`. Can be used namely to transform prop values for more readability or filter out props that have little value when debugging tests (e.g. svg's path prop).
```
mdjastrzebski, line 787: I would rather refer user to `debug` `options` docs for details rather than repeating the use cases.

```suggestion
Default `debugOptions` to be used when calling `debug()`. These default options will be overridden by the ones you specify directly when calling `debug()`. 
```

website/docs/API.md
```diff
@@ -223,6 +250,24 @@ Hold the value of latest render call for easier access to query and other functi
 
 Its value is automatically cleared after each test by calling [`cleanup`](#cleanup). If no `render` call has been made in a given test then it holds a special object that implements `RenderResult` but throws a helpful error on each property and method access.
 
+This can also be used to build test utils that would normally require to be in render scope, either in a test file or globally for your project. For instance in a test file where we fill a form in multiple tests:
```
mdjastrzebski, line 253: The concept of building test utils seems like a useful idea, however the `typeEmail` example seems quite artificial and I would not recommend using such util over explicit `fireEvent` + query.

@pierrezimmermannbam do you have idea for something that we could actually recommend to your users here?

website/docs/API.md
```diff
@@ -193,6 +202,24 @@ optional message
 </View>
 ```
 
+You can use the mapProps option to transform the props that will be printed :
```
mdjastrzebski, line 205:
```suggestion
You can use the `mapProps` option to transform the props that will be printed : 
```

website/docs/API.md
```diff
@@ -193,6 +202,24 @@ optional message
 </View>
 ```
 
+You can use the mapProps option to transform the props that will be printed : 
+
+```jsx
+render(<View style={{backgroundColor: 'red'}}/>);
+debug({ mapProps : ({style, ...props}) => ({props}) })
```
mdjastrzebski, line 209: formatting:
```suggestion
render(<View style={{ backgroundColor: 'red' }}/>);
debug({ mapProps : ({style, ...props}) => ({ props }) })
```

website/docs/API.md
```diff
@@ -193,6 +202,24 @@ optional message
...
+```
+
+This will log the rendered JSX without the `style` props. 
+
+This option can be used to target specfic props when debugging a query (for instance keeping only children prop when debugging a byText query) and you can also transform prop values so that they are more readable (e.g. flatten styles).
```
mdjastrzebski, line 214: Not sure about mentioning `children` prop in this case, as it is treated specially by `mapProps`

```suggestion
This option can be used to target specific props when debugging a query (for instance keeping only `children` prop when debugging a `getByText` query) and you can also transform prop values so that they are more readable (e.g. flatten styles).
```

website/docs/API.md
```diff
@@ -193,6 +202,24 @@ optional message
...
+
+The `children` prop cannot be filtered so the following will print all rendered components with all props but children filtered.
```
mdjastrzebski, line 216:
```suggestion
The `children` prop cannot be filtered out so the following will print all rendered components with all props but children filtered.
```

website/docs/API.md
```diff
@@ -170,15 +171,23 @@ Usually you should not need to call `unmount` as it is done automatically if you
...
+  message?: string;
+  mapProps?: MapPropsFunction;
+}
+
+debug(options?: DebugOptions | string): void
 ```
 
-Pretty prints deeply rendered component passed to `render` with optional message on top.
+Pretty prints deeply rendered component passed to `render`. 
+
```
mdjastrzebski, line 184: Pls add lower-level headers for options as we do for other APIs

```suggestion
#### `message` option
```

website/docs/API.md
```diff
@@ -193,6 +202,24 @@ optional message
 </View>
 ```
```
mdjastrzebski, line 204:
```suggestion
#### `mapProps` option
```

typings/index.flow.js
```diff
@@ -413,6 +426,7 @@ declare module '@testing-library/react-native' {
 
   declare interface Config {
     asyncUtilTimeout: number;
+    debugOptions?: $Shape<DebugOptions>;
```
mdjastrzebski, line 429:
```suggestion
    defaultDebugOptions?: $Shape<DebugOptions>;
```

typings/index.flow.js
```diff
@@ -349,8 +349,21 @@ interface Thenable {
...
+  node: ReactTestRendererJSON
+) => { [string]: mixed };
+
+type FormatOptions = {
+  mapProps?: MapPropsFunction,
+};
+
+type DebugOptions = {
+  message?: string,
+} & FormatOptions;
```
mdjastrzebski, line 363: `FormatOptions` does not seem to be used anywhere, let's merge it with `DebugOptions` for flow


```suggestion
type DebugOptions = {
  message?: string,
  mapProps?: MapPropsFunction,
};
```

src/render.tsx
```diff
@@ -135,18 +136,31 @@ function updateWithAct(
...
 }
 
 function debug(
   instance: ReactTestInstance,
   renderer: ReactTestRenderer
 ): DebugFunction {
-  function debugImpl(message?: string) {
+  function debugImpl(options?: DebugOptions | string) {
+    const { defaultDebugOptions } = getConfig();
     const json = renderer.toJSON();
```
mdjastrzebski, line 149: style, nit:
Pls move `const json = ...` line just above the `if (json)` statement, as it is not used during option calculation part.

src/helpers/format.ts
```diff
@@ -1,10 +1,53 @@
...
+  mapProps?: MapPropsFunction;
+};
+
+const format = (
+  input: ReactTestRendererJSON | ReactTestRendererJSON[],
+  options: FormatOptions = {}
+) =>
   prettyFormat(input, {
-    plugins: [plugins.ReactTestComponent, plugins.ReactElement],
+    plugins: [getCustomPlugin(options.mapProps)],
```
mdjastrzebski, line 18: I think it's still a good idea to make `getCustomPlugin` tweak only `ReactTestComponent` while passing `ReactElement` normally
```suggestion
    plugins: [getTestComponentPlugin(options.mapProps), plugins.ReactElement],
```

src/helpers/debugDeep.ts
```diff
@@ -1,18 +1,27 @@
...
  */
 export default function debugDeep(
   instance: ReactTestRendererJSON | ReactTestRendererJSON[],
-  message?: string
+  options?: DebugOptions | string
 ) {
+  const message = typeof options === 'string' ? options : options?.message;
+
+  const formatOptions =
+    typeof options === 'object' ? { mapProps: options.mapProps } : undefined;
```
mdjastrzebski, line 18: Should prevent crash if user passes `null`, because `typeof null === 'object'`

```suggestion
    typeof options === 'object' ? { mapProps: options?.mapProps } : undefined;
```

src/__tests__/render.test.tsx
```diff
@@ -191,11 +211,121 @@ test('debug changing component', () => {
...
     'bananaFresh button message should now be "fresh"'
   );
 });
 
+test('debug with only children prop', () => {
+  jest.spyOn(console, 'log').mockImplementation((x) => x);
+
+  const { debug, getByText } = render(<Banana />);
+
+  debug({ mapProps: ({ children }) => ({ children }) });
```
mdjastrzebski, line 224: This test seems confusing, as `children` in not included in the props. Pls adjust or remove the test.

src/__tests__/render.test.tsx
```diff
@@ -191,11 +211,121 @@ test('debug changing component', () => {
...
+      return filterProps;
+    },
+  });
+
+  // eslint-disable-next-line no-console
+  const mockCalls = (console.log as any as ConsoleLogMock).mock.calls;
+
+  expect(stripAnsi(mockCalls[0][0])).toMatchSnapshot();
+
+  expect(getByText('Change freshness!')).toBeTruthy();
```
mdjastrzebski, line 256: Having `getByText` assertion does not seem to contribute to the test.
```suggestion
```

src/__tests__/render.test.tsx
```diff
@@ -191,11 +211,121 @@ test('debug changing component', () => {
...
+  debug({
+    mapProps: (props, node) => (node.type === 'TextInput' ? props : {}),
+  });
+
+  // eslint-disable-next-line no-console
+  const mockCalls = (console.log as any as ConsoleLogMock).mock.calls;
+
+  expect(stripAnsi(mockCalls[0][0])).toMatchSnapshot();
+
+  expect(getByText('Change freshness!')).toBeTruthy();
```
mdjastrzebski, line 273: Having `getByText` assertion does not seem to contribute to the test.
```suggestion
```

src/__tests__/render.test.tsx
```diff
@@ -191,11 +211,121 @@ test('debug changing component', () => {
...
+  const { debug, getByText } = render(<Banana />);
+
+  debug({ mapProps: () => ({}) });
+
+  // eslint-disable-next-line no-console
+  const mockCalls = (console.log as any as ConsoleLogMock).mock.calls;
+
+  expect(stripAnsi(mockCalls[0][0])).toMatchSnapshot();
+
+  expect(getByText('Change freshness!')).toBeTruthy();
```
mdjastrzebski, line 288: Having `getByText` assertion does not seem to contribute to the test.
```suggestion
```

src/__tests__/render.test.tsx
```diff
@@ -191,11 +211,121 @@ test('debug changing component', () => {
   // eslint-disable-next-line no-console
   const mockCalls = (console.log as any as ConsoleLogMock).mock.calls;
 
-  expect(stripAnsi(mockCalls[4][0])).toMatchSnapshot(
+  expect(stripAnsi(mockCalls[0][0])).toMatchSnapshot(
```
mdjastrzebski, line 214: What caused the index change here?

src/helpers/format.ts
```diff
@@ -1,10 +1,53 @@
...
   });
 
+const getCustomPlugin = (mapProps?: MapPropsFunction): NewPlugin => {
+  return {
+    test: (val) =>
+      plugins.ReactTestComponent.test(val) || plugins.ReactElement.test(val),
+    serialize: (val, config, indentation, depth, refs, printer) => {
+      if (plugins.ReactTestComponent.test(val)) {
+        if (mapProps && val.props) {
+          val.props = mapProps(val.props, val);
```
mdjastrzebski, line 29: Not sure if it's a good idea to mutate the passed `val` object. Can we rather use object spread operator to create a modified copy?

mdjastrzebski (changes requested):
@pierrezimmermannbam This PR looks really solid. I've added some suggestions for finishing touches and edge cases. 

Pls add some real-life examples to the docs on how users could you this to improve their debugging output.

website/docs/API.md
pierrezimmermannbam: agreed, I added a link to debug options doc

website/docs/API.md
pierrezimmermannbam: This is something is sometimes do when I test forms and have a lot of tests, imo it makes tests more readable and very easy to understand while not making them harder to debug since you still have the same error messages. It's even more interesting when it allows to hide testID usage since the queries by testID do not really help to understand the test. An example of this is waiting for a loader to disapear, which is often done in every test when testing at page level

```tsx
const waitForLoaderToDisapear = async () => {
   await waitForElementToBeRemoved(() => screen.getByTestID('PageLoader'));
}
```

Does this example make more sense to you ?

website/docs/API.md
pierrezimmermannbam: I changed the order so that it is explained first that children prop cannot be filtered then give the example of usage, since I believe this will be one of the most common, at least until byRole queries become more popular

website/docs/API.md
pierrezimmermannbam: Very good idea I hadn't noticed this was done for other apis !

typings/index.flow.js
pierrezimmermannbam: Oh good catch, I keep missing the flow types updates

src/helpers/format.ts
pierrezimmermannbam: Absolutely, it also makes the code way simpler !

src/__tests__/render.test.tsx
pierrezimmermannbam: Oh you're right, I had written them before I realized the children prop wasn't filtered and it does not really make sense anymore. I deleted one test and modified the other one

src/__tests__/render.test.tsx
pierrezimmermannbam: Actually what I was trying to do here was make sure that filtering out props wouldn't really delete them but since the children prop is never filtered it doesn't make sense. I'll replace this by a more explicit and relevant test

src/__tests__/render.test.tsx
pierrezimmermannbam: The fact that mocks are being cleared automatically do to the change I made in jest.config.js. We were checking the 4th call because it was called in previous test but now mock calls are reset before each test

src/helpers/format.ts
pierrezimmermannbam: I had my doubt on this, this is why I added a test to verify that the original object is not modified and it's not the case so I think it's fine as is. Also I believe that using the spread operator would create a shallow copy so it would not really be a protection against modifications either

pierrezimmermannbam:
@mdjastrzebski  I think I answered all your comments. I added an example to the doc and detailed more the flatten style example, do you think of anything else ?

src/helpers/format.ts
mdjastrzebski: format works on JSON output not ReactTestInstances, but still invoking `format` would mutate the JSON output. I've changed that to object copy, not mutation.

mdjastrzebski (approved):
Looks great! Good work @pierrezimmermannbam 🚀 

I've made some cosmetic tweaks to docs and code before merging.

PR #1161, opened 2022-10-07
Query by role and accessibilityState and improve ByRole query errors

### LINKED ISSUE #1134: Supports accessibilityStates in ByRole queries
## Describe the Feature
Like in `@testing-library/dom` (see [docs](https://testing-library.com/docs/queries/byrole#api)), we should support accessibility states when querying by role.

The accessibilityStates existing are
- `disabled`: Indicates whether the element is disabled or not
- `selected`: Indicates whether a selectable element is currently selected or not
- `checked`: Indicates the state of a checkable element.
- `busy`: Indicates whether an element is currently busy or not
- `expanded`: Indicates whether an expandable element is currently expanded or collapsed

## Possible Implementations
We could like in DTL check which states correspond to which role and throw. But I'd treat that as a bonus and not a necessity.

## Related Issues
#1133

AugustinLF:
### Summary

RTL does have some form of validation to make sure that we don't use incompatible combinations (roles/states that don't make sense, like a checked button), but I'm not sure about that, and it is out of scope, we can create an additional issue if we feel like it.

Closes #1134


src/queries/role.ts
```diff
@@ -28,22 +29,73 @@ const matchAccessibleNameIfNeeded = (
...
+
+      if (!matchRole) return false;
+
+      if (options?.name) {
+        if (!matchAccessibleNameIfNeeded(node, options.name)) {
+          return false;
+        }
+      }
+
+      return accessibilityStates.every((accessibilityState) => {
```
mdjastrzebski, line 57: Can we extract the last part into `matchAccessibilityStateIfNeeded`?

src/queries/role.ts
```diff
@@ -28,22 +29,73 @@ const matchAccessibleNameIfNeeded = (
...
-    );
+        matchStringProp(node.props.accessibilityRole, role);
+
+      if (!matchRole) return false;
+
+      if (options?.name) {
+        if (!matchAccessibleNameIfNeeded(node, options.name)) {
+          return false;
+        }
+      }
```
mdjastrzebski, line 55: This should equivalent with less nesting
```suggestion
      const matchesAccessibleName = matchAccessibleNameIfNeeded(node, options.name);
      if (!matchesAccessibleName) { 
        return false;
      }
```

src/queries/role.ts
```diff
@@ -28,22 +29,73 @@ const matchAccessibleNameIfNeeded = (
...
+
+      return accessibilityStates.every((accessibilityState) => {
+        const queriedState = options?.[accessibilityState];
+
+        if (typeof queriedState !== 'undefined') {
```
mdjastrzebski, line 60: We might alter the order of the checks and early false exits from cheapest to most expensive to check. So checking role name is the cheapes, then checking accessiblity state, then accessibleName

src/queries/role.ts
mdjastrzebski: Is there an option to use the same predict for checking a11yState in both `*ByRole` and `*ByA11yState` queries?

AugustinLF: Not as is, because `ByA11yState` checks for exact match (options == state), while here we want all options to match state, but it's fine skipping some. And we do match `false` and `undefined`.

mdjastrzebski: I don't think this is actually the case. Looks like `*ByA11yState` is only checking the expected props and ignoring others:

```ts
  function queryAllByA11yStateFn(state) {
    return instance.findAll(
      (node) =>
        matchObjectProp(node.props.accessibilityState, state)
    );
  };
```

```ts
export function matchObjectProp<T extends Record<string, unknown>>(
  prop: T | undefined,
  matcher: T
): boolean {
  if (!prop || Object.keys(matcher).length === 0) {
    return false;
  }

  return (
    Object.keys(prop).length !== 0 &&
    Object.keys(matcher).every((key) => prop[key] === matcher[key])
  );
}
```

mdjastrzebski: The only difference I can spot is that `*ByRole` `state` checks default missing state options as `false` while `*ByA11yState` treats `undefined` and `false` as different values. I think both of these queries should have the behaviour you proposed for `*ByRole` query, as we want to model the underlying RN behaviour, which treats `undefined` the same as `false`. Additionally such discrepancy will confuse the users.

Therefore, I think we should fix by `*ByA11yState` query handling. I would even consider that rather to be a bug fix to that query, rather than a change in behaviour, but we can consider both options.

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +135,415 @@ describe('supports name option', () => {
...
+      );
+
+      expect(getByRole('checkbox', { checked: 'mixed' })).toBeTruthy();
+    });
+
+    it('does not return mixed checkboxes when querying for checked: true', () => {
+      const { queryByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="checkbox"
+          accessibilityState={{ checked: 'mixed' }}
```
mdjastrzebski, line 295: 👍🏻 for supporting `mixed` value

src/queries/role.ts
AugustinLF: Yep you're right about skipping non relevant props, but it doesn't check implicit state, so all the tests in the form of `returns an implicitly enabled element` fail.

mdjastrzebski (commented):
Kudos for implementing this feature @AugustinLF 

1. This implementation should automatically support upcoming `aria-states` props, when checked on host components, as they are taken into consideration when passing them from exported composite component. See [here](https://github.com/facebook/react-native/blob/43636267011a97490ed7495b08e500c5d0d54872/Libraries/Components/View/View.js#L61), not sure if these are present in the current released version of RN.

2. It seems that some components take into account other props as well, when calcualting `accessibilityState` that gets passed to host component, e.g. [`TouchableOpacity`](https://github.com/facebook/react-native/blob/43636267011a97490ed7495b08e500c5d0d54872/Libraries/Components/Touchable/TouchableOpacity.js#L231) also checks `disabled` prop. The current implementation already handles that but we might add some tests to verify it.

src/queries/role.ts
mdjastrzebski: Therefore, I think we should change `*ByA11yState` matching to include `default` states, because the underlying RN behaviour is to treat `undefined` in the same way as `false`. Otherwise that query contradicts platform behaviour, while our goal is to simulate RN behaviour the best we can.

src/queries/role.ts
```diff
@@ -28,22 +29,76 @@ const matchAccessibleNameIfNeeded = (
...
+
+      return matchAccessibleNameIfNeeded(node, options?.name);
+    });
   };
 
-const getMultipleError = (role: TextMatch) =>
-  `Found multiple elements with accessibilityRole: ${String(role)} `;
-const getMissingError = (role: TextMatch) =>
-  `Unable to find an element with accessibilityRole: ${String(role)}`;
+const computeErrorMessage = (role: TextMatch, options: ByRoleOptions = {}) => {
```
mdjastrzebski, line 76:
```suggestion
const buildErrorMessage = (role: TextMatch, options: ByRoleOptions = {}) => {
```

src/queries/role.ts
```diff
@@ -28,22 +29,76 @@ const matchAccessibleNameIfNeeded = (
...
+  ) {
+    errorMessage += ', accessibilityStates:';
+    accessibilityStates.forEach((accessibilityState) => {
+      if (options[accessibilityState]) {
+        errorMessage += ` ${accessibilityState}:${options[accessibilityState]}`;
+      }
+    });
+  }
+
+  return errorMessage;
```
mdjastrzebski, line 96: style:
Maybe we can use following pattern for simplier formatting:
```
const errorParts = [`accessibilityRole: ${String(role)}`];
// ...
errorParts.push(`name: ${String(options.name)}`;
// ...
return errorParts.join(', ')
```

src/queries/role.ts
mdjastrzebski: I've logged #1165 in order to fix `*ByA11yState` query behavior

src/queries/role.ts
```diff
@@ -28,22 +29,76 @@ const matchAccessibleNameIfNeeded = (
...
+            );
+          } else {
+            return false;
+          }
+        }
+      );
+
+      if (mismatchAccessibilityState) {
+        return false;
+      }
```
mdjastrzebski, line 70: readability: Pls extract to `matchAccessibleStateIfNeeded` or similar

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+  describe('disabled', () => {
+    test('returns a disabled element when required', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="button"
+          accessibilityState={{ disabled: true }}
+        />
+      );
+
+      expect(getByRole('button', { disabled: true })).toBeTruthy();
```
mdjastrzebski, line 156:
```suggestion
      expect(getByRole('button', { disabled: true })).toBeTruthy();
      expect(queryByRole('button', { disabled: false })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+        getByRole('button', { name: 'Save', disabled: true }).props.testID
+      ).toBe('correct');
+    });
+
+    test('returns an implicitly enabled element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity accessibilityRole="button"></TouchableOpacity>
+      );
+
+      expect(getByRole('button', { disabled: false })).toBeTruthy();
```
mdjastrzebski, line 185:
```suggestion
      expect(getByRole('button', { disabled: false })).toBeTruthy();
      expect(queyrByRole('button', { disabled: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    test('returns an explicitly enabled element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="button"
+          accessibilityState={{ disabled: false }}
+        ></TouchableOpacity>
+      );
+
+      expect(getByRole('button', { disabled: false })).toBeTruthy();
```
mdjastrzebski, line 196:
```suggestion
      expect(getByRole('button', { disabled: false })).toBeTruthy();
      expect(queyrByRole('button', { disabled: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+        <TouchableOpacity
+          accessibilityRole="button"
+          accessibilityState={{ disabled: true }}
+        ></TouchableOpacity>
+      );
+
+      expect(queryByRole('button', { disabled: false })).toBe(null);
+    });
+
+    test('returns elements using the built-in disabled prop', () => {
```
mdjastrzebski, line 210: Can we simplify this test to only contain a single `TouchableOpacity` with `disabled={true}`?

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+  describe('selected', () => {
+    test('returns a selected element when required', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="tab"
+          accessibilityState={{ selected: true }}
+        />
+      );
+
+      expect(getByRole('tab', { selected: true })).toBeTruthy();
```
mdjastrzebski, line 253:
```suggestion
      expect(getByRole('tab', { selected: true })).toBeTruthy();
      expect(queryByRole('tab', { selected: false })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+        getByRole('tab', { name: 'Save', selected: true }).props.testID
+      ).toBe('correct');
+    });
+
+    test('returns an implicitly non selected element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity accessibilityRole="tab"></TouchableOpacity>
+      );
+
+      expect(getByRole('tab', { selected: false })).toBeTruthy();
```
mdjastrzebski, line 282:
```suggestion
      expect(getByRole('tab', { selected: false })).toBeTruthy();
      expect(queryByRole('tab', { selected: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    test('returns an explicitly non selected element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="tab"
+          accessibilityState={{ selected: false }}
+        ></TouchableOpacity>
+      );
+
+      expect(getByRole('tab', { selected: false })).toBeTruthy();
```
mdjastrzebski, line 293:
```suggestion
      expect(getByRole('tab', { selected: false })).toBeTruthy();
      expect(queryByRole('tab', { selected: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+  describe('checked', () => {
+    test('returns a checked element when required', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="checkbox"
+          accessibilityState={{ checked: true }}
+        />
+      );
+
+      expect(getByRole('checkbox', { checked: true })).toBeTruthy();
```
mdjastrzebski, line 317:
```suggestion
      expect(getByRole('checkbox', { checked: true })).toBeTruthy();
      expect(queryByRole('checkbox', { checked: false })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    it('returns `mixed` checkboxes', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="checkbox"
+          accessibilityState={{ checked: 'mixed' }}
+        />
+      );
+
+      expect(getByRole('checkbox', { checked: 'mixed' })).toBeTruthy();
```
mdjastrzebski, line 328:
```suggestion
      expect(getByRole('checkbox', { checked: 'mixed' })).toBeTruthy();
      expect(queryByRole('checkbox', { checked: true })).toBeFalsy();
      expect(queryByRole('checkbox', { checked: false })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    it('does not return mixed checkboxes when querying for checked: true', () => {
+      const { queryByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="checkbox"
+          accessibilityState={{ checked: 'mixed' }}
+        />
+      );
+
+      expect(queryByRole('checkbox', { checked: false })).toBe(null);
```
mdjastrzebski, line 339:
```suggestion
      expect(queryByRole('checkbox', { checked: false })).toBeFalsy;
      expect(queryByRole('checkbox', { checked: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+        getByRole('checkbox', { name: 'Save', checked: true }).props.testID
+      ).toBe('correct');
+    });
+
+    test('returns an implicitly non checked element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity accessibilityRole="checkbox"></TouchableOpacity>
+      );
+
+      expect(getByRole('checkbox', { checked: false })).toBeTruthy();
```
mdjastrzebski, line 368:
```suggestion
      expect(getByRole('checkbox', { checked: false })).toBeTruthy();
      expect(queryByRole('checkbox', { checked: true })).toBeFalsy();
      expect(queryByRole('checkbox', { checked: 'mixed' })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    test('returns an explicitly non checked element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="checkbox"
+          accessibilityState={{ checked: false }}
+        ></TouchableOpacity>
+      );
+
+      expect(getByRole('checkbox', { checked: false })).toBeTruthy();
```
mdjastrzebski, line 379:
```suggestion
      expect(getByRole('checkbox', { checked: false })).toBeTruthy();
      expect(queryByRole('checkbox', { checked: true })).toBeFalsy();
      expect(queryByRole('checkbox', { checked: 'mixed' })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+        getByRole('button', { name: 'Save', busy: true }).props.testID
+      ).toBe('correct');
+    });
+
+    test('returns an implicitly non busy element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity accessibilityRole="button"></TouchableOpacity>
+      );
+
+      expect(getByRole('button', { busy: false })).toBeTruthy();
```
mdjastrzebski, line 432:
```suggestion
      expect(getByRole('button', { busy: false })).toBeTruthy();
      expect(getByRole('button', { busy: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    test('returns an explicitly non busy element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="button"
+          accessibilityState={{ busy: false }}
+        ></TouchableOpacity>
+      );
+
+      expect(getByRole('button', { busy: false })).toBeTruthy();
```
mdjastrzebski, line 443:
```suggestion
      expect(getByRole('button', { busy: false })).toBeTruthy();
      expect(getByRole('button', { busy: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+  describe('expanded', () => {
+    test('returns a expanded element when required', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="button"
+          accessibilityState={{ expanded: true }}
+        />
+      );
+
+      expect(getByRole('button', { expanded: true })).toBeTruthy();
```
mdjastrzebski, line 467:
```suggestion
      expect(getByRole('button', { expanded: true })).toBeTruthy();
      expect(getByRole('button', { expanded: false })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+        getByRole('button', { name: 'Save', expanded: true }).props.testID
+      ).toBe('correct');
+    });
+
+    test('returns an implicitly non expanded element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity accessibilityRole="button"></TouchableOpacity>
+      );
+
+      expect(getByRole('button', { expanded: false })).toBeTruthy();
```
mdjastrzebski, line 496:
```suggestion
      expect(getByRole('button', { expanded: false })).toBeTruthy();
      expect(getByRole('button', { expanded: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+
+    test('returns an explicitly non expanded element', () => {
+      const { getByRole } = render(
+        <TouchableOpacity
+          accessibilityRole="button"
+          accessibilityState={{ expanded: false }}
+        ></TouchableOpacity>
+      );
+
+      expect(getByRole('button', { expanded: false })).toBeTruthy();
```
mdjastrzebski, line 507:
```suggestion
      expect(getByRole('button', { expanded: false })).toBeTruthy();
      expect(getByRole('button', { expanded: true })).toBeFalsy();
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+    );
+  });
+
+  test('gives a descriptive error message when querying with a role, a name and accessibility state', () => {
+    const { getByRole } = render(<View />);
+
+    expect(() =>
+      getByRole('button', { name: 'Save', disabled: true })
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Unable to find an element with accessibilityRole: button, name: Save, accessibilityStates: disabled:true"`
```
mdjastrzebski, line 598:
```suggestion
      `"Unable to find an element with accessibilityRole: button, name: Save, accessibilityState: { disabled:true" }`
```

src/queries/__tests__/role.test.tsx
```diff
@@ -135,3 +142,470 @@ describe('supports name option', () => {
...
+    );
+  });
+
+  test('gives a descriptive error message when querying with a role and an accessibility state', () => {
+    const { getByRole } = render(<View />);
+
+    expect(() =>
+      getByRole('button', { disabled: true })
+    ).toThrowErrorMatchingInlineSnapshot(
+      `"Unable to find an element with accessibilityRole: button, accessibilityStates: disabled:true"`
```
mdjastrzebski, line 608:
```suggestion
      `"Unable to find an element with accessibilityRole: "button", accessibilityStates: { disabled: true }"`
```

mdjastrzebski (changes requested):
The overall structure looks correct now 💯

I've added some comments related to various implementation details to polish things out.

We also need to have our documentation updated to account for new options. 

I've logged a separate #1165 to fix make `*ByA11yStates` to have the same matching behaviour as in this PR.

src/queries/__tests__/role.test.tsx
AugustinLF: Those three RN component use different implementation behind the hood, so my goal was to test all of them (`TouchableOpacity` and `TouchableHighlight` extend `TouchableWithoutFeedback`), and I felt like covering our bases.

On the other hand I do see your point, since we're just testing RN internal implementation.

I'll keep the three ones, if you feel strongly about that (or splitting it in three tests) let me know).

mdjastrzebski: It makes sense to test all three `TouchablesXxx` and I would also metion `Pressable` in that context. Maybe you can use `test.each` to automate testing over these component variants?

src/queries/role.ts
```diff
@@ -28,22 +29,76 @@ const matchAccessibleNameIfNeeded = (
...
+
+      const mismatchAccessibilityState = accessibilityStates.some(
+        (accessibilityState) => {
+          const queriedState = options?.[accessibilityState];
+
+          if (typeof queriedState !== 'undefined') {
+            return (
+              queriedState !==
+              // default to false because disabled:undefined is equivalent to disabled:false
+              (node.props.accessibilityState?.[accessibilityState] ?? false)
```
mdjastrzebski, line 60: Seems like `checked` does not have default to `false` behaviour in real world tests for both iOS and Android. See more details [here](https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State). Doing some quick testing on iOS only, also `expanded` button does not default to `false`. For `busy`, `disabled` and `selected` it does however default to `false`.

I wonder what to do with that fact. We might want to default to `false` only for `busy`, `disabled` and `selected`

src/queries/__tests__/role.test.tsx
AugustinLF: I tried, but the difference in types (`Button` doesn't use `children`), it leads to pretty ugly code, and I'd say that it's not worth it, it'll make the tests harder to read, and here it's fairly simple.

AugustinLF:
@mdjastrzebski should have addressed all your comments I believe!

mdjastrzebski:
@AugustinLF I'll need pause a couple of days on open source work as I've got cold-like symptoms. I'll review the PRs as soon as I get better.

src/queries/role.ts
```diff
@@ -14,7 +15,9 @@ import type {
 
 type ByRoleOptions = {
   name?: TextMatch;
-};
+} & AccessibilityState;
+
+type AccessibilityStateField = keyof AccessibilityState;
```
mdjastrzebski, line 20: nit: more typical name would be using Key suffix
```suggestion
type AccessibilityStateKey = keyof AccessibilityState;
```

src/queries/role.ts
```diff
@@ -28,22 +31,90 @@ const matchAccessibleNameIfNeeded = (
...
+];
+
 const queryAllByRole = (
   instance: ReactTestInstance
 ): ((role: TextMatch, options?: ByRoleOptions) => Array<ReactTestInstance>) =>
   function queryAllByRoleFn(role, options) {
-    return instance.findAll(
-      (node) =>
+    return instance.findAll((node) => {
+      // run the cheapest checks first, and early exit too avoid unneeded computations
```
mdjastrzebski, line 76: style: we do not need to put the early returns by hand. JS is smart enough to skil calculating `b()` in `a() && b()` when it knows that `a() == false`

```suggestion
    return instance.findAll((node) => {
      typeof node.type === 'string' &&
      matchStringProp(node.props.accessibilityRole, role) &&
      matchAccessibleStateIfNeeded(node, options) &&
      matchAccessibleNameIfNeeded(node, options?.name) 
```

| The logical AND expression is a short-circuit operator. As each operand is converted to a boolean, if the result of one conversion is found to be false, the AND operator stops and returns the original value of that falsy operand; it does not evaluate any of the remaining operands.

Consider the pseudocode below.

(some falsy expression) && expr
The expr part is never evaluated because the first operand (some falsy expression) is evaluated as [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy). If expr is a function, the function is never called. See the example below:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND#short-circuit_evaluation

website/docs/Queries.md
```diff
@@ -193,6 +193,16 @@ const element = screen.getByRole('button');
 
 `name`: Finds an element with given `accessibilityRole` and an accessible name (equivalent to `byText` or `byLabelText` query).
 
+`disabled`: You can filter elements by their disabled state. The possible values are `true` or `false`. See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `disabled` state.
```
mdjastrzebski, line 196: We should mention next to each item whether missing value is or isn't treated as false.

mdjastrzebski (commented):
I've done a final round of review. Overall looks very good for me:
- All the logic seems correct
- we have a  **comprehensive** set of tests

I've left two readability/style final comments & requested to document behaviour around defaulting or not to `false` for each prop. 

I didn't give approve, only because as I worry that @thymikee could merge it before addressing these ;-)

src/queries/role.ts
AugustinLF: Yeah, I'm well aware, I think it's just a left over from when we didn't have helpers function for all cases. I do tend to find explicit return more readable in a lot of cases, but here it probably doesn't matter!

AugustinLF:
@mdjastrzebski a bit delayed, but everything should have been addressed now :)

AugustinLF:
Also added typedefs following the changes in #1175, see 3e01734838a37b361282c49386bb69435051531b

website/docs/Queries.md
```diff
@@ -255,6 +260,16 @@ const element = screen.getByRole('button');
 
 `name`: Finds an element with given `accessibilityRole` and an accessible name (equivalent to `byText` or `byLabelText` query).
 
+`disabled`: You can filter elements by their disabled state. The possible values are `true` or `false`. Querying `disabled: false` will also query elements with `disabled: undefined` (see the [wiki](https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State) for more details). See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `disabled` state.
```
mdjastrzebski, line 263:
```suggestion
`disabled`: You can filter elements by their disabled state. The possible values are `true` or `false`. Querying `disabled: false` will also match elements with `disabled: undefined` (see the [wiki](https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State) for more details). See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `disabled` state.
```

website/docs/Queries.md
```diff
@@ -255,6 +260,16 @@ const element = screen.getByRole('button');
...
+
+`selected`: You can filter elements by their selected state. The possible values are `true` or `false`. Querying `selected: false` will also query elements with `selected: undefined` (see the [wiki](https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State) for more details). See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `selected` state.
```
mdjastrzebski, line 265:
```suggestion
`selected`: You can filter elements by their selected state. The possible values are `true` or `false`. Querying `selected: false` will also match elements with `selected: undefined` (see the [wiki](https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State) for more details). See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `selected` state.
```

mdjastrzebski (approved):
Awesome work @AugustinLF!

`*ByRole` gets really powerful these days: first `name`, and now a11y states

website/docs/Queries.md
```diff
@@ -255,6 +260,16 @@ const element = screen.getByRole('button');
...
+
+`checked`: You can filter elements by their checked state. The possible values are `true`, `false`, or `"mixed"`. See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `checked` state.
+
+`busy`: You can filter elements by their busy state. The possible values are `true` or `false`. Querying `busy: false` will also query elements with `busy: undefined` (see the [wiki](https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State) for more details). See [React Native's accessibilityState](https://reactnative.dev/docs/accessibility#accessibilitystate) docs to learn more about the `busy` state.
```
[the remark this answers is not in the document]
AugustinLF: [bc26162](https://github.com/callstack/react-native-testing-library/pull/1161/commits/bc26162bce076c822dff2512bad71d82828feca3)

website/docs/Queries.md
AugustinLF: [bc26162](https://github.com/callstack/react-native-testing-library/pull/1161/commits/bc26162bce076c822dff2512bad71d82828feca3)

website/docs/Queries.md
AugustinLF: [bc26162](https://github.com/callstack/react-native-testing-library/pull/1161/commits/bc26162bce076c822dff2512bad71d82828feca3)

mdjastrzebski:
Let's merge it!

AugustinLF:
Thanks for the help! When are we shipping a new release, we've had at least two improvements to `getByRole` in since `11.2` 👀

mdjastrzebski:
@AugustinLF we've published 11.3.0 a couple of days ago, but I just created a release docs in GH: https://github.com/callstack/react-native-testing-library/releases/tag/v11.3.0

PR #1166, opened 2022-10-09
fix: *ByA11yState default value false value for busy, disabled & selected state

### LINKED ISSUE #1165: bug: `*ByA11yState` query ignores default state values of false
## Describe the bug

When using `*ByA11yState` queries they perform match given element only when given option has the exact values as passed. 

## Current behavior

* `getByA11yState({ disabled: true })` will match `<View accessibilityState={{ disabled: true }} />` (OK)
* `getByA11yState({ disabled: false })` will match `<View accessibilityState={{ disabled: false }} />` (OK)
* `getByA11yState({ disabled: false })` will NOT match `<View accessibilityState={{}} />` (NOT OK)

## Expected behavior

* `getByA11yState({ disabled: false })` should match `<View />`
* similarly for other state options we should assume the each of them has a default value of `false` if not specified otherwise.

## Versions

11.2.0 and below

## Related issues

#1161 - issue where we discovered the bug

CC: @thymikee @AugustinLF @pierrezimmermannbam

mdjastrzebski:
### Summary

Resolves #1165.

Based on empirical research results: https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State

### Test plan

Added additional test for matching behaviour.

### DIFF
--- a/src/helpers/matchers/accessibilityState.ts
+++ b/src/helpers/matchers/accessibilityState.ts
@@ -1,5 +1,6 @@
 import { AccessibilityState } from 'react-native';
 import { ReactTestInstance } from 'react-test-renderer';
+import { accessibilityStateKeys } from '../accessiblity';
 
 /**
  * Default accessibility state values based on experiments using accessibility
@@ -20,20 +21,16 @@ export function matchAccessibilityState(
   matcher: AccessibilityState
 ) {
   const state = node.props.accessibilityState;
-  return (
-    matchState(state, matcher, 'disabled') &&
-    matchState(state, matcher, 'selected') &&
-    matchState(state, matcher, 'checked') &&
-    matchState(state, matcher, 'busy') &&
-    matchState(state, matcher, 'expanded')
-  );
+  return accessibilityStateKeys.every((key) => matchState(state, matcher, key));
 }
 
 function matchState(
-  value: AccessibilityState,
+  state: AccessibilityState,
   matcher: AccessibilityState,
   key: keyof AccessibilityState
 ) {
-  const valueWithDefault = value?.[key] ?? defaultState[key];
-  return matcher[key] === undefined || matcher[key] === valueWithDefault;
+  return (
+    matcher[key] === undefined ||
+    matcher[key] === (state?.[key] ?? defaultState[key])
+  );
 }
--- a/src/helpers/matchers/accessibilityState.ts
+++ b/src/helpers/matchers/accessibilityState.ts
@@ -1,0 +1,39 @@
+import { AccessibilityState } from 'react-native';
+import { ReactTestInstance } from 'react-test-renderer';
+
+/**
+ * Default accessibility state values based on experiments using accessibility
+ * inspector/screen reader on iOS and Android.
+ *
+ * @see https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State
+ */
+const defaultState: AccessibilityState = {
+  disabled: false,
+  selected: false,
+  checked: undefined,
+  busy: false,
+  expanded: undefined,
+};
+
+export function matchAccessibilityState(
+  node: ReactTestInstance,
+  matcher: AccessibilityState
+) {
+  const state = node.props.accessibilityState;
+  return (
+    matchState(state, matcher, 'disabled') &&
+    matchState(state, matcher, 'selected') &&
+    matchState(state, matcher, 'checked') &&
+    matchState(state, matcher, 'busy') &&
+    matchState(state, matcher, 'expanded')
+  );
+}
+
+function matchState(
+  value: AccessibilityState,
+  matcher: AccessibilityState,
+  key: keyof AccessibilityState
+) {
+  const valueWithDefault = value?.[key] ?? defaultState[key];
+  return matcher[key] === undefined || matcher[key] === valueWithDefault;
+}
--- a/src/queries/a11yState.ts
+++ b/src/queries/a11yState.ts
@@ -10,25 +10,41 @@ import type {
   QueryByQuery,
 } from './makeQueries';
 
+/**
+ * Default accessibility state values based on experiments using accessibility
+ * inspector/screen reader on iOS and Android.
+ *
+ * @see https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State
+ */
+const defaultState: AccessibilityState = {
+  disabled: false,
+  selected: false,
+  checked: undefined,
+  busy: false,
+  expanded: undefined,
+};
+
 export function matchAccessibilityState(
   node: ReactTestInstance,
   matcher: AccessibilityState
 ) {
-  const stateProp = node.props.accessibilityState;
-
-  // busy, disabled & selected states default to false,
-  // while checked & expended states treat false and default as sepatate values
+  const state = node.props.accessibilityState;
   return (
-    matchState(stateProp?.busy ?? false, matcher.busy) &&
-    matchState(stateProp?.disabled ?? false, matcher.disabled) &&
-    matchState(stateProp?.selected ?? false, matcher.selected) &&
-    matchState(stateProp?.checked, matcher.checked) &&
-    matchState(stateProp?.expanded, matcher.expanded)
+    matchState(state, matcher, 'disabled') &&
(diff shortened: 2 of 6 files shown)


AugustinLF:
Might be worth reusing the implementation from #1161, no?

mdjastrzebski:
@AugustinLF Added exported `matchAccessibilityState` that might be shared with `*ByRole` implementation.

src/queries/a11yState.ts
```diff
@@ -10,6 +10,23 @@ import type {
...
 
+export function matchAccessibilityState(
+  node: ReactTestInstance,
+  matcher: AccessibilityState
+) {
+  const stateProp = node.props.accessibilityState;
+
+  // busy, disabled & selected states default to false,
+  // while checked & expended states treat false and default as sepatate values
+  return (
```
pierrezimmermannbam, line 21: I think that the default values are a bit implicit here (hence the need for a comment), wdyt about having an object for those default values ? 

```tsx

const DEFAULT_ACCESSIBILITY_VALUES: Record<
  keyof AccessibilityState,
  boolean | undefined
> = {
  busy: false,
  disabled: false,
  selected: false,
  checked: undefined,
  expanded: undefined,
};

export function matchAccessibilityState(
  node: ReactTestInstance,
  matcher: AccessibilityState
) {
  const stateProp = node.props.accessibilityState;
  // busy, disabled & selected states default to false,
  // while checked & expended states treat false and default as sepatate values
  return (
    matchState(
      stateProp?.busy ?? DEFAULT_ACCESSIBILITY_VALUES.busy,
      matcher.busy
    ) &&
    matchState(
      stateProp?.disabled ?? DEFAULT_ACCESSIBILITY_VALUES.disabled,
      matcher.disabled
    ) &&
    matchState(
      stateProp?.selected ?? DEFAULT_ACCESSIBILITY_VALUES.selected,
      matcher.selected
    ) &&
    matchState(
      stateProp?.checked ?? DEFAULT_ACCESSIBILITY_VALUES.checked,
      matcher.checked
    ) &&
    matchState(
      stateProp?.expanded ?? DEFAULT_ACCESSIBILITY_VALUES.expanded,
      matcher.expanded
    )
  );
}

```

Something like that for instance ? I also think that there should be a way to refactor the calls to matchState but I don't know how to map on Accessibility states properly with correct typing

src/queries/a11yState.ts
```diff
@@ -10,6 +10,23 @@ import type {
...
+    matchState(stateProp?.busy ?? false, matcher.busy) &&
+    matchState(stateProp?.disabled ?? false, matcher.disabled) &&
+    matchState(stateProp?.selected ?? false, matcher.selected) &&
+    matchState(stateProp?.checked, matcher.checked) &&
+    matchState(stateProp?.expanded, matcher.expanded)
+  );
+}
+
 function matchState(value: unknown, matcher: unknown) {
```
pierrezimmermannbam, line 30: If my understanding is correct , the following will always be true : 

```ts

  expect(view.queryAllByA11yState({ checked: undefined })).toBeTruthy();

```

Is this the intended behaviour ? If checked : undefined results in a state that's different from check: true or checked : false I guess this shouldn't work in any case, however i'm not sure this query is relevant

pierrezimmermannbam (changes requested):

src/queries/a11yState.ts
mdjastrzebski: `undefined` is similar to not having given entry when you get the prop from the object: `state.checked`. We could use `hasOwnProp` or something similar but not sure if its worth the effort.

In regular tests you want to check if given checkbox is checked or not checked (or mixed state). I do not see value in querying `undefined` (not set) state.

The goal of this PR is to make `*ByA11yState` queries have the same behavior as we add to `*ByRole` in #1134, where you query by role, name and state at the same time. From that point of view, you would make example queries:

```ts
getByRole('button', { name: "OK", disabled: false }) // <- you want to match implicitly enable buttons
getByRole('checkbox', { name: "Mr", checked: true/false }) // <- you want to examine checkbox state
getByRole('whatever', { name: "Details", expanded: true/false }) // <- you want to examine expandable section state
```

Which would translate to following `*ByA11yState` queries:

```ts
getByA11yState({ disabled: false }) // <- you want to match implicitly enable buttons
getByA11yState({ checked: true/false }) // <- you want to examine checkbox state
getByA11yState({ expanded: true/false }) // <- you want to examine expandable section state
```

IMO in both cases querying for `checked: undefined` or `expanded: undefined` is not useful for the user.

@pierrezimmermannbam do you think this is reasonable from users POV?

pierrezimmermannbam: Yes you're probably right this is probably not a valid use case and not worth supporting. An alternative could be to change matcher typing so that querying {{checked: undefined}} is not allowed. This would make sense if we consider this not relevant and don't support it and it would be less complicated to implement. I'm still not sure it's worth either though, current implementation is also fine to me

mdjastrzebski: Afaik using TS optional type: `{ checked?: boolean | 'mixed' }` is allowing both omitting the prop (`{}`) and using `{ checked: undefined }`. How could we allow omitting it, because you want to query other state props, without allowing passing explicit `undefined`?

mdjastrzebski:
@AugustinLF would you mind if I would use the exported `matchAccessibilityState` predicate from `*ByA11yState` in `*ByRole` to have common logic? Seems like `*ByA11yState` and `*ByRole` share the same behaviour in terms of a11y state matching.

mdjastrzebski:
@pierrezimmermannbam I've addressed your comment about default state object.

mdjastrzebski:
@AugustinLF I've used `matchAccessibilityState` from `*ByA11yState` to reduce duplicated state matching in `*ByRole` query.

AugustinLF:
> would you mind if I would use the exported matchAccessibilityState predicate from *ByA11yState in *ByRole to have common logic? Seems like *ByA11yState and *ByRole share the same behaviour in terms of a11y state matching.

@mdjastrzebski Yep good for me, makes sense

src/queries/role.ts
```diff
@@ -12,13 +12,22 @@ import type {
...
 } from './makeQueries';
+import { matchAccessibilityState } from './a11yState';
 
 type ByRoleOptions = {
   name?: TextMatch;
 } & AccessibilityState;
 
 type AccessibilityStateKey = keyof AccessibilityState;
 
+const accessibilityStateKeys: AccessibilityStateKey[] = [
```
AugustinLF, line 23: We could consider exporting that directly from `./a11yState`, or compute it from there

AugustinLF (approved):

AugustinLF:
Nice work!

src/queries/role.ts
mdjastrzebski: Ok, makes sense to move it, but I think that `.accessiblity.ts` would probably be a better place

src/helpers/matchers/accessibilityState.ts
```diff
@@ -0,0 +1,39 @@
...
+  busy: false,
+  expanded: undefined,
+};
+
+export function matchAccessibilityState(
+  node: ReactTestInstance,
+  matcher: AccessibilityState
+) {
+  const state = node.props.accessibilityState;
+  return (
```
pierrezimmermannbam, line 23: Maybe we could use accessibilityStateKeys to refactor this part, something like this maybe ?

```ts
accessibilityStateKeys.forEach(accessibilityStateKey => {
   if (!matchState(state, matcher, accessibilityStateKey) {
       return false;
} 

return true;
 }
```

mdjastrzebski: `accessibilityStateKeys.every(key => matchState(state, matcher, key))` probably. It should do short circuiting by itself without ifs and early returns.

src/queries/a11yState.ts
pierrezimmermannbam: I made a few tries but I couldn't find a way to make this work. Still I'd be surprised that user query for undefined accessibility state so I think we can stick to current behavior

pierrezimmermannbam (approved):
@mdjastrzebski I added just an additional suggestion of refactor, otherwise changes look good to me, well done !

Issue #886, opened 2021-12-29
RNTL accidentally triggers `jest.useFakeTimers`


Niryo:
## Describe the bug
I am using sinon-fake-timers together with jest, because jest's exposes only minimal part of sinon timers and I need more advanced capabilities. The thing is that React native testing library mistakenly thinks that I am using "jest.useFakeTimers", because of [this check](https://github.com/callstack/react-native-testing-library/blob/c42bae123d69375e27b94ff0329380a13e31e87b/src/helpers/timers.js#L40):  

```javascript
  if (
    typeof globalObj.setTimeout.clock !== 'undefined' &&
    // $FlowIgnore[prop-missing]
    typeof jest.getRealSystemTime !== 'undefined'
  )
```

The first statement is true because I am using sinon fake timers, and the second is true because I am using jest, so it assumes that I am using `jest.useFakeTimers`, and it automatically turn it on for me [here](https://github.com/callstack/react-native-testing-library/blob/main/src/helpers/timers.js#L17):

```javascript
  if (fakeTimersType) {
    jest.useFakeTimers(fakeTimersType);
  }
```

This results in both custom sinon timer and jest's fake timers being installed, and it causes issues in my tests. 

## Expected behavior

`jest.useFakeTimers` should not be called if the user didn't call it explicitly.

## Steps to Reproduce
use jest 27 and sinon-fake-timers

## Versions

  npmPackages:
    @testing-library/react-native: ^9.0.0 => 9.0.0 
    react-test-renderer: ^17.0.2 => 17.0.2

mdjastrzebski:
@Niryo can you check if the issue persists with current version of RNTL with Jest 29?

Niryo:
Yea, I'll check it later this week

Niryo:
@mdjastrzebski yep, it's still an issue

mdjastrzebski:
@Niryo thanks for checking it. BTW Jest recently (v27 or v28) added more features to its fake timers implementation. Does sinon-fake-timers still has some features that Jest does not?

Niryo:
@mdjastrzebski jest still exposes only a subset of sinon's options

Niryo:
I agree that it's a rare use case. what about introducing an env param? you already have this env param? you already have this `RNTL_SKIP_AUTO_CLEANUP`, I can add another one like  `RNTL_SKIP_SETTING_TIME_MOCKS`.

mdjastrzebski:
[...]

PR #1169, opened 2022-10-11
Ignore fake timers

Niryo:
### Summary

Allow explicitly disabling automatic jest fake timers detection in order to solve https://github.com/callstack/react-native-testing-library/issues/886

### DIFF
--- a/src/helpers/__tests__/timers.test.ts
+++ b/src/helpers/__tests__/timers.test.ts
@@ -1,7 +1,7 @@
 import { jestFakeTimersAreEnabled } from '../timers';
 describe('timers', () => {
-  it('should not mock timers if SKIP_AUTO_FAKE_TIMERS_DETECTION is set', async () => {
-    process.env.SKIP_AUTO_FAKE_TIMERS_DETECTION = 'true';
+  it('should not mock timers if RNTL_SKIP_AUTO_DETECT_FAKE_TIMERS is set', async () => {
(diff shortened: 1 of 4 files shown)


src/helpers/__tests__/timers.test.ts
```diff
@@ -0,0 +1,8 @@
+import { jestFakeTimersAreEnabled } from '../timers';
+describe('timers', () => {
+  it('should not mock timers if SKIP_AUTO_FAKE_TIMERS_DETECTION is set', async () => {
+    process.env.SKIP_AUTO_FAKE_TIMERS_DETECTION = 'true';
```
mdjastrzebski, line 4:
```suggestion
    process.env.RNTL_SKIP_AUTO_DETECT_FAKE_TIMERS = 'true';
```

Let's add `RNTL_` prefix as for `RNTL_SKIP_AUTO_CLEANUP`.

nit: I've also suggested `AUTO_DETECT_FAKE_TIMERS` vs `AUTO_FAKE_TIMER_DETECTION` as slightly more readable.

mdjastrzebski (commented):
Looks good!

@Niryo pls document this settings. This should probably be very brief, you can put it under [configuration section](https://callstack.github.io/react-native-testing-library/docs/api#configuration) in API.md file

src/helpers/__tests__/timers.test.ts
Niryo: done

mdjastrzebski (approved):

PR #1175, opened 2022-10-15
docs: add typedefs for queries

mdjastrzebski:
### Summary

Improve documentation by adding expanded type definitions for all queries. This should make our query docs easier to understand. In line with DTL docs, I provided only typedef for `getBy` variant.

### Test plan

None.

### DIFF
--- a/website/docs/Queries.md
+++ b/website/docs/Queries.md
@@ -21,6 +21,7 @@ title: Queries
   - [`ByLabelText`](#bylabeltext)
   - [`ByHintText`, `ByA11yHint`, `ByAccessibilityHint`](#byhinttext-bya11yhint-byaccessibilityhint)
   - [`ByRole`](#byrole)
+    - [Options](#options-1)
   - [`ByA11yState`, `ByAccessibilityState`](#bya11ystate-byaccessibilitystate)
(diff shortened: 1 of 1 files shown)


pierrezimmermannbam (approved):
Great improvement !

website/docs/Queries.md
```diff
@@ -90,6 +91,16 @@ Usually query first argument can be a **string** or a **regex**. Some queries ac
...
+  options?: {
+    exact?: boolean;
+    normalizer?: (text: string) => string;
+  }
+): ReactTestInstance;
+```
+
 Returns a `ReactTestInstance` with matching text – may be a string or regular expression.
 
 This method will join `<Text>` siblings to find matches, similarly to [how React Native handles these components](https://reactnative.dev/docs/text#containers). This will allow for querying for strings that will be visually rendered together, but may be semantically separate React components.
```
pierrezimmermannbam, line 106: This is not related to this pr but isn't this false now ? If I'm correct byText queries do not match text rendered in several components

mdjastrzebski: I think you are right, we should revisit our `*ByText` handling in a separate issue/PR, as there are some things worth discussing, e.g. should we match text spread between different child `Text` instance, compat with Jest Native, compat with RTL behaviour

PR #1177, opened 2022-10-16
fix: do not use a real timeout in waitfor when using fake timers

### LINKED ISSUE #1176: When using fake timers, waitFor still uses a setTimeout with real timers
## Describe the bug

waitFor timeout is using real time when using fake timers, so the fake timers are not necessarily advanced of the timeout. This bug is visible only if the time to perform the check is greater than the interval but it can happen for queries on large components (I met this bug when working on a real codebase). 

The following test will fail because the time needed to perform the check is greater than 5ms

```ts
test.each([false, true])(
  'it should not depend on real time when using fake timers (legacyFakeTimers = %s)',
  async () => {
    jest.useFakeTimers({ legacyFakeTimers: false });

    const mockErrorFn = jest.fn(() => {
      fibonaci(30);
      throw new Error('test');
    });

    try {
      await waitFor(() => mockErrorFn(), {
        timeout: 200,
        interval: 5,
      });
    } catch (error) {
      // suppress
    }

    expect(mockErrorFn).toHaveBeenCalledTimes(41);
  }
);

```

## Expected behavior

I think this is confusing, as we can't use waitFor in a stable way to advance fake timers and I'd expect than when waitFor fails, the timers have been advanced by its timeout. 
However I'm not sure if there is a reason why it was implemented like this, @mdjastrzebski any ideas ?

## Steps to Reproduce

The test above fails on the RNTL repo

pierrezimmermannbam:
### Summary

Fixes #1176 

Use a timeout in waitFor only when using realTimers and timeout with fake timers only after fakeTimeRemaining is over

### Test plan

I added a test case that initially failed. I compute fibonaci suite so that it takes some time to perform the checks (it must take longer than interval for this bug to be visible)

This shouldn't be a breaking change, all tests that previously passed should still do because we're not performing any check after fakeTimeRemaining is over anyway

### DIFF
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -236,9 +236,9 @@ test.each([true, false])(
         })
     ).rejects.toThrow();
 
-    // Verify that even though time to perform check is longer than interval
-    // test won't timeout until number of checks * interval >= timeout
-    // i.e. fake timers have been advanced (at least?) by timeout when waitFor rejects
+    // Verify that the `waitFor` callback has been called the expected number of times
+    // (timeout / interval + 1), so it confirms that the real duration of callback did not
+    // cause the real clock timeout when running using fake timers.
     expect(mockErrorFn).toHaveBeenCalledTimes(
       WAIT_FOR_TIMEOUT / WAIT_FOR_INTERVAL + 1
     );
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -204,21 +204,25 @@ test.each([false, true])(
   }
 );
 
-const fibonaci = (n: number): number => {
-  if (n === 0 || n === 1) {
-    return 1;
+const blockThread = (timeToBlockThread: number, legacyFakeTimers: boolean) => {
+  jest.useRealTimers();
+  let end = Date.now() + timeToBlockThread;
+
+  while (Date.now() < end) {
+    // do nothing
   }
 
-  return fibonaci(n - 1) + fibonaci(n - 2);
+  jest.useFakeTimers({ legacyFakeTimers });
 };
 
-test.each([false, true])(
+test.each([true, false])(
   'it should not depend on real time when using fake timers (legacyFakeTimers = %s)',
-  async () => {
-    jest.useFakeTimers({ legacyFakeTimers: false });
+  async (legacyFakeTimers) => {
+    jest.useFakeTimers({ legacyFakeTimers });
 
     const mockErrorFn = jest.fn(() => {
-      fibonaci(30);
+      // Wait 10 seconds so that check time is longer than interval
+      blockThread(10, legacyFakeTimers);
       throw new Error('test');
     });
 
@@ -227,10 +231,13 @@ test.each([false, true])(
         timeout: 200,
         interval: 5,
       });
-    } catch (error) {
-      // suppress
+    } catch (e) {
+      // do nothing
     }
 
+    // Verify that even though time to perform check is longer than interval
+    // test won't timeout until number of checks * interval >= timeout
+    // ie fake timers have been advanced by timeout when waitfor rejects
     expect(mockErrorFn).toHaveBeenCalledTimes(41);
   }
 );
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -221,15 +221,15 @@ test.each([true, false])(
     jest.useFakeTimers({ legacyFakeTimers });
 
     const mockErrorFn = jest.fn(() => {
-      // Wait 10 seconds so that check time is longer than interval
-      blockThread(10, legacyFakeTimers);
+      // Wait 5 seconds so that check time is longer than interval
+      blockThread(5, legacyFakeTimers);
       throw new Error('test');
     });
 
     try {
       await waitFor(() => mockErrorFn(), {
-        timeout: 200,
-        interval: 5,
+        timeout: 6,
+        interval: 2,
       });
     } catch (e) {
       // do nothing
@@ -238,7 +238,7 @@ test.each([true, false])(
     // Verify that even though time to perform check is longer than interval
     // test won't timeout until number of checks * interval >= timeout
     // ie fake timers have been advanced by timeout when waitfor rejects
-    expect(mockErrorFn).toHaveBeenCalledTimes(41);
+    expect(mockErrorFn).toHaveBeenCalledTimes(4);
   }
 );
 
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -221,7 +221,7 @@ test.each([true, false])(
     jest.useFakeTimers({ legacyFakeTimers });
 
     const mockErrorFn = jest.fn(() => {
-      // Wait 5 seconds so that check time is longer than interval
+      // Wait 5 ms so that check time is longer than interval
       blockThread(5, legacyFakeTimers);
       throw new Error('test');
     });
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -226,14 +226,13 @@ test.each([true, false])(
       throw new Error('test');
     });
 
-    try {
-      await waitFor(() => mockErrorFn(), {
-        timeout: 6,
-        interval: 2,
-      });
-    } catch (e) {
-      // do nothing
-    }
+    await expect(
+      async () =>
+        await waitFor(() => mockErrorFn(), {
+          timeout: 6,
+          interval: 2,
+        })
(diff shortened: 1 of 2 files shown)


src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,37 @@ test.each([false, true])(
   }
 );
 
+const fibonaci = (n: number): number => {
+  if (n === 0 || n === 1) {
+    return 1;
+  }
+
+  return fibonaci(n - 1) + fibonaci(n - 2);
+};
```
mdjastrzebski, line 213: Could we replace this with something both more controllable and explicit:
```suggestion
function blockThread(milliseconds)
{
    var end = Date.now() + milliseconds;
    while (Date.now() < end) {
      // Intentionally do nothing
    };
}
```

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,37 @@ test.each([false, true])(
...
+
+test.each([false, true])(
+  'it should not depend on real time when using fake timers (legacyFakeTimers = %s)',
+  async () => {
+    jest.useFakeTimers({ legacyFakeTimers: false });
```
mdjastrzebski, line 218: This test does not seem to use `each` parameter. Pls remove each param or apply it to `legacyFakeTimers`

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,37 @@ test.each([false, true])(
...
+      throw new Error('test');
+    });
+
+    try {
+      await waitFor(() => mockErrorFn(), {
+        timeout: 200,
+        interval: 5,
+      });
+    } catch (error) {
+      // suppress
```
mdjastrzebski, line 231: If error is expected then perhaps it's better to use `expect(...).toThrow()` as otherwise the current code works fine whether `waitFor` throws or not, which seems to be a bit fragile.

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,37 @@ test.each([false, true])(
...
+
+test.each([false, true])(
+  'it should not depend on real time when using fake timers (legacyFakeTimers = %s)',
+  async () => {
+    jest.useFakeTimers({ legacyFakeTimers: false });
+
+    const mockErrorFn = jest.fn(() => {
+      fibonaci(30);
+      throw new Error('test');
+    });
```
mdjastrzebski, line 223:
```suggestion
    const mockErrorFn = jest.fn(() => {
      blockThread(6); // 6 ms
      throw new Error('test');
    });
```

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,37 @@ test.each([false, true])(
...
+    try {
+      await waitFor(() => mockErrorFn(), {
+        timeout: 200,
+        interval: 5,
+      });
+    } catch (error) {
+      // suppress
+    }
+
+    expect(mockErrorFn).toHaveBeenCalledTimes(41);
```
mdjastrzebski, line 234: Not sure what is the intent of this check. Pls add some code comment to explain what is happening

mdjastrzebski (changes requested):
@pierrezimmermannbam I've reviewed the code the best I could, but I do not feel an expert in fake timers. 

The changes in the `waitFor.ts` file seem reasonable. Though I felt safer just sticking to the DTL code which we copied, but that might be just my poor understanding of fake timers ;-)

I've left some comments & suggestions about test code.

@thymikee pls take a look at this PR, you seem to know much more about fake timers than I.

src/__tests__/waitFor.test.tsx
pierrezimmermannbam: Very good idea, it's way more explicit that way! I just had to switch to real timers when calling blockThread so that it works

src/__tests__/waitFor.test.tsx
pierrezimmermannbam: Oh my bad, fixed it

src/__tests__/waitFor.test.tsx
pierrezimmermannbam: Agreed, I use expect now

pierrezimmermannbam:
@mdjastrzebski  thanks for the review ! I believe I implemented all the changes you requested

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,43 @@ test.each([false, true])(
...
+  jest.useFakeTimers({ legacyFakeTimers });
+};
+
+test.each([true, false])(
+  'it should not depend on real time when using fake timers (legacyFakeTimers = %s)',
+  async (legacyFakeTimers) => {
+    jest.useFakeTimers({ legacyFakeTimers });
+
+    const mockErrorFn = jest.fn(() => {
+      // Wait 5 seconds so that check time is longer than interval
```
mdjastrzebski, line 224: The seems to be ms from check the other code.
```suggestion
      // Wait 5 ms so that check time is longer than interval
```

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,43 @@ test.each([false, true])(
...
+      blockThread(5, legacyFakeTimers);
```
mdjastrzebski, line 225: I would wait a little bit longer to make the test more robust using the exact value might cause some race conditions
```suggestion
      blockThread(8, legacyFakeTimers);
```

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,43 @@ test.each([false, true])(
...
+      throw new Error('test');
+    });
+
+    await expect(
+      async () =>
+        await waitFor(() => mockErrorFn(), {
```
mdjastrzebski, line 231:
```suggestion
        await waitFor(mockErrorFn, {
```

mdjastrzebski (commented):
I think I start to understand this a bit more. I've left some comments for the testing code.

src/__tests__/waitFor.test.tsx
pierrezimmermannbam: Actually I'm not using the exact value, the aim is that we block the thread for a time longer than the interval, but I'll make it a bit more explicit

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,43 @@ test.each([false, true])(
...
+          timeout: 6,
+          interval: 2,
```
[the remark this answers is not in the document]
pierrezimmermannbam: I don't see how using tight intervals could be fragile since we're using fake timers but maybe I'm thinking too much based in the current implementation so I increased it to 20

pierrezimmermannbam:
@mdjastrzebski thanks again for your feedbacks, I treated all your comments

src/__tests__/waitFor.test.tsx
```diff
@@ -204,6 +204,47 @@ test.each([false, true])(
...
+          interval: WAIT_FOR_INTERVAL,
+        })
+    ).rejects.toThrow();
+
+    // Verify that even though time to perform check is longer than interval
+    // test won't timeout until number of checks * interval >= timeout
+    // i.e. fake timers have been advanced (at least?) by timeout when waitFor rejects
+    expect(mockErrorFn).toHaveBeenCalledTimes(
+      WAIT_FOR_TIMEOUT / WAIT_FOR_INTERVAL + 1
+    );
```
mdjastrzebski, line 244:
```suggestion
    // Verify that the `waitFor` callback has been called the expected number of times
    // (timeout / interval + 1), so it confirms that the real duration of callback did not cause
    // real clock timeout when running using fake timers.
    expect(mockErrorFn).toHaveBeenCalledTimes(
      WAIT_FOR_TIMEOUT / WAIT_FOR_INTERVAL + 1
    );
```

mdjastrzebski (approved):
Looks good, I think I finally grasped what is happening both in the code and in the test.

Good job @pierrezimmermannbam for catching it. I've suggested a rewrite to the expect comment to clarify the intent. Pls use/modify as you find it suitable.

mdjastrzebski:
@thymikee @AugustinLF pls take a look at this PR, as I do not feel expert on fake timers.

AugustinLF:
@mdjastrzebski it looks good to me, but same, I'm not familiar with the timers part of the codebase. I'd suggest releasing this commit alone in a dedicated rc release to be able to do some testing on it, it's the type of changes that could bring some weird errors in some codebases.

mdjastrzebski:
I've asked @thymikee for review, as our "fake timer specialist". That might delay merging for couple of days, but I think it's worth it.

thymikee (approved):
Makes sense, thanks!

mdjastrzebski:
@pierrezimmermannbam you might want to submit a similar bug/PR for DOM Testing Library, as they should have the same issue.

PR #1179, opened 2022-10-17
fix: update render flow type with new option unstable_validateStringsRenderedWithinText

pierrezimmermannbam:
### Summary

Adds new option unstable_validateStringsRenderedWithinText to renderOptions type in flow types

### Test plan

I don't know how to test flow types, is there a recommended way to do so ?

### DIFF
--- a/typings/index.flow.js
+++ b/typings/index.flow.js
@@ -346,6 +346,7 @@ declare module '@testing-library/react-native' {
   declare interface RenderOptions {
     wrapper?: React.ComponentType<any>;
     createNodeMock?: (element: React.Element<any>) => any;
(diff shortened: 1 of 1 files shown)


mdjastrzebski (approved):
👍🏻

mdjastrzebski:
We have `flow` check but it only checks is typedef is syntactically correct. There is no check that would verify whether the flow types match TS code.

AugustinLF:
Just move it to flow-typed and it's not our problem anymore 👀

AugustinLF:
Yeah I know, was just saying that for the next one ;)

Issue #1183, opened 2022-10-17
Feature: `*ByLabelText` to support `accessibilityLabelledBy`


mdjastrzebski:
## Describe the Feature
Currently when using `*ByLabelText` it only gets the label value from `accessibilityLabel` props. The goal of this request is to get it also from control referred by [`accessibilityLabelledBy`](https://reactnative.dev/docs/accessibility#accessibilitylabelledby-android) prop if present. This props refers to other view using its `nativeID` identifier.

## Possible Implementations
* expand current `*ByLabelText` predict to analyse both `accessibilityLabel` and text content of view referenced by `accessibilityLabelledBy`. 
* in the initial implementation we could assume that target view is `Text`, as most natural type for labels. Later we could use existing `getNodeByText` function to get its text.

pafry7:
Hey @mdjastrzebski, I would like to work on this one.

mdjastrzebski:
Go ahead @pafry7 !

AugustinLF:
> in the initial implementation we could assume that target view is Text, as most natural type for labels. Later we could use existing getNodeByText function to get its text.

Please do that in the same, I don't think it'd make things really more complicated, and it could lead to a confusing error if we don't.

PR #1191, opened 2022-10-20
feat: check accessibilityLabelledBy in *ByLabelText queries

pafry7:
### Summary
Adds support for [accessibilityLabelledBy](https://reactnative.dev/docs/accessibility#accessibilitylabelledby-android) in `*ByLabelText` queries as in #1183.

### DIFF
--- a/src/helpers/matchers/matchTextContent.ts
+++ b/src/helpers/matchers/matchTextContent.ts
@@ -6,6 +6,13 @@ import { filterNodeByType } from '../filterNodeByType';
 import { getTextContent } from '../getTextContent';
 import { getHostComponentNames } from '../host-component-names';
 
+/**
+ * Matches the given node's text content against string or regex matcher.
+ *
+ * @param node - Node which text content will be matched
+ * @param text - The string or regex to match.
+ * @returns - Whether the node's text content matches the given string or regex.
+ */
 export function matchTextContent(
   node: ReactTestInstance,
   text: TextMatch,
--- a/src/helpers/matchers/matchTextContent.ts
+++ b/src/helpers/matchers/matchTextContent.ts
@@ -1,10 +1,6 @@
-import { Text } from 'react-native';
 import type { ReactTestInstance } from 'react-test-renderer';
-import { getConfig } from '../../config';
 import { matches, TextMatch, TextMatchOptions } from '../../matches';
-import { filterNodeByType } from '../filterNodeByType';
 import { getTextContent } from '../getTextContent';
-import { getHostComponentNames } from '../host-component-names';
 
 /**
  * Matches the given node's text content against string or regex matcher.
@@ -18,13 +14,6 @@ export function matchTextContent(
   text: TextMatch,
   options: TextMatchOptions = {}
 ) {
-  const textType = getConfig().useBreakingChanges
-    ? getHostComponentNames().text
-    : Text;
-  if (!filterNodeByType(node, textType)) {
-    return false;
-  }
-
   const textContent = getTextContent(node);
   const { exact, normalizer } = options;
   return matches(text, textContent, normalizer, exact);
--- a/src/queries/__tests__/labelText.test.tsx
+++ b/src/queries/__tests__/labelText.test.tsx
@@ -170,6 +170,7 @@ test('getByLabelText supports accessibilityLabelledBy', async () => {
   const { getByLabelText, getByTestId } = render(
     <>
       <Text nativeID="label">Label for input</Text>
+      {/* @ts-expect-error: waiting for RN 0.71.2 to fix incorrectly omitted `accessibilityLabelledBy` typedef. */}
       <TextInput testID="textInput" accessibilityLabelledBy="label" />
     </>
   );
@@ -184,6 +185,7 @@ test('getByLabelText supports nested accessibilityLabelledBy', async () => {
       <View nativeID="label">
         <Text>Label for input</Text>
       </View>
+      {/* @ts-expect-error: waiting for RN 0.71.2 to fix incorrectly omitted `accessibilityLabelledBy` typedef. */}
       <TextInput testID="textInput" accessibilityLabelledBy="label" />
     </>
   );
--- a/src/queries/__tests__/labelText.test.tsx
+++ b/src/queries/__tests__/labelText.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { View, Text, TouchableOpacity, TextInput } from 'react-native';
+import { View, Text, TextInput, TouchableOpacity } from 'react-native';
 import { render } from '../..';
 
 const BUTTON_LABEL = 'cool button';
@@ -173,8 +173,9 @@ test('getByLabelText supports accessibilityLabelledBy', async () => {
       <TextInput testID="textInput" accessibilityLabelledBy="label" />
     </>
   );
-  expect(getByLabelText('Label for input')).toEqual(getByTestId('textInput'));
-  expect(getByLabelText(/input/)).toEqual(getByTestId('textInput'));
+
+  expect(getByLabelText('Label for input')).toBe(getByTestId('textInput'));
+  expect(getByLabelText(/input/)).toBe(getByTestId('textInput'));
 });
 
 test('getByLabelText supports nested accessibilityLabelledBy', async () => {
@@ -187,6 +188,6 @@ test('getByLabelText supports nested accessibilityLabelledBy', async () => {
     </>
   );
 
-  expect(getByLabelText('Label for input')).toEqual(getByTestId('textInput'));
-  expect(getByLabelText(/input/)).toEqual(getByTestId('textInput'));
+  expect(getByLabelText('Label for input')).toBe(getByTestId('textInput'));
+  expect(getByLabelText(/input/)).toBe(getByTestId('textInput'));
 });
--- a/src/queries/__tests__/labelText.test.tsx
+++ b/src/queries/__tests__/labelText.test.tsx
@@ -1,11 +1,13 @@
 import * as React from 'react';
-import { View, Text, TouchableOpacity } from 'react-native';
+import { View, Text, TouchableOpacity, TextInput } from 'react-native';
 import { render } from '../..';
 
 const BUTTON_LABEL = 'cool button';
 const BUTTON_HINT = 'click this button';
 const TEXT_LABEL = 'cool text';
 const TEXT_HINT = 'static text';
+const INPUT_LABEL = 'cool input';
+const INPUT_LABELLED_BY = 'formLabel';
 // Little hack to make all the methods happy with type
 const NO_MATCHES_TEXT: any = 'not-existent-element';
 
@@ -38,6 +40,15 @@ const Section = () => (
       Title
     </Typography>
     <Button>{TEXT_LABEL}</Button>
+  </>
+);
+
+const FormInput = () => (
+  <>
+    <View nativeID={INPUT_LABELLED_BY}>
+      <Text>{INPUT_LABEL}</Text>
+    </View>
+    <TextInput accessibilityLabelledBy={INPUT_LABELLED_BY} />
   </>
 );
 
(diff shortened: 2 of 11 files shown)


src/queries/text.ts
```diff
@@ -54,7 +54,7 @@ const getChildrenAsText = (children: React.ReactChild[]) => {
   return textContent;
 };
 
-const getNodeByText = (
+export const getNodeByText = (
```
pafry7, line 57: As it is used in two different queries, I thought about moving it to `helpers` folder.

src/queries/labelText.ts
```diff
@@ -11,14 +14,38 @@ import type {
...
   instance: ReactTestInstance
 ): ((text: TextMatch) => Array<ReactTestInstance>) =>
   function queryAllByLabelTextFn(text) {
     return instance.findAll(
       (node) =>
         typeof node.type === 'string' &&
-        matchStringProp(node.props.accessibilityLabel, text)
+        (matchStringProp(node.props.accessibilityLabel, text) ||
+          matchAccessibilityLabelledByText(
+            instance,
```
pafry7, line 45: I don't like passing `instance` object here, but I couldn't figure out a better way, as I have to iterate over the whole tree to find an element with particular `nativeID`.

src/queries/text.ts
mdjastrzebski: `helpers/matchers/text.ts` should be good

src/queries/labelText.ts
```diff
@@ -11,14 +14,38 @@ import type {
...
+
 const queryAllByLabelText = (
   instance: ReactTestInstance
 ): ((text: TextMatch) => Array<ReactTestInstance>) =>
   function queryAllByLabelTextFn(text) {
     return instance.findAll(
       (node) =>
         typeof node.type === 'string' &&
-        matchStringProp(node.props.accessibilityLabel, text)
+        (matchStringProp(node.props.accessibilityLabel, text) ||
```
mdjastrzebski, line 43: Pls extract `matchAccessibilityLabel` function to encapsulate the `||` logic so that the condition here is simplier and contains only `&&` operatos?

src/queries/labelText.ts
mdjastrzebski: I think it's fine as is. We have to find the matching element somehow, and that requires starting from the root.

website/docs/Queries.md
```diff
@@ -195,7 +195,7 @@ getByLabelText(
 ): ReactTestInstance;
 ```
 
-Returns a `ReactTestInstance` with matching `accessibilityLabel` prop.
+Returns a `ReactTestInstance` with matching `accessibilityLabel` or text content of view referenced by `accessibilityLabelledBy`.
```
mdjastrzebski, line 198:
```suggestion
Returns a `ReactTestInstance` with matching label:
* by matching `accessibilityLabel` prop
* or by matching text content of text view referenced by `accessibilityLabelledBy` prop
```

mdjastrzebski (changes requested):
Good work @pafry7. The code looks pretty solid. I've left some comments regarding code organisation and docs.

src/queries/labelText.ts
```diff
@@ -11,14 +14,38 @@ import type {
...
+) => {
+  if (!nativeID) {
+    return false;
+  }
+
+  return (
+    rootInstance.findAll(
+      (node) =>
+        filterNodeByType(node, Text) &&
+        matchStringProp(node.props.nativeID, nativeID) &&
```
AugustinLF, line 30: We should be able to match `<View nativeID="input"><Text>A label</Text></View>`, and I believe this implementation won't work.

Ideally we should do something like `rootInstance.findAll(node => matchStringProp(node.props.nativeID, nativeID)).findAll(node => filterNodeByType(node, Text) && getNodeByText(node, text))`.

And add the relevant test covering that.

src/queries/__tests__/labelText.test.tsx
```diff
@@ -46,6 +50,12 @@ test('getByLabelText, queryByLabelText, findByLabelText', async () => {
     <Section />
   );
 
+  expect(getByLabelText(INPUT_LABEL).props.accessibilityLabelledBy).toEqual(
```
AugustinLF, line 53: Would it be possible to get that in a different `test()` case? I know there was only one so far, but I prefer explicitly named `test` cases for the different edge cases we want to cover. Makes it easier to figure out the issue when a test starts failing.

AugustinLF (changes requested):

src/queries/text.ts
pafry7: Hey, I've moved it to `helpers/matchers/matchTextContent`. Is it okay?

src/queries/labelText.ts
pafry7: Hey, I've changed implementation and moved the test to a separate test case. Is it okay?

src/helpers/matchers/matchTextContent.ts
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import { Text } from 'react-native';
+import { ReactTestInstance } from 'react-test-renderer';
+import { matches, TextMatch } from '../../matches';
+import { TextMatchOptions } from '../../queries/text';
+import { filterNodeByType } from '../filterNodeByType';
+
+const getChildrenAsText = (children: React.ReactChild[]) => {
```
pafry7, line 8: `React.ReactChild` is deprecated. Do we want to change it right now?

mdjastrzebski:
I've added `hacktoberfest-accept` label to mark this PR as eligible for Hacktoberfest contribution. I plan to review it again in following days.

src/queries/__tests__/labelText.test.tsx
```diff
@@ -93,3 +104,13 @@ test('getAllByLabelText, queryAllByLabelText, findAllByLabelText', async () => {
...
+
+test('getByLabelText supports accessibilityLabelledBy', async () => {
+  const { getByLabelText, queryByLabelText } = render(<FormInput />);
+
+  expect(getByLabelText(INPUT_LABEL).props.accessibilityLabelledBy).toEqual(
+    INPUT_LABELLED_BY
+  );
+  const textInput = queryByLabelText(/input/g);
+  expect(textInput?.props.accessibilityLabelledBy).toEqual(INPUT_LABELLED_BY);
+});
```
mdjastrzebski, line 116: Some improvements to the test:
1. Test both direct and nested reference
2. Verify the matched element directly against `getByTestId` result

```suggestion
test('getByLabelText supports accessibilityLabelledBy', async () => {
  const { getByLabelText, queryByLabelText } = render(<>
    <Text nativeID={INPUT_LABELLED_BY}>{INPUT_LABEL}</Text>
    <TextInput testID="textInput" accessibilityLabelledBy={INPUT_LABELLED_BY} />
  </>);

  expect(getByLabelText(INPUT_LABEL)).toEqual(getByTestId('textInput'));
  expect(getByLabelText(/input/)).toEqual(getByTestId('textInput'));
});

test('getByLabelText supports nested accessibilityLabelledBy', async () => {
  const { getByLabelText, queryByLabelText } = render(<>
    <View nativeID={INPUT_LABELLED_BY}>
      <Text >{INPUT_LABEL}</Text>
    </View>
    <TextInput testID="textInput" accessibilityLabelledBy={INPUT_LABELLED_BY} />
  </>);

  expect(getByLabelText(INPUT_LABEL)).toEqual(getByTestId('textInput'));
  expect(getByLabelText(/input/)).toEqual(getByTestId('textInput'));
});
```

src/helpers/matchers/matchTextContent.ts
```diff
@@ -0,0 +1,61 @@
...
+  return textContent;
+};
+
+/**
+ * Matches the given string property again string or regex matcher.
+ *
+ * @param prop - The string prop to match.
+ * @param matcher - The string or regex to match.
+ * @returns - Whether the string prop matches the given string or regex.
+ */
```
mdjastrzebski, line 45:
```suggestion
/**
 * Matches the given node's text content against string or regex matcher.
 *
 * @param node - Node which text content will be matched
 * @param matcher - The string or regex to match.
 * @returns - Whether the node's text content matches the given string or regex.
 */
```

mdjastrzebski (changes requested):
@pafry7 Looks pretty solid. I've added some suggestions for improvements to code and tests.

src/queries/labelText.ts
```diff
@@ -17,12 +18,48 @@ type ByLabelTextOptions = CommonQueryOptions & TextMatchOptions;
...
+) => {
+  if (!nativeID) {
+    return false;
+  }
+
+  const { exact, normalizer } = options;
+
+  return rootInstance
+    .findAll((node) =>
+      matches(nativeID, node.props?.nativeID, normalizer, exact)
```
pafry7, line 41: `React.Fragment` doesn't have neither `nativeID` prop nor `testID` and tests started to fail. I've extended the type of `text` in `matches` to `string | undefined`, because we are checking it anyway:
```
 if (typeof text !== 'string') {
    return false;
  }
```
Adding `typeof node.type === 'string'` would fix it, but then we have to add 
```
const isTextComponent = filterNodeByType(node, Text) | filterNodeByType(node, "Text")
```
to `matchTextContent`. 
I didn't know what fits best here. @mdjastrzebski

mdjastrzebski:
Let's pause this waiting for #1222, as it has some overlapping changes, which might be useful in this PR.

mdjastrzebski:
Rebased to current `main`.

mdjastrzebski:
Rebased to `main`, cleaned up to account for recent code improvements in `main`.

@AugustinLF could you take another look at this PR?

mdjastrzebski (approved):
Looks good.

website/docs/Queries.md
```diff
@@ -212,7 +212,9 @@ getByLabelText(
 ): ReactTestInstance;
 ```
 
-Returns a `ReactTestInstance` with matching `accessibilityLabel` prop.
+Returns a `ReactTestInstance` with matching label:
+- either by matching [`accessibilityLabel`](https://reactnative.dev/docs/accessibility#accessibilitylabel) prop
+- or by matching text content of view referenced by [`accessibilityLabelledBy`](https://reactnative.dev/docs/accessibility#accessibilitylabelledby-android) prop
```
MattAgn, line 217: should we mention that this prop only works for android? it's also specified in the RN doc but I feel it's the first time an RNTL query would validate code working only on android so it might be worth pointing out

mdjastrzebski: This is a good question, whether RNTL should work like `ios`, `android`, common part of `iOS` and `android` or either `iOS` or `Android`. So far we seem to take the last approach when it comes to `isHiddenFromAccessibility`: element is hidden if any iOS or Android prop indicates that it is hidden.

src/queries/text.ts
```diff
@@ -19,7 +20,10 @@ import type {
   QueryByQuery,
 } from './makeQueries';
 import type { CommonQueryOptions } from './options';
-import { getHostComponentNames } from '../helpers/host-component-names';
+
+beforeEach(() => {
+  resetToDefaults();
+});
```
MattAgn, line 26: I don't understand these lines, this file is not a test file. Maybe an issue occurred during the rebase?

website/docs/Queries.md
MattAgn: true, did not notice it back when we implemented it. For isHiddenFromAccessibility, we could add a little note in [this page](https://callstack.github.io/react-native-testing-library/docs/api#ishiddenfromaccessibility)? Or is overkill?

MattAgn (approved):

website/docs/Queries.md
mdjastrzebski: It is something that would be good to analyse and discuss in the mid-future. We have various options here should figure out which one makes most sense from user POV. For now I would stick with the current approach of: "this works with either iOS or Android".

AugustinLF (approved):

PR #1195, opened 2022-10-21
docs: testing environment

mdjastrzebski:
### Summary

This docs describes in simple terms how our testing environment operates, so that our users can learn more on behaviour and limitations of using RNTL.

### Test plan

None


website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+- [Composite components](https://reactnative.dev/architecture/glossary#react-composite-components) are React code organisation units that exist only on the JavaScript side of your app. Typical examples are components you create (both function and class components), components imported from React Native (`View`, `Text`, etc) or from 3rd party packages.
+
+That might sound a bit confusing at first, since we put React Native’s `View` in both categories. There are actually two `View` components: composite one and host one. The relation between them is as follows:
+
+- composite `View` is the type imported from `react-native` package. It’s a JavaScript component, which renderers host `View` as its only child in the element tree.
```
AugustinLF, line 65:
```suggestion
- composite `View` is the type imported from `react-native` package. It’s a JavaScript component, which renders host `View` as its only child in the element tree.
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+  * <View accessible={true} {...}> (host)
+    * children prop passed in JSX
+```
+
+#### Differentiating between host and composite elements
+
+Any easy way to differentiate between host and composite elements is the `type` prop of `ReactTestInstance`:
+
+- for host components it’s always a string value representing component name, e.g. `"View"`
+- for composite components it’s function or class defining the component
```
AugustinLF, line 97:
```suggestion
- for composite components it’s a function or class corresponding to the component
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+    <Pressable>
+      <Text>{title}</Text>
+    </Pressable>
+  );
+}
+```
+
+In the above example user defined components accepts both `onPress` and `style` props but does not pass it (through `Pressable`) to host views, so these props will not affect the user interface.
+
+### Tree navigation
```
AugustinLF, line 127: Should we add a disclaimer? People should just not use that if they're not 100% they need to. And by that I don't just mean "be careful it's hard", but rather, "you're probably doing something you shouldn't be doing at all"

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+When navigating three using `parent` or `children` props of `ReactTestInstance` element, you will encounter both host and composite elements. You should be careful when navigating the element tree, as the tree structure for 3rd party components and change independently from your code and cause unexpected test failures.
```
AugustinLF, line 129:
```suggestion
When navigating a tree of react elements using `parent` or `children` props of a `ReactTestInstance` element, you will encounter both host and composite elements. You should be careful when navigating the element tree, as the tree structure for 3rd party components and change independently from your code and cause unexpected test failures.
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+### React renderers
+
+React allows you to write declarative code using JSX, write function or class components, or use hooks like `useState`. In order to output the results of your components it needs to work with a render. Every React app uses some type of renderer: React Native is a renderer for mobile apps, web apps use React DOM, and there are other more [specialised renderers](https://github.com/chentsulin/awesome-react-renderer) that can can e.g. render to console or HTML canvas.
+
+When you run your tests in React Native Testing Library, somewhat contrary to what the name suggest, they are actually **not** using React Native render. This is because this renderer needs to be run on iOS or Android operating system, so it would need to run on device or simulator.
+
+### React Test Render
```
AugustinLF, line 16:
```suggestion
### React Test Renderer
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+Instead, RNTL uses React Test Renderer which is a specialised renderer that allows rending to pure JavaScript objects without access to mobile OS, and can run in Node.js environment using Jest (or other JavaScript test runners).
```
AugustinLF, line 18:
```suggestion
Instead, RNTL uses React Test Renderer which is a specialised renderer that allows rendering to pure JavaScript objects without access to mobile OS, and that can run in a Node.js environment using Jest (or any other JavaScript test runner). 
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+React allows you to write declarative code using JSX, write function or class components, or use hooks like `useState`. In order to output the results of your components it needs to work with a render. Every React app uses some type of renderer: React Native is a renderer for mobile apps, web apps use React DOM, and there are other more [specialised renderers](https://github.com/chentsulin/awesome-react-renderer) that can can e.g. render to console or HTML canvas.
+
+When you run your tests in React Native Testing Library, somewhat contrary to what the name suggest, they are actually **not** using React Native render. This is because this renderer needs to be run on iOS or Android operating system, so it would need to run on device or simulator.
+
+### React Test Render
+
+Instead, RNTL uses React Test Renderer which is a specialised renderer that allows rending to pure JavaScript objects without access to mobile OS, and can run in Node.js environment using Jest (or other JavaScript test runners). 
+
+Using React Test Renders has both pros and cons.
```
AugustinLF, line 20:
```suggestion
Using React Test Renderer has pros and cons.
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+Benefits:
+
+- tests can run on most CIs (linux, etc) and do not require mobile device or emulator
```
AugustinLF, line 24:
```suggestion
- tests can run on most CIs (linux, etc) and do not require a mobile device or emulator
```

AugustinLF (changes requested):
Very clear and well written, nice!

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
+---
+id: testing-env
+title: Testing Environment
+---
+
+## Testing Environment
+
+React Native Testing Library allows you to write integration and component tests for your React Native app or library. While the JSX code used in tests closely resembles your React Native app, the things are not quite simple as they might appear. In this document we will describe the key elements of our testing environment and highlight things to be aware when writing more advanced tests or diagnosing issues.
```
pierrezimmermannbam, line 8:
```suggestion
React Native Testing Library allows you to write integration and component tests for your React Native app or library. While the JSX code used in tests closely resembles your React Native app, the things are not quite as simple as they might appear. In this document we will describe the key elements of our testing environment and highlight things to be aware of when writing more advanced tests or diagnosing issues.
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+### React renderers
+
+React allows you to write declarative code using JSX, write function or class components, or use hooks like `useState`. In order to output the results of your components it needs to work with a render. Every React app uses some type of renderer: React Native is a renderer for mobile apps, web apps use React DOM, and there are other more [specialised renderers](https://github.com/chentsulin/awesome-react-renderer) that can can e.g. render to console or HTML canvas.
```
pierrezimmermannbam, line 12:
```suggestion
React allows you to write declarative code using JSX, write function or class components, or use hooks like `useState`. In order to output the results of your components it needs to work with a renderer. Every React app uses some type of renderer: React Native is a renderer for mobile apps, web apps use React DOM, and there are other more [specialised renderers](https://github.com/chentsulin/awesome-react-renderer) that can e.g. render to console or HTML canvas.
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+When you run your tests in React Native Testing Library, somewhat contrary to what the name suggest, they are actually **not** using React Native render. This is because this renderer needs to be run on iOS or Android operating system, so it would need to run on device or simulator.
```
pierrezimmermannbam, line 14:
```suggestion
When you run your tests in React Native Testing Library, somewhat contrary to what the name suggest, they are actually **not** using React Native renderer. This is because this renderer needs to be run on iOS or Android operating system, so it would need to run on device or simulator.
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+- Runtime behaviours are simulated, sometimes imperfectly
+
+It’s worth noting that React Testing Library (web one), works a bit different. While RTL also runs in Jest, it also has access to simulated browser DOM environment from jsdom package, so it can use a regular React DOM renderer. Unfortunately, there is no similar React Native runtime environment package. This is probably due to to the fact that while browser environment is well defined and highly standardised, the React Native environment is in constant evolution, in sync with the evolution of underlying OS-es. Maintaining such environment would require duplicating countless React Native behaviours, and keeping that in sync as React Native evolves.
+
+### Element tree
+
+Invoking `render()` functions results in creation of element tree. This is done internally by invoking `TestRendere.create()` method. The output tree represents your React Native component tree, each node of that tree is an “instance” of some React component (to be more precise: each node represents a React fiber, and only class components have instances, while function components store the hook state using fiber).
```
pierrezimmermannbam, line 39:
```suggestion
Invoking `render()` function results in creation of an element tree. This is done internally by invoking `TestRenderer.create()` method. The output tree represents your React Native component tree, each node of that tree is an “instance” of some React component (to be more precise: each node represents a React fiber, and only class components have instances, while function components store the hook state using fiber).
```

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+Most of the Testing Library queries return host components, in order to encourage best practices described above.
+
+At this stage, there are some noteworthy exceptions:
+
+- `*ByText` queries returns composite `Text` element
+- `*ByDisplayValue` queries returns composite `TextInput` element
+- `*ByPlaceholderText` queries returns composite `TextInput` element
+
+This will change in the near future, as we make efforts for all queries to return host components. Meanwhile it should be a huge issue, as composite `Text` and `TextInput`generally pass their props down to host counterparts.
```
pierrezimmermannbam, line 158:
```suggestion
This will change in the near future, as we make efforts for all queries to return host components. Meanwhile it shouldn't be a huge issue, as composite `Text` and `TextInput`generally pass their props down to host counterparts.
```

website/docs/TestingEnvironment.md
pierrezimmermannbam, line 129: I find it really interesting to explain more what's going on under the hood but the purpose to me is to help people to better use RNTL and to debug more efficiently. I'm not sure about recommending navigating the tree to access host elements, I think good practices should be enforced by the library instead

website/docs/TestingEnvironment.md
pierrezimmermannbam: I totally agree, I find it really interesting to explain more what's going on under the hood but the purpose to me is to help people to better use RNTL and to debug more efficiently. I'm not sure about recommending navigating the tree to access host elements, I think good practices should be enforced by the library instead

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+function ForgotToPassPropsButton({ title, onPress, style }) {
+  return (
+    <Pressable>
+      <Text>{title}</Text>
+    </Pressable>
+  );
+}
+```
+
+In the above example user defined components accepts both `onPress` and `style` props but does not pass it (through `Pressable`) to host views, so these props will not affect the user interface.
```
pierrezimmermannbam, line 125: I would have added here that the prop names aren't always the same, for instance onClick vs onPress or disabled vs accessibilityState : disabled

pierrezimmermannbam (changes requested):
Really interesting doc, I think it's a great addition ! I'm just not sure the tree navigation part is relevant, users should be able to use only RNTL api or @testing-library/jest-native matchers to write tests, even though understanding how it works is definitely useful

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Tree navigation
+
+When navigating three using `parent` or `children` props of `ReactTestInstance` element, you will encounter both host and composite elements. You should be careful when navigating the element tree, as the tree structure for 3rd party components and change independently from your code and cause unexpected test failures.
+
+If you want to find a host element for given element they you might use following code:
+
+```jsx
+function getHostParent(element: ReactTestInstance) {
```
thymikee, line 134: I think we don't ever want this kind of code to end up in the user space, do we?

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
+---
+id: testing-env
+title: Testing Environment
+---
+
+## Testing Environment
```
thymikee, line 6: I read through all of that and it's great deep dive into how things work under the hood! However, this is definitely for more advanced users, or when someone needs to debug the problems in this library or their tests. This gives me a hint, that we should be very explicit about it. And either add a note, like React docs does in many places, or make it, idk, a contributing or advanced section :)

website/docs/TestingEnvironment.md
```diff
@@ -0,0 +1,160 @@
...
+
+It’s worth noting that React Testing Library (web one), works a bit different. While RTL also runs in Jest, it also has access to simulated browser DOM environment from jsdom package, so it can use a regular React DOM renderer. Unfortunately, there is no similar React Native runtime environment package. This is probably due to to the fact that while browser environment is well defined and highly standardised, the React Native environment is in constant evolution, in sync with the evolution of underlying OS-es. Maintaining such environment would require duplicating countless React Native behaviours, and keeping that in sync as React Native evolves.
+
+### Element tree
+
+Invoking `render()` functions results in creation of element tree. This is done internally by invoking `TestRendere.create()` method. The output tree represents your React Native component tree, each node of that tree is an “instance” of some React component (to be more precise: each node represents a React fiber, and only class components have instances, while function components store the hook state using fiber).
+
+These tree elements are represented by `ReactTestInstance` type:
+
+```jsx
```
thymikee, line 43:
```suggestion
```tsx
```

website/docs/TestingEnvironment.md
mdjastrzebski: Removed that code.

mdjastrzebski:
@thymikee @AugustinLF @pierrezimmermannbam I've moved document to new `Advanced` sidebar section, added a note indicating that this is for more advanced users, and reworked the part on tree navigation.

AugustinLF:
@mdjastrzebski did you push your new commit?^^' I don't understand where this `Advanced` sidebar section is.

mdjastrzebski:
@AugustinLF LOL, nope ;-)

thymikee (approved):

AugustinLF (approved):

PR #1203, opened 2022-10-28
chore: add codecov

### LINKED ISSUE #1186: feature: setup code coverage reporting
## Describe the Feature
Setup code coverage reporting so that we are aware how well are out code paths covered. 

## Possible Implementations
Popular solution seems to be: https://about.codecov.io/ we could use something else if there are compelling reasons to do so.

## Related Issues
N/A

jaworek:
Setup codecov. Closes #1186.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,11 @@
+coverage:
+  precision: 2
+  round: down
+  range: 70...100
+
+  status:
+    project: true
+    patch: true
+    changes: false
+
+comment: false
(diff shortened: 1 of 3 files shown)


.codecov.yml
```diff
@@ -0,0 +1,4 @@
+coverage:
+  precision: 2
+  round: down
+  range: 70...100
```
mdjastrzebski, line 4:
```suggestion
  range: 70...100
  
```

mdjastrzebski (approved):
Looks great! Thank you for contributing this @jaworek !

PR #1210, opened 2022-11-03
feat: *ByRole a11y value option

mdjastrzebski:
### Summary

1. Add `value` option to `*ByRole` query
2. Improve `*ByA11yValue` query by allowing `text` entry to be string or regex

### Test plan

Added relevant tests for `*ByRole` query and `*ByA11yValue`.

### DIFF
--- a/src/helpers/matchers/accessibilityValue.ts
+++ b/src/helpers/matchers/accessibilityValue.ts
@@ -14,7 +14,7 @@ export function matchAccessibilityValue(
   node: ReactTestInstance,
   matcher: AccessibilityValueMatcher
 ): boolean {
-  const value = (node.props.accessibilityValue ?? {}) as AccessibilityValue;
+  const value: AccessibilityValue = node.props.accessibilityValue ?? {};
   return (
     (matcher.min === undefined || matcher.min === value.min) &&
     (matcher.max === undefined || matcher.max === value.max) &&
--- a/src/helpers/matchers/accessibilityValue.ts
+++ b/src/helpers/matchers/accessibilityValue.ts
@@ -1,0 +1,24 @@
+import { AccessibilityValue } from 'react-native';
+import { ReactTestInstance } from 'react-test-renderer';
+import { TextMatch } from '../../matches';
+import { matchStringProp } from './matchStringProp';
+
+export interface AccessibilityValueMatcher {
+  min?: number;
+  max?: number;
+  now?: number;
+  text?: TextMatch;
+}
+
+export function matchAccessibilityValue(
+  node: ReactTestInstance,
+  matcher: AccessibilityValueMatcher
+): boolean {
+  const value = (node.props.accessibilityValue ?? {}) as AccessibilityValue;
+  return (
+    (matcher.min === undefined || matcher.min === value.min) &&
+    (matcher.max === undefined || matcher.max === value.max) &&
+    (matcher.now === undefined || matcher.now === value.now) &&
+    (matcher.text === undefined || matchStringProp(value.text, matcher.text))
+  );
+}
--- a/src/queries/__tests__/a11yValue.test.tsx
+++ b/src/queries/__tests__/a11yValue.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { TouchableOpacity, Text } from 'react-native';
+import { View, Text, TouchableOpacity } from 'react-native';
 import { render } from '../..';
 
 const TEXT_LABEL = 'cool text';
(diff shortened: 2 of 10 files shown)


website/docs/Queries.md
```diff
@@ -357,21 +371,24 @@ getByA11yValue(
...
   options?: {
     hidden?: boolean;
   }
 ): ReactTestInstance;
 ```
 
-Returns a `ReactTestInstance` with matching `accessibilityValue` prop.
+Returns a host element with matching `accessibilityValue` prop entries. Only entires provided to the query will be used to match elements. Element might have additional accessibility value entries and still be matched.
+
+When querying by `text` entry a stirng or regex might be used.
```
pierrezimmermannbam, line 384:
```suggestion
When querying by `text` entry a string or regex might be used.
```

src/queries/a11yValue.ts
```diff
@@ -12,33 +16,38 @@ import type {
...
+        typeof node.type === 'string' && matchAccessibilityValue(node, value),
       queryOptions
     );
   };
 
-const getMultipleError = (value: A11yValue) =>
-  `Found multiple elements with accessibilityValue: ${JSON.stringify(value)} `;
-const getMissingError = (value: A11yValue) =>
-  `Unable to find an element with accessibilityValue: ${JSON.stringify(value)}`;
+const buildErrorMessage = (value: AccessibilityValueMatcher) => {
```
pierrezimmermannbam, line 34: I didn't understand this function at first, I think the naming error is a bit deceiving since it's not really errors at this point but rather the list of all accessibility values that are supposed to match. What about renaming this to something like formatAccessibilityValueMatcherToString ?

src/queries/a11yValue.ts
```diff
@@ -12,33 +16,38 @@ import type {
...
+  const errors: string[] = [];
+
+  accessiblityValueKeys.forEach((valueKey) => {
+    if (value[valueKey] !== undefined) {
+      errors.push(`${valueKey} value: ${value[valueKey]}`);
+    }
+  });
+
+  return errors.join(', ');
```
pierrezimmermannbam, line 43: I would also rename this to accessibilityValueMatcherEntries

pierrezimmermannbam (commented):
Looks very good ! I just added a comment on some naming that I find a bit confusing and a smaill typo in docs

pierrezimmermannbam (approved):

src/helpers/matchers/accessibilityValue.ts
```diff
@@ -0,0 +1,24 @@
...
+  max?: number;
+  now?: number;
+  text?: TextMatch;
+}
+
+export function matchAccessibilityValue(
+  node: ReactTestInstance,
+  matcher: AccessibilityValueMatcher
+): boolean {
+  const value = (node.props.accessibilityValue ?? {}) as AccessibilityValue;
```
AugustinLF, line 17: Does `const value: AccessibilityValue = node.props.accessibilityValue ?? {}` typecheck? That's usually safer, `as` can lead to bugs

src/queries/__tests__/a11yValue.test.tsx
```diff
@@ -46,15 +38,15 @@ test('getByA11yValue, queryByA11yValue, findByA11yValue', async () => {
   });
 
   expect(() => getByA11yValue({ min: 50 })).toThrow(
-    getNoInstancesFoundMessage('{"min":50}')
+    'Unable to find an element with min value: 50'
```
AugustinLF, line 41: Would you consider adding a test to see the error with both a min and max?

src/queries/__tests__/role.test.tsx
```diff
@@ -696,6 +696,24 @@ describe('error messages', () => {
       `"Unable to find an element with role: "button", disabled state: true"`
     );
   });
+
```
AugustinLF, line 699: Can we have a test with both accessibilityState and accessibilityValue?

src/helpers/matchers/accessibilityValue.ts
mdjastrzebski: `node.props.accessibilityValue` is `any` as `ReactTestInstances` props are not typed.

src/queries/__tests__/role.test.tsx
mdjastrzebski: I've added some more tests in this area.

src/queries/__tests__/a11yValue.test.tsx
mdjastrzebski: I've added some more tests in this area.

src/helpers/matchers/accessibilityValue.ts
AugustinLF: Then it should still pass, so I'd suggest to use this pattern, it's better to have this habit :)

mdjastrzebski: done

AugustinLF (approved):

PR #1220, opened 2022-11-10
Add alias to hidden in query options

### LINKED ISSUE #1194: RFC: readability vs RTL compatibility in API naming
RNTL is modelled after RTL, as one of our unstated goals we strive to reimplement RTL for React Native platform, expose familiar features and APIs while accounting for platform differences. 

So far we have been reasonably successful with this approach as it delivered familiar and well thought out API. However recently, when we started to implement A11y-related features I’ve found that some of them are somewhat confusingly named:

* `isInaccessible` - which checks if the element is hidden from accessibility, which is a different concept that being accessibility element as indicated by `accessible` prop. Values of `accessibility` prop and results of `isInaccessible` are independent and all four combinations (true/false) are allowed.
* `hidden` query option - this option controls whether we include elements hidden from accessibility in our queries. However, it is unintuitive whether `hidden: true` or `hidden: false` includes the hidden elements and I have to refer to the documentation to check that.

I think readability of our code could be improved by renaming these in a following way:
* `isInaccessible` -> `isHiddenFromAccessibility`
* `hidden` option -> `includeHidden`

Note: the change is only about the naming, the arguments or behaviour do not change.

Such renaming would however break our naming compatibility with RTL.

I see couple of options how we could approach that:
1. Keep RTL naming
2. Rename them
3. Expose both names: promote the readable name as primary, while allowing RTL-compatible name as a secondary option.

I think we should consider using 3rd option, as it gives us both more explicit and readable code while catering for people with RTL background.

@thymikee @AugustinLF @pierrezimmermannbam @MattAgn What do you think?

MattAgn:
This PR aims to tackle [this issue](https://github.com/callstack/react-native-testing-library/issues/1194). Basically the idea is to add a new option `includeHidden` that will be an alias to the query option `hidden` but much clearer. We keep the old name `hidden` for compatibility with RTL

### Summary

- [x] Alias `hidden` query option to `hidden`
- [x] Alias `defaultHidden` config to `defaultIncludeHidden`
- [x] add new alias to docs

### Test plan

- [x] Check new `includeHidden` option is working for all queries
- [x] Check new `defaultIncludeHidden` option is working

### Remaining work

- [x] Choose a naming between `includeHidden` and `includeHiddenFromAccessibility`

Resolves #1194

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,13 @@ export type Config = {
   /** Default timeout, in ms, for `waitFor` and `findBy*` queries. */
   asyncUtilTimeout: number;
 
-  /** Default hidden value for all queries */
+  /** Default includeHidden value for all queries */
+  defaultIncludeHidden: boolean;
+
+  /** Default hidden value for all queries, alias to defaultIncludeHidden to respect react-testing-library API
+   * WARNING: if both defaultHidden and defaultIncludeHidden values are defined,
+   * then defaultIncludeHidden will take precedence
+   */
   defaultHidden: boolean;
 
   /** Default options for `debug` helper. */
@@ -14,6 +20,7 @@ export type Config = {
 const defaultConfig: Config = {
   asyncUtilTimeout: 1000,
   defaultHidden: true,
+  defaultIncludeHidden: true,
 };
 
 let config = { ...defaultConfig };
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,12 +4,12 @@ export type Config = {
   /** Default timeout, in ms, for `waitFor` and `findBy*` queries. */
   asyncUtilTimeout: number;
 
-  /** Default includeHidden value for all queries */
-  defaultIncludeHidden: boolean;
+  /** Default includeHiddenElements value for all queries */
+  defaultIncludeHiddenElements: boolean;
 
-  /** Default hidden value for all queries, alias to defaultIncludeHidden to respect react-testing-library API
-   * WARNING: if both defaultHidden and defaultIncludeHidden values are defined,
-   * then defaultIncludeHidden will take precedence
+  /** Default hidden value for all queries, alias to defaultIncludeHiddenElements to respect react-testing-library API
+   * WARNING: if both defaultHidden and defaultIncludeHiddenElements values are defined,
+   * then defaultIncludeHiddenElements will take precedence
    */
   defaultHidden: boolean;
 
@@ -20,7 +20,7 @@ export type Config = {
 const defaultConfig: Config = {
   asyncUtilTimeout: 1000,
   defaultHidden: true,
-  defaultIncludeHidden: true,
+  defaultIncludeHiddenElements: true,
 };
 
 let config = { ...defaultConfig };
--- a/src/config.ts
+++ b/src/config.ts
@@ -4,7 +4,7 @@ export type Config = {
   /** Default timeout, in ms, for `waitFor` and `findBy*` queries. */
   asyncUtilTimeout: number;
 
-  /** Default includeHiddenElements value for all queries */
+  /** Default value for `includeHiddenElements` query option. */
   defaultIncludeHiddenElements: boolean;
 
   /** Default options for `debug` helper. */
@@ -12,10 +12,7 @@ export type Config = {
 };
 
 export type ConfigAliasOptions = {
-  /** Default hidden value for all queries, alias to defaultIncludeHiddenElements to respect react-testing-library API
-   * WARNING: if both defaultHidden and defaultIncludeHiddenElements values are defined,
-   * then defaultIncludeHiddenElements will take precedence
-   */
+  /** RTL-compatibility alias to `defaultIncludeHiddenElements` */
   defaultHidden: boolean;
 };
 
@@ -34,14 +31,10 @@ export function configure(options: Partial<Config & ConfigAliasOptions>) {
     defaultHidden ??
     config.defaultIncludeHiddenElements;
 
-  const optionsToSet = {
-    ...restOptions,
-    defaultIncludeHiddenElements,
-  };
-
   config = {
     ...config,
-    ...optionsToSet,
+    ...restOptions,
+    defaultIncludeHiddenElements,
   };
 }
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -7,42 +7,41 @@ export type Config = {
   /** Default includeHiddenElements value for all queries */
   defaultIncludeHiddenElements: boolean;
 
+  /** Default options for `debug` helper. */
+  defaultDebugOptions?: Partial<DebugOptions>;
+};
+
+export type ConfigAliasOptions = {
   /** Default hidden value for all queries, alias to defaultIncludeHiddenElements to respect react-testing-library API
    * WARNING: if both defaultHidden and defaultIncludeHiddenElements values are defined,
    * then defaultIncludeHiddenElements will take precedence
    */
   defaultHidden: boolean;
-
-  /** Default options for `debug` helper. */
-  defaultDebugOptions?: Partial<DebugOptions>;
 };
 
 const defaultConfig: Config = {
   asyncUtilTimeout: 1000,
-  defaultHidden: true,
   defaultIncludeHiddenElements: true,
 };
 
 let config = { ...defaultConfig };
 
-export function configure(options: Partial<Config>) {
-  if (
-    options.defaultHidden !== undefined &&
-    options.defaultIncludeHiddenElements === undefined
-  ) {
-    options.defaultIncludeHiddenElements = options.defaultHidden;
-  }
+export function configure(options: Partial<Config & ConfigAliasOptions>) {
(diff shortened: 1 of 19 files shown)


MattAgn:
@mdjastrzebski I've got some doubts about my jsdocs for the `hidden` alias, is it clear?
I was also wondering how to deal with the fact that we have two options with the same name. Ideally, we'd want to have both options in the api but internally to handle only one. However, it seems much more simple to handle everything in the `findAll` (both for the global config and the queries). What are your thoughts on that?

src/helpers/findAll.ts
```diff
@@ -12,8 +13,11 @@ export function findAll(
   options?: FindAllOptions
 ) {
   const results = root.findAll(predicate);
+  const includeHiddenQueryOption = options?.includeHidden ?? options?.hidden;
+  const defaultIncludeHidden =
+    getConfig()?.defaultIncludeHidden ?? getConfig()?.defaultHidden;
 
-  const hidden = options?.hidden ?? getConfig().defaultHidden;
+  const hidden = includeHiddenQueryOption ?? defaultIncludeHidden;
```
mdjastrzebski, line 20:
```
const includeHidden = 
  options?.includeHidden ??
  options?.hidden ??
  getConfig()?.defaultIncludeHidden ??
  getConfig()?.defaultHidden
 }

```

MattAgn: i wanted to avoid nested operators to maximise readibility, do you find it more readable your way?

MattAgn:
@mdjastrzebski I pushed a new commit with the other naming `includeHiddenFromAccessibility` so we can compare
Which do you prefer?

mdjastrzebski:
@MattAgn `includeHiddenElements` seems like a reasonable option. I would give it a go an see how the resulting code looks.

src/helpers/findAll.ts
```diff
@@ -12,8 +13,13 @@ export function findAll(
   options?: FindAllOptions
 ) {
   const results = root.findAll(predicate);
+  const includeHiddenElementsQueryOption =
```
mdjastrzebski, line 16: The following should be easier to read and give the same result
```ts
const includeHiddenElements = 
  options?.includeHiddenElements ??
  options?.hidden ??
  getConfig()?.defaultIncludeHiddenElements ??
  getConfig()?.defaultHidden;
```

typings/index.flow.js
```diff
@@ -8,7 +8,10 @@ type QueryAllReturn = Array<ReactTestInstance> | [];
 type FindReturn = Promise<ReactTestInstance>;
 type FindAllReturn = Promise<ReactTestInstance[]>;
 
-type CommonQueryOptions = { hidden?: boolean };
+type CommonQueryOptions = {
+  hidden?: boolean,
+  includeHiddenElements?: boolean,
```
mdjastrzebski, line 13:
```suggestion
  includeHiddenElements?: boolean,
  
  /** Alias to `includeHiddenElements` for RTL compatibility */
  hidden?: boolean,

```

typings/index.flow.js
```diff
@@ -463,6 +466,8 @@ declare module '@testing-library/react-native' {
 
   declare interface Config {
     asyncUtilTimeout: number;
+    defaultHidden: boolean;
+    defaultIncludeHiddenElements: boolean;
```
mdjastrzebski, line 470:
```suggestion
    defaultIncludeHiddenElements: boolean;
    defaultHidden: boolean;
```

website/docs/API.md
```diff
@@ -791,9 +792,13 @@ function configure(options: Partial<Config>) {}
 
 Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementToBeRemoved`) and `findBy*` queries. Defaults to 1000 ms.
 
+#### `defaultIncludeHiddenElements` option
+
+Default [includeHiddenElements](Queries.md#includehidden-option) query option for all queries. This default option will be overridden by the one you specify directly when using your query.
+
 #### `defaultHidden` option
```
mdjastrzebski, line 799: I would rather not give it a full header but rather put a mention in `defaultIncludeHiddenElements` entry that there is an alias for RTL compatibility.

website/docs/Queries.md
```diff
@@ -389,18 +390,22 @@ An element is considered to be hidden from accessibility based on [`isHiddenFrom
...
+expect(screen.getByText('I am hidden from accessibility')).toBeTruthy(); // Defaults to includeHiddenElements: true for now
 expect(
-  screen.getByText('I am hidden from accessibility', { hidden: true })
+  screen.getByText('I am hidden from accessibility', { includeHiddenElements: true })
 ).toBeTruthy();
 ```
 
+### `hidden` option
+
+This is just an alias to the `includeHiddenElements` option. It only exists to match react-testing-library naming used in [byRole](https://testing-library.com/docs/queries/byrole/#hidden). Prefer the use of `includeHiddenElements` if possible as `includeHiddenElements: true` is clearer than `hidden: true`.
```
mdjastrzebski, line 407: Similar to `configure` comment, I would rather not give it a full option header but only mention it under   `includeHiddenElements` option docs.

mdjastrzebski (commented):
I like the new name `includeHiddenElements` seems to strike a proper balance between being readable and concise. As they say "naming is hard" but I think you nailed it 👍🏻 

Regarding usage I would make sure we emphasis the readable variant, i.e. put it first in relevant tests, examples, docs, etc.

At this stage the goal for `hidden` alias is to help folks migrating from RTL or building dual RNTL/RTL wrappers. So it should work but should not be prominent at all not to confuse the main audience.

src/helpers/findAll.ts
mdjastrzebski: Yes, I find it much more readable, as a single nullish coalescing operator. Having it as separate intermediate vars would be useful if it used a mix of two operators (like `&&` and `||`) with different precedence, etc. In case all vars use the same operator, I found putting it together removes the question why are there some intermediate variables :-)

src/queries/__tests__/a11yState.test.tsx
```diff
@@ -239,9 +239,18 @@ test('byA11yState queries support hidden option', () => {
 
   expect(getByA11yState({ expanded: false })).toBeTruthy();
   expect(getByA11yState({ expanded: false }, { hidden: true })).toBeTruthy();
+  expect(
```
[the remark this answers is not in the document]
mdjastrzebski: Or in case you find it useful the current way, just put the `includeHiddenElements` first as the recommended option.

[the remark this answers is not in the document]
MattAgn: extracting them in another test file sounds good :) (helped me find a bug in the process)

MattAgn:
@mdjastrzebski thanks for the review, should be all good now :) 

I've fixed a bug I introduced as well: 
- since we have default values for `defaultIncludeHiddenElements: true`, 
- `configure({ defaultHidden: false })` would get us a config with `{ defaultHidden: false, defaultIncludeHiddenElements: true }``
- in findAll, we would then have  `includeHiddenElements = true` since `getConfig().defaultIncludeHiddenElements`would never be undefined 

That's why I added some code in the configure option to tackle that

src/config.ts
```diff
@@ -14,11 +20,26 @@ export type Config = {
 const defaultConfig: Config = {
   asyncUtilTimeout: 1000,
   defaultHidden: true,
+  defaultIncludeHiddenElements: true,
 };
 
 let config = { ...defaultConfig };
 
 export function configure(options: Partial<Config>) {
+  if (
```
mdjastrzebski, line 29: I wonder if we should rather than syncing these two keys have a solution that would set `defaultIncludeHiddenElements` when user passes `defaultHidden`:

```ts
const { defaultHidden, ...restOptions } = options;
const optionsToSet {
  ...restOptions,
  defaultIncludeHiddenElements: restOptions.defaultIncludeHiddenElements ?? defaultHidden ?? config.defaultIncludeHiddenElements
}
```

That way we would also simplify the consuming code. 

Regarding typing this could be handled as:
```ts
interface Config {
  asyncUtilTimeout: number;
  defaultIncludeHiddenElements: boolean;
  // no defaultHidden
}
let config: Config

interface ConfigAliasOptions = {
  defaultHidden?: boolean;
}

function configure(config: Config & ConfigAliasOptions)
```

@MattAgn wdyt?

src/helpers/findAll.ts
```diff
@@ -4,6 +4,7 @@ import { isHiddenFromAccessibility } from './accessiblity';
 
 interface FindAllOptions {
   hidden?: boolean;
+  includeHiddenElements?: boolean;
```
mdjastrzebski, line 7:
```suggestion
  includeHiddenElements?: boolean;
  /** RTL-compatible alias to `includeHiddenElements` */
  hidden?: boolean;
```

src/config.ts
MattAgn: yes I like that!

mdjastrzebski (approved):
👍🏻

PR #1222, opened 2022-11-11
fix: improve text matching

### LINKED ISSUE #1181: Feature: improved `findAll` method
## Describe the Feature
We currently use [`ReactTestInstance.findAll`](https://github.com/facebook/react/blob/9ecf84ed7ffb21c8746ecf227209c050d99393cd/packages/react-test-renderer/src/ReactTestRenderer.js#L402) method in order to traverse element tree. This is a basic tree walking algorithm that starts at the root and traverses recursively to the children. 

There is one major issue with the current method: for query predicates that inspect all descendants (e.g. concat all string content) there is no easy way to find only the deepest element matching given predicate.

```
<Text testID="outer">
  <Text testID="inner">
    Hello.
  </Text>
</Text>
```

When using a predict that will concat all nested string contents, then both outer and inner text will be matched. 

## Possible Implementations
This should be a function, with similar signature:

```ts
function findAll(element: ReactTestInstance, predicate: (element: ReactTestInstance) => boolean, options: FindAllOptions)
```

* it should work by traversing tree top-down from root (as current algorithm)
* but it should be inspecting the elements on its way back, so that the deepest element is matched first
* having information about deeper element having been matched inside given branch it would prevent it's ancestor elements also getting match
* this should be an optional behavior controlled by option passed inside `options` argument. Proposed initial option name: `deepestOnly` (default `false`)
* it might be worth considering adding some additional argument to `predicate` callback to facilitate this

## Applications

* This would be an internal API used by queries.
* It could allow us to improve some existing non-trivial queries, e.g. `*ByText` or `*ByRole` with `name` option

### LINKED ISSUE #1221: ByText queries don't seem to match across multiple Text elements
Give this test:
```
import React from 'react';
import { screen, render } from '@testing-library/react-native';
import { Text, View } from 'react-native';

function MyComponent() {
  return (
    <View>
      <Text>Hello</Text>
      <Text>World</Text>
    </View>
  );
}

it('renders correctly', async () => {
  render(<MyComponent />);
  const text = screen.getByText('Hello World');
  expect(text).toBeTruthy();
});
```
I would expect the test to pass, according to the docs the `Text` elements should be joined together. I've tried various permutations (regex's, non-exact match, extra Text with a space) and nothing seems to match.

## Expected behavior

Test should pass

## Steps to Reproduce

Run simple test as above (I can put in a repo if it helps)

## Screenshots

Screenshot 2022-11-10 at 15 09 06

## Versions
npmPackages:
    @testing-library/react-native: ^11.4.0 => 11.4.0 
    react: 18.1.0 => 18.1.0 
    react-native: 0.70.3 => 0.70.3 
    react-test-renderer: 18.1.0 => 18.1.0

mdjastrzebski:
### Summary

TLDR: Make this pass:
```tsx
test('byText matches accross multiple texts', () => {
  const { getByText } = render(
    <Text testID="outer">
      <Text testID="inner-1">Hello</Text> <Text testID="inner-2">World</Text>
    </Text>
  );
  expect(getByText('Hello World').props.testID).toBe('outer');
});
```

while not breaking this:
```tsx
test('byText matches inner nested text', () => {
  const { getByText } = render(
    <Text testID="outer">
      <Text testID="inner">Hello World</Text>
    </Text>
  );
  expect(getByText('Hello World').props.testID).toBe('inner');
});
```

Resolves #1221 
Resolves #1181

### What

Match only the deepest `Text` element containing given content while supporting nested `Text` elements.

In the light of #1221, user expectation for handling nested texts is reasonable, this MIGHT be considered as a bug fix rather than a breaking change. Or we might consider this breaking change, since the case is not clear cut.

All in all, as text matching has been complex subject in the past, let's be extra careful with this PR.

### How

1. Improve `findAll` function to accept `deepestOnly` option that when enabled prevents from matching given node if any of its descendants is matched. 
  a. This option could be used with other queries if useful, but can be breaking change depending on usage
  b. The updated `findAll` preserves the same tree walk order as initial one so `getByAll*` unit tests relying on matching order should work the same.
  c. New `findAll` is based on `findAll` from React Test Renderer, with necessary tweaks
2. Create a new `getTextContent()` function that can work with nested `Text` elements. Implementation based on [`toHaveTextContent()`](https://github.com/testing-library/jest-native/blob/main/src/to-have-text-content.ts) matcher from Jest Native 
3. Update `*ByText` queries to match only the deepest `Text` using new `getTextContent()` function.

This should keep only return a single match even in the case of nested `Text` elements. The matched `Text` element will be the deepest one that fulfils the predicate.

### Test plan

Add some simple test for desired effect. Make all existing test pass as is. Tweak only tests that actually checked that there is no nesting text matching.


AugustinLF:
Should we also considering matching?
```tsx
    <View>
      <Text>Hello</Text>
      <Text>World</Text>
    </View>
```

That's an open question, I feel it had some value. Do screen readers can read this type of text together, or do they treat that as distinct elements?

mdjastrzebski:
@AugustinLF I think in the short term we should not consider supporting that, as current `*ByText` queries have assumption that text can be only put into `Text` components, hence the return form `*ByText` should be only `Text` components. Changes to that behaviour would definitely be breaking.

Afaik how a11y & screen readers work, in your example the `Text` elements would be treated as separate focusable elements. That might be changed by grouping them under `View` using `<View accessible>`, which should group them together. We are getting into an "edge case" territory here, I would rather not support returning views, unless we can build a compelling real world & wide audience use case for such feature.

AugustinLF:
Yeah, the breaking change factor is a fair one, you're right. And if we want to allow that, it should probably start behind a flag.

src/helpers/textContent.ts
```diff
@@ -0,0 +1,20 @@
+import type { ReactTestInstance } from 'react-test-renderer';
+
+export function getTextContent(
```
MattAgn, line 3: i would name the file `getTextContent`, I find it easier to find, especially for people who discover the repo

src/matches.ts
```diff
@@ -1,5 +1,10 @@
 export type NormalizerFn = (textToNormalize: string) => string;
+
 export type TextMatch = string | RegExp;
+export type TextMatchOptions = {
```
MattAgn, line 4: did you move it in this file to make it closer to NormalizerFn?

src/helpers/findAll.ts
```diff
@@ -3,15 +3,19 @@ import { getConfig } from '../config';
 import { isHiddenFromAccessibility } from './accessiblity';
 
 interface FindAllOptions {
+  // Include elements that are hidden from accessibility
   hidden?: boolean;
+
+  // Exclude any ancestors of deepest matched elements even if they match the predicate
+  deepestOnly?: boolean;
```
MattAgn, line 10: for the naming, I think `matchDeepestOnly` would be a little clearer wdyt ?

src/helpers/findAll.ts
```diff
@@ -3,15 +3,19 @@ import { getConfig } from '../config';
 import { isHiddenFromAccessibility } from './accessiblity';
 
 interface FindAllOptions {
+  // Include elements that are hidden from accessibility
   hidden?: boolean;
+
+  // Exclude any ancestors of deepest matched elements even if they match the predicate
```
MattAgn, line 9: we could js doc here for better tooling

src/helpers/findAll.ts
```diff
@@ -3,15 +3,19 @@ import { getConfig } from '../config';
 import { isHiddenFromAccessibility } from './accessiblity';
 
 interface FindAllOptions {
+  // Include elements that are hidden from accessibility
```
MattAgn, line 6: we could js doc here for better tooling

MattAgn: we could js doc here for better tooling

src/helpers/findAll.ts
```diff
@@ -23,3 +27,35 @@ export function findAll(
...
+  const matchingDescendants: ReactTestInstance[] = [];
+  root.children.forEach((child) => {
+    if (typeof child === 'string') {
+      return;
+    }
+    matchingDescendants.push(...findAllInternal(child, predicate, options));
+  });
+
+  if (
+    // Deepest only mode: add current element only if no descendants match
```
MattAgn, line 50:
```suggestion
    // In deepest only mode: add current element only if no descendants match
```
I would add "in" otherwise we think at first the if condition concerns only the deep only mode

src/matches.ts
mdjastrzebski: I've moved it primarily to put it ntext to `TextMatch` :-)

MattAgn (approved):

AugustinLF (approved):
Really like the new API/file structure, I feel it makes things much clearer :)

I approve but do agree with the comment about renaming the `textContent` file to `getTextContent`.

MattAgn (approved):

src/__tests__/react-native-api.test.tsx
```diff
@@ -45,7 +45,7 @@ test('React Native API assumption: nested <Text> renders single host element', (
       </Text>
     </Text>
   );
-  expect(getHostSelf(view.getByText('Hello'))).toBe(view.getByTestId('test'));
+  expect(getHostSelf(view.getByText(/Hello/))).toBe(view.getByTestId('test'));
```
thymikee, line 48: is this a breaking change?

src/queries/__tests__/text.test.tsx
```diff
@@ -193,7 +216,7 @@ test('queryByText not found', () => {
 });
 
 test('queryByText does not match nested text across multiple <Text> in <Text>', () => {
```
thymikee, line 218: now it _does_ match

src/__tests__/react-native-api.test.tsx
mdjastrzebski: This test relied of nested Text not being matched, so I needed to change whole string match to partial match. This was a limitation on previous text matching and not a feature per se. IMO that is not a breaking change as test required a minor tweak without affecting the essence of the test.

thymikee (approved):

PR #1225, opened 2022-11-15
fix: remounting root (when using screen)

### LINKED ISSUE #996: `rerender`/`update` is not working correctly when unmounting root component
## Describe the bug

The following test fails
```
test('rerender works when re-mounting root', () => {
  const result = render(<Text>Mt. Everest</Text>);
  result.rerender(
    <View>
      <Text>Śnieżka</Text>
    </View>
  );
  expect(result.getByText('Śnieżka')).toBeTruthy();
});
```
with following error:

image

while the following test works ok
```
test('rerender works when re-mounting nested component', () => {
  const result = render(
    <View>
      <Text>Mt. Everest</Text>
    </View>
  );
  screen.rerender(
    <View>
      <View>
        <Text>Śnieżka</Text>
      </View>
    </View>
  );
  expect(result.getByText('Śnieżka')).toBeTruthy();
});
```

## Expected behavior

Both tests should work. Remounting root in rerender is not something you usually do, since you rather pass new `props` values. But that still could be used for some valid tests like for things like re-mounting (unmount + mount).

## Steps to Reproduce

Write the first test, can be in RNTL repo.

## Screenshots
None

## Versions

```
  npmPackages:
    react: ^17.0.2 => 17.0.2 
    react-native: ^0.66.0 => 0.66.0 
    react-test-renderer: ^17.0.2 => 17.0.2
    @testing-library/react-native: 10.0.0 => 10.0.0
``

### PR DESCRIPTION
**Pull request by mdjastrzebski**
### Summary

This partially resolves #996.

1. Supported use case:
```tsx
test('screen works with re-mounting root', () => {
  render(<Text key="1">Mt. Everest</Text>);
  expect(screen.getByText('Mt. Everest')).toBeTruthy();

  screen.rerender(<Text key="2">Śnieżka</Text>);
  expect(screen.getByText('Śnieżka')).toBeTruthy();
});
```

2. Not supported use case:
```tsx
test('render result works with re-mounting root', () => {
  const { rerender, getByText } = render(<Text key="1">Mt. Everest</Text>);
  expect(getByText('Mt. Everest')).toBeTruthy();

  rerender(<Text key="2">Śnieżka</Text>);
  // Next line throws "Unable to find node on an unmounted component."
  expect(getByText('Śnieżka')).toBeTruthy();
});
```
This fix rebinds `screen` on `update`/`rerender` call that change the `root` instance. In order to maintain current behaviour, queries are not re-bound when `root` instance is the same.

Due to our code structure it's quite complex to support 2. as the queries returned from `render` are already bound when captured by the user. This could be mitigated by applying some form of delegation but that would be quite complex change for edge case issue. I think that supporting `screen` use case is good enough for user that might encounter this issue. 

Resolves #996 

### Test plan

Added test covering the relevant case.

### DIFF
--- a/src/render.tsx
+++ b/src/render.tsx
@@ -132,6 +132,19 @@ function updateWithAct(
     act(() => {
       renderer.update(wrap(component));
     });
+
+    // Rebind queries and helper only when root instance changes.
+    // This will not affect render calls using `wrapper` or
+    // `unstable_validateStringsRenderedWithinText` options as they will create
+    // a wrapping root that does not change.
+    if (screen.container !== renderer.root) {
+      setRenderResult({
+        ...screen,
+        ...getQueriesForElement(renderer.root),
+        container: renderer.root,
+        debug: debug(renderer.root, renderer),
+      });
+    }
   };
 }
 
--- a/src/__tests__/screen.test.tsx
+++ b/src/__tests__/screen.test.tsx
@@ -23,12 +23,22 @@ test('screen holds last render result', () => {
   expect(screen.queryByText('Mt. Blanc')).toBeFalsy();
 });
 
-test('screen works with updating rerender', () => {
+test('screen works with rerender', () => {
   const result = render(<Text>Mt. Everest</Text>);
   expect(screen).toBe(result);
+  expect(screen.getByText('Mt. Everest')).toBeTruthy();
 
   screen.rerender(<Text>Śnieżka</Text>);
   expect(screen).toBe(result);
+  expect(screen.getByText('Śnieżka')).toBeTruthy();
+});
+
+test('screen works with re-mounting root', () => {
+  const result = render(<Text key="1">Mt. Everest</Text>);
+  expect(screen).toBe(result);
+  expect(screen.getByText('Mt. Everest')).toBeTruthy();
+
+  screen.rerender(<Text key="2">Śnieżka</Text>);
   expect(screen.getByText('Śnieżka')).toBeTruthy();
 });
 


src/render.tsx
```diff
@@ -132,6 +132,19 @@ function updateWithAct(
     act(() => {
       renderer.update(wrap(component));
     });
+
+    // Rebind queries and helper only when root instance changes.
+    // This will not affect render calls using `wrapper` or
+    // `unstable_validateStringsRenderedWithinText` options as they will create
+    // a wrapping root that does not change.
+    if (screen.container !== renderer.root) {
```
pierrezimmermannbam, line 140: When commenting this and adding the option unstable_validateStringsRenderedWithinText in the test it fails so I think the Profiler when using string validation does not prevent the root from changing.
 
Also if having a wrapper prevents the root from changing, could it be a solution to use some kind of default wrapper that would not change the queries behavior while still allow the root to stay the same ? (not sure at all if that's possible). If it worked it would probably cover both the use cases that currently fail

mdjastrzebski: Hmmm, good catch, I need to investigate it tomorrow.

pierrezimmermannbam (changes requested):
I think this is a very good approach ! I agree this is really an edge case and I think it's fine not supporting the 2nd case 

I made some tests to understand better the issue and it appeared to me  that your comment is not quite correct regarding the behavior when using unstable_validateStringsRenderedWithinText. 

I have also suggested another approach to consider but I'm not even sure it's worth looking into. This solution seems good enough to me considering it's very unlikely to encounter this issue

src/render.tsx
mdjastrzebski: re default wrapper: that would be the easiest solution, and is certain terms this is what RTL is doing. However, there are serious downsides:
1. it's a breaking change as `container` will now point to the artificial root
2. all `toJSON()` tests will break, because of artificial root.

While 1. would be acceptable, and we could wait till next major release. The 2. issue is actually quite nasty, as we do not have focused `toJSON()` and can only call it on the whole render tree.

mdjastrzebski:
This issue is connected with #1296 and can be affected by chosen solution

mdjastrzebski:
Closing as stale and relatively low importance.

PR #1233, opened 2022-11-18
docs: rewrite documentation on waitFor

pierrezimmermannbam:
### Summary

Rewrite waitFor documentation to be more complete. Explains further in details the internal behavior of waitFor and some of its pitfalls. Also add a section on how it works when using fake timers. I added as well a link to the readme for the part regarding the custom jest preset but I wonder if this should be in the docs as well rather than only in the readme.

 @mdjastrzebski @AugustinLF @MattAgn wdyt ?

### DIFF
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -547,7 +547,7 @@ waitFor(() => expect(1).toBe(2))
 You can enforce awaiting waitFor by using the [await-async-utils](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-utils.md) rule from [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library).
 :::
 
-Since `waitFor` is likely to run `expectation()` multiple times, it is highly recommended not to have any side effects in `waitFor`.
+Since `waitFor` is likely to run `expectation()` multiple times, it is [highly recommended not to have any side effects](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#performing-side-effects-in-waitfor) in `waitFor`.
 
 ```jsx
 await waitFor(() => {
@@ -560,6 +560,8 @@ await waitFor(() => {
 :::note
 Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
 :::
+
+It is also recommended to have a [single expectation per waitFor](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) for more consistency and faster failing tests. If you want to make several assertions, then they should be in seperated `waitFor`, or maybe the second assertion won't even require a `waitFor`.
 
 ### waitFor with jest fake timers
 
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -30,8 +30,8 @@ title: API
     - [On a `ScrollView`](#on-a-scrollview)
     - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
-    - [`waitFor with jest fake timers`](#waitfor-with-jest-fake-timers)
-      - [`waitFor with modern fake timers`](#waitfor-with-modern-fake-timers)
+    - [`With jest fake timers`](#with-jest-fake-timers)
+      - [`with modern fake timers`](#with-modern-fake-timers)
 - [`waitForElementToBeRemoved`](#waitforelementtoberemoved)
 - [`within`, `getQueriesForElement`](#within-getqueriesforelement)
 - [`query` APIs](#query-apis)
@@ -561,7 +561,7 @@ Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side
 
 It is also recommended to have a [single expectation per waitFor](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) for more consistency and faster failing tests. If you want to make several assertions, then they should be in seperated `waitFor`, or maybe the second assertion won't even require a `waitFor`.
 
-### waitFor with jest fake timers
+### With jest fake timers
 
 `waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. The following snippet is a simplified version of how it behaves when fake timers are enabled :
 
@@ -601,7 +601,7 @@ await waitFor(() => {
 If you want to wait for a long period of time, it can result in a very long test, especially when it is costly to run `expectation()`. In those cases, you can increase the `interval` as well so that you don't perform as many checks.
 :::
 
-#### waitFor with modern fake timers
+#### With modern fake timers
 
 When using modern fake timers (the default for jest >= 27), `waitFor` will systematically timeout unless you use the [custom @testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
 
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -558,7 +558,7 @@ await waitFor(() => {
 ```
 
 :::note
-Not using fireEvent in `waitFor` be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
+Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
 :::
 
 ### waitFor with jest fake timers
@@ -571,7 +571,7 @@ The following snippet is a simplified version of what waitFor does when fake tim
 let fakeTimeRemaining = timeout;
 
 while(fakeTimeRemaining > 0) {
-  fakeTimeReamining = fakeTimeRemaining - interval;
+  fakeTimeRemaining = fakeTimeRemaining - interval;
   jest.advanceTimersByTime(interval);
   try {
     expectation();
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -523,10 +523,8 @@ function waitFor<T>(
 
 Should be used when testing asynchronous code or code with timers. `waitFor` allows to wait for some expectation (e.g. the presence of an element on the screen). The `expectation` param is a callback that should throw until what you're waiting for has realized.
 
-For instance, this is how we'd wait for an element to appear:
-
 ```tsx
-await waitFor(() => expect(screen.getByText('Banana ready').toBeTruthy()))
+await waitFor(() => expect(mockFunction).toHaveBeenCalledWith()))
 ```
 
 What `waitFor` does is executing `expectation()` at intervals of duration equal to the optional `interval` param (default is 50ms) until it doesn't throw anymore. It will throw if it reaches its timeout, defined by the optional `timeout` param (default is 1000ms). If `expectation()` does not throw then `waitFor` will resolve right away.
@@ -553,7 +551,7 @@ Since `waitFor` is likely to run `expectation()` multiple times, it is [highly r
 await waitFor(() => {
   // button will be pressed on each waitFor iteration
   fireEvent.press(screen.getByText('press me'))
-  expect(screen.getByText('pressed')).toBeTruthy()
+  expect(mockOnPress).toHaveBeenCalled()
 })
 ```
 
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -30,8 +30,7 @@ title: API
     - [On a `ScrollView`](#on-a-scrollview)
     - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
-    - [`With jest fake timers`](#with-jest-fake-timers)
-      - [`with modern fake timers`](#with-modern-fake-timers)
+    - [With jest fake timers](#with-jest-fake-timers)
 - [`waitForElementToBeRemoved`](#waitforelementtoberemoved)
 - [`within`, `getQueriesForElement`](#within-getqueriesforelement)
 - [`query` APIs](#query-apis)
@@ -527,17 +526,17 @@ Waits for a period of time for the `expectation` callback to pass. `waitFor` may
 await waitFor(() => expect(mockFunction).toHaveBeenCalledWith()))
 ```
 
-What `waitFor` does is executing `expectation()` at intervals of duration equal to the optional `interval` param (default is 50ms) until it doesn't throw anymore. It will throw if it reaches its timeout, defined by the optional `timeout` param (default is 1000ms). If `expectation()` does not throw then `waitFor` will resolve right away.
+`waitFor` function will be executing `expectation` callback every `interval` (default: every 50 ms) until `timeout` (default: 1000 ms) is reached. The repeated execution of callback is stopped as soon as it does not throw an error, in such case the value returned by the callback is returned to `waitFor` caller. Otherwise, when it reaches the timeout, the final error thrown by `expectation` will be re-thrown by `waitFor` to the calling code.
 
 ```tsx
-// ❌ won't throw because callback never throws
+// ❌ `waitFor` will return immediately because callback does not throw
 await waitFor(() => false);
 ```
 
-`waitFor` only rejects when it reaches its timeout so if you don't await it it will never throw.
+`waitFor` is an async function so you need to `await` the result to pause test execution.
 
 ```jsx
-// ❌ won't throw because waitFor is not awaited
+// ❌ missing `await`: `waitFor` will just return Promise that will be rejected when the timeout is reached
 waitFor(() => expect(1).toBe(2))
 ```
 
@@ -563,7 +562,11 @@ It is also recommended to have a [single expectation per waitFor](https://kentcd
 
 ### With jest fake timers
 
-`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. The following snippet is a simplified version of how it behaves when fake timers are enabled:
+:::caution
+When using modern fake timers (the default for jest >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the custom [@testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
+:::
+
+`waitFor` checks whether Jest fake timers are enabled and adapts its behavior in such case. The following snippet is a simplified version of how it behaves when fake timers are enabled:
 
 ```tsx
 let fakeTimeRemaining = timeout;
@@ -580,7 +583,7 @@ while(fakeTimeRemaining > 0) {
 // reject
 ```
 
-This implementation doesn't depend on real time and automatically advances fake timers.
+This implementation doesn't depend on real time and automatically advances fake timers so you can use `waitFor` when testing timeouts without using Jest fake timers'API to manually advance time.
 
 ```tsx
 // in component
@@ -596,14 +599,6 @@ await waitFor(() => {
   expect(someFunction).toHaveBeenCalledWith();
 }, 10000)
 ```
-
-:::note
-If you want to wait for a long period of time, it can result in a very long test, especially when it is costly to run `expectation()`. In those cases, you can increase the `interval` as well so that you don't perform as many checks.
-:::
-
-#### With modern fake timers
-
-When using modern fake timers (the default for jest >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the [custom @testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
 
 :::info
 In order to properly use `waitFor` you need at least React >=16.9.0 (featuring async `act`) or React Native >=0.61 (which comes with React >=16.9.0).
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -30,6 +30,8 @@ title: API
     - [On a `ScrollView`](#on-a-scrollview)
     - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
+    - [`waitFor with jest fake timers`](#waitfor-with-jest-fake-timers)
+      - [`waitFor with modern fake timers`](#waitfor-with-modern-fake-timers)
 - [`waitForElementToBeRemoved`](#waitforelementtoberemoved)
 - [`within`, `getQueriesForElement`](#within-getqueriesforelement)
 - [`query` APIs](#query-apis)
@@ -515,21 +517,95 @@ Defined as:
 ```jsx
 function waitFor<T>(
   expectation: () => T,
-  { timeout: number = 4500, interval: number = 50 }
+  { timeout: number = 1000, interval: number = 50 }
 ): Promise<T> {}
 ```
 
-Waits for non-deterministic periods of time until your element appears or times out. `waitFor` periodically calls `expectation` every `interval` milliseconds to determine whether the element appeared or not.
+Should be used when testing asynchronous code or code with timers. `waitFor` allows to wait for some expectation (e.g. the presence of an element on the screen). The `expectation` param is a callback that should throw until what you're waiting for has realized.
+
+For instance, this is how we'd wait for an element to appear:
+
+```tsx
+await waitFor(() => expect(screen.getByText('Banana ready').toBeTruthy()))
+```
+
+What `waitFor` does is executing `expectation()` at intervals of duration equal to the optional `interval` param (default is 50ms) until it doesn't throw anymore. It will throw if it reaches its timeout, defined by the optional `timeout` param (default is 1000ms). If `expectation()` does not throw then `waitFor` will resolve right away.
+
+```tsx
+// won't throw
+await waitFor(() => false);
+```
+
+`waitFor` only rejects when it reaches its timeout so if you don't await it it will never throw.
 
 ```jsx
-import { render, screen, waitFor } from '@testing-library/react-native';
-
-test('waiting for an Banana to be ready', async () => {
-  render(<Banana />);
-
-  await waitFor(() => screen.getByText('Banana ready'));
-});
+// won't throw
+waitFor(() => expect(1).toBe(2))
 ```
+
+:::note
+You can enforce awaiting waitFor by using the [await-async-utils](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-utils.md) rule from [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library).
+:::
+
+Since `waitFor` is likely to run `expectation()` multiple times, it is highly recommended not to have any side effects in `waitFor`.
+
+```jsx
+await waitFor(() => {
+  // button will be pressed on each waitFor iteration
+  fireEvent.press(screen.getByText('press me'))
+  expect(screen.getByText('pressed')).toBeTruthy()
+})
+```
+
+:::note
+Not using fireEvent in `waitFor` be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
+:::
+
+### waitFor with jest fake timers
+
+`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. It will create a variable counting the time remaining which initially equals `timeout` and is decreased on each iteration by `interval`. When that variable is negative it will reject. Additionnaly, on each iteration it will call `jest.advanceTimersByTime(interval)`.
+
+The following snippet is a simplified version of what waitFor does when fake timers are enabled:
+
+```tsx
+let fakeTimeRemaining = timeout;
+
+while(fakeTimeRemaining > 0) {
+  fakeTimeReamining = fakeTimeRemaining - interval;
+  jest.advanceTimersByTime(interval);
+  try {
+    expectation();
+    // resolve
+  }
+}
+
+// reject
+```
+
+This allows to test code with timers by only using waitFor and without manually advancing timers through jest API.
+
+```tsx
+// in component
+setTimeout(() => {
+  someFunction();
+}, 10000)
+
+
+// in test
+jest.useFakeTimers();
+
+await waitFor(() => {
+  expect(someFunction).toHaveBeenCalledWith();
+}, 10000)
+```
+
+:::note
+If you want to wait for a long period of time, it can result in a very long test, especially when it is costly to run `expectation()`. In those cases, you can increase the `interval` as well so that you don't perform as many checks.
+:::
+
+#### waitFor with modern fake timers
+
+When using modern fake timers (the default for jest >= 27), `waitFor` will systematically timeout unless you use the [custom @testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
 
 :::info
 In order to properly use `waitFor` you need at least React >=16.9.0 (featuring async `act`) or React Native >=0.61 (which comes with React >=16.9.0).
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -549,7 +549,7 @@ Since `waitFor` is likely to run `expectation()` multiple times, it is [highly r
 
 ```jsx
 await waitFor(() => {
-  // button will be pressed on each waitFor iteration
+  // ❌ button will be pressed on each waitFor iteration
   fireEvent.press(screen.getByText('press me'))
   expect(mockOnPress).toHaveBeenCalled()
 })
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -563,9 +563,7 @@ Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side
 
 ### waitFor with jest fake timers
 
-`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. It will create a variable counting the time remaining which initially equals `timeout` and is decreased on each iteration by `interval`. When that variable is negative it will reject. Additionnaly, on each iteration it will call `jest.advanceTimersByTime(interval)`.
-
-The following snippet is a simplified version of what waitFor does when fake timers are enabled:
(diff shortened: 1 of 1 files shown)


website/docs/API.md
```diff
@@ -515,22 +517,96 @@ Defined as:
...
+```jsx
+await waitFor(() => {
+  // button will be pressed on each waitFor iteration
+  fireEvent.press(screen.getByText('press me'))
+  expect(screen.getByText('pressed')).toBeTruthy()
+})
 ```
 
+:::note
+Not using fireEvent in `waitFor` be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
```
MattAgn, line 561:
```suggestion
Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
```

website/docs/API.md
```diff
@@ -515,22 +517,96 @@ Defined as:
...
+:::
+
+### waitFor with jest fake timers
```
MattAgn, line 564: I love the detailed explanation but I think a big part of this new doc should go in the Advanced Section like [Understanding the act function](https://callstack.github.io/react-native-testing-library/docs/understanding-act). When you look at the rest of the docs, it's very concise and does not go too much in details about the implementation, wdyt?

website/docs/API.md
```diff
@@ -515,22 +517,96 @@ Defined as:
...
+
+`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. It will create a variable counting the time remaining which initially equals `timeout` and is decreased on each iteration by `interval`. When that variable is negative it will reject. Additionnaly, on each iteration it will call `jest.advanceTimersByTime(interval)`.
```
MattAgn, line 566: I think your simplified code below is enough to understand what's happening, no need to paraphrase it here

website/docs/API.md
```diff
@@ -515,22 +517,96 @@ Defined as:
...
+}, 10000)
+```
+
+:::note
+If you want to wait for a long period of time, it can result in a very long test, especially when it is costly to run `expectation()`. In those cases, you can increase the `interval` as well so that you don't perform as many checks.
+:::
+
+#### waitFor with modern fake timers
+
+When using modern fake timers (the default for jest >= 27), `waitFor` will systematically timeout unless you use the [custom @testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset).
```
MattAgn, line 608: Super nice to mention that! I struggled quite a bit when I first encountered this issue

website/docs/API.md
```diff
@@ -515,22 +517,96 @@ Defined as:
...
+
+`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. It will create a variable counting the time remaining which initially equals `timeout` and is decreased on each iteration by `interval`. When that variable is negative it will reject. Additionnaly, on each iteration it will call `jest.advanceTimersByTime(interval)`.
+
+The following snippet is a simplified version of what waitFor does when fake timers are enabled:
+
+```tsx
+let fakeTimeRemaining = timeout;
+
+while(fakeTimeRemaining > 0) {
+  fakeTimeReamining = fakeTimeRemaining - interval;
```
MattAgn, line 574: `fakeTimeReamining` -> `fakeTimeRemaining`

website/docs/API.md
pierrezimmermannbam: I agree this is not at the right place but I'm not sure either that moving this doc to the advanced section is the best solution. What I don't like about advanced section is that users are way less likely to read it as it seems unnecessary to correctly use the library. I'm not exactly sure what should go in advanced section or not but I think understanding how waitFor works is both too valuable and not too complex to be in advanced section. Also the detailed explanations can easily be skipped so I don't think it's an issue for this doc to be in a single place.

Maybe the API page is meant to only give available APIs and a brief explanation but if that's the case if think more sections should be added to cover stuff like this more in detail. For instance this could be in a section about testing asynchronous code or code with timers. I feel like right now there is a huge list of available APIs but it's kind of hard to know what these APIs allow to do. For instance if I don't know how to test asynchronous code, I will need to read through the API page and find the waitFor section, and even then it may not be detailed enough because this page is quite succinct. 

What I'm thinking of right now would be additional sections like 'render component', 'use queries', 'trigger events', 'test asynchronous code' for instance, that would be under the guide section and each of these sections would provide more detailed examples of the API usage.

@MattAgn  @mdjastrzebski  @AugustinLF wdyt ?

website/docs/API.md
pierrezimmermannbam: Agreed, I reworked this part a bit, I think it's better, tell me how you feel about it

website/docs/API.md
MattAgn: I like your proposition to have more `guides` section, could be useful to many people

website/docs/API.md
```diff
@@ -563,9 +563,7 @@ Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side
 
 ### waitFor with jest fake timers
 
-`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. It will create a variable counting the time remaining which initially equals `timeout` and is decreased on each iteration by `interval`. When that variable is negative it will reject. Additionnaly, on each iteration it will call `jest.advanceTimersByTime(interval)`.
-
-The following snippet is a simplified version of what waitFor does when fake timers are enabled:
+`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. The following snippet is a simplified version of how it behaves when fake timers are enabled :
```
MattAgn, line 566: `are enabled :` -> `are enabled:`

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
 ): Promise<T> {}
 ```
 
-Waits for non-deterministic periods of time until your element appears or times out. `waitFor` periodically calls `expectation` every `interval` milliseconds to determine whether the element appeared or not.
+Should be used when testing asynchronous code or code with timers. `waitFor` allows to wait for some expectation (e.g. the presence of an element on the screen). The `expectation` param is a callback that should throw until what you're waiting for has realized.
+
+For instance, this is how we'd wait for an element to appear:
+
+```tsx
+await waitFor(() => expect(screen.getByText('Banana ready').toBeTruthy()))
```
mdjastrzebski, line 529: No need to wrap `getBy*` queries in `expect` since `getBy` is already throwing in case of match failure. 
```suggestion
await waitFor(() => screen.getByText('Banana ready'))
```

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+```
+
+What `waitFor` does is executing `expectation()` at intervals of duration equal to the optional `interval` param (default is 50ms) until it doesn't throw anymore. It will throw if it reaches its timeout, defined by the optional `timeout` param (default is 1000ms). If `expectation()` does not throw then `waitFor` will resolve right away.
```
mdjastrzebski, line 532:
```suggestion
What `waitFor` does is executing `expectation()` at intervals of duration equal to the optional `interval` param (default is 50ms) until it doesn't throw anymore. It will re-throw the final error thrown by `expectation` when it reaches the timeout, defined by the optional `timeout` param (default is 1000ms). If `expectation()` does not throw then `waitFor` will resolve right away.
```

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+// won't throw
+await waitFor(() => false);
+```
+
+`waitFor` only rejects when it reaches its timeout so if you don't await it it will never throw.
 
 ```jsx
-import { render, screen, waitFor } from '@testing-library/react-native';
+// won't throw
+waitFor(() => expect(1).toBe(2))
```
mdjastrzebski, line 543: Isn't this an incorrect usage of `waitFor`? Is there any use case user would want to just call `waitFor` without awaiting it?

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
 
-test('waiting for an Banana to be ready', async () => {
-  render(<Banana />);
+:::note
+You can enforce awaiting waitFor by using the [await-async-utils](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-utils.md) rule from [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library).
+:::
 
-  await waitFor(() => screen.getByText('Banana ready'));
-});
+Since `waitFor` is likely to run `expectation()` multiple times, it is highly recommended not to have any side effects in `waitFor`.
```
mdjastrzebski, line 550: I think we should express clearly that expectiation should contain only a single condition to wait for e.g. query, not multiple ones. If user needs to wait for couple of things, they should use multiple waitFors: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+  { timeout: number = 1000, interval: number = 50 }
 ): Promise<T> {}
 ```
 
-Waits for non-deterministic periods of time until your element appears or times out. `waitFor` periodically calls `expectation` every `interval` milliseconds to determine whether the element appeared or not.
+Should be used when testing asynchronous code or code with timers. `waitFor` allows to wait for some expectation (e.g. the presence of an element on the screen). The `expectation` param is a callback that should throw until what you're waiting for has realized.
+
+For instance, this is how we'd wait for an element to appear:
+
+```tsx
```
mdjastrzebski, line 528: I think we should avoid cases for `waitFor` that uses queries, as users should be rather using `findBy*` queries rather than wrapping `getBy*` queries in `waitFor`. Ref; https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find

website/docs/API.md
mdjastrzebski: Re side effects: ref. https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#performing-side-effects-in-waitfor

website/docs/API.md
```diff
@@ -30,6 +30,8 @@ title: API
     - [On a `ScrollView`](#on-a-scrollview)
     - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
+    - [`waitFor with jest fake timers`](#waitfor-with-jest-fake-timers)
```
mdjastrzebski, line 33: Can we simplify headers to avoid `waitFor` repetition, e.g. `Using with Jest fake timers`, `Modern fake timers`

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+}, 10000)
 ```
 
+:::note
+If you want to wait for a long period of time, it can result in a very long test, especially when it is costly to run `expectation()`. In those cases, you can increase the `interval` as well so that you don't perform as many checks.
+:::
+
+#### waitFor with modern fake timers
+
+When using modern fake timers (the default for jest >= 27), `waitFor` will systematically timeout unless you use the [custom @testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset).
```
mdjastrzebski, line 606: Not sure I understand the description here. What does it mean that it "will systematically timeout"? Does it mean that it basically won't work unless users uses the preset? Could we revise the working here for clarity?

website/docs/API.md
mdjastrzebski, line 532: This is a bit of simplification, as under node `setTimeout`/`setInterval` are not guaranteeing exact amount of timeout/interval but rather "at least" amount. They might be run later, e.g. due to JS thread being busy by some longer computation. I don't think we need to explain the whole story here, but just tweak the wording a bit:
``waitFor` periodically calls `expectation` approximately every `interval` milliseconds`

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
 ```jsx
 function waitFor<T>(
   expectation: () => T,
-  { timeout: number = 4500, interval: number = 50 }
+  { timeout: number = 1000, interval: number = 50 }
 ): Promise<T> {}
 ```
 
-Waits for non-deterministic periods of time until your element appears or times out. `waitFor` periodically calls `expectation` every `interval` milliseconds to determine whether the element appeared or not.
+Should be used when testing asynchronous code or code with timers. `waitFor` allows to wait for some expectation (e.g. the presence of an element on the screen). The `expectation` param is a callback that should throw until what you're waiting for has realized.
```
mdjastrzebski, line 524: I did like to have a concise explanation of the function as before (expect that is was inaccurate). We can have a longer explanation following it, but we should not expect the user to read the whole explanation to get all key points.

Using the previous style, but being accurate, this might be something like:
```
Waits for a period of time for the `expectation` callback to pass. `waitFor` may run the callback a number of times until timeout is reached, as specified by the `timeout` and `interval` options. The callback must throw an error when the expectation is not met. Returning any value, including a falsy one, will be treated as meeting the expectation, and the callback result will be returned to the caller of `waitFor` function.
```

website/docs/API.md
AugustinLF: I think that's a problem we have with our whole doc, if you look at [RTL's docs](https://testing-library.com/docs/dom-testing-library/api-async), they have both API docs and something more detailed.

With our current structure, I'd suggest to keep this in `API.md`, advanced section is too hidden.

website/docs/API.md
pierrezimmermannbam: While this also works I find it to be less explicit regarding the intent of this line, with explicit expect I find tests to be easier to understand. This is also the approach recommended by Kent C. Dodds (https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions). If the documentation uses plain get queries then it is very likely that users will use them as is so I think tests in the documentation should mirror what we consider to be the best practices. The question is wether we think or not that this is the best practice, it's not if the upmost importance so if you feel strongly about it we can recommend using plain get queries

website/docs/API.md
pierrezimmermannbam: It is, I wanted to showcase it but it must mean that it's not explicit enough as is, I'll try to make it more obvious

website/docs/API.md
pierrezimmermannbam: Definitely, I added this as well and linked the refs

website/docs/API.md
pierrezimmermannbam: Agreed, I have updated the examples

website/docs/API.md
pierrezimmermannbam: Yes it won't work, I rewrote this part (just added that it won't work), let me know if that's clear enough for you

website/docs/API.md
pierrezimmermannbam: You're right I hadn't thought of this. However that's never really been an issue for me, I wonder if it's worth mentioning in the doc. I fear that the term `approximately` may just be the source of more confusion as it's really not cleat as to why it would be approximated, is there value in users knowing this ? wdyt ?

website/docs/API.md
pierrezimmermannbam: If find your suggestion really good ! Very clear and concise that's perfect ! I replaced the first paragraph with your suggestion and it feels a lot better

website/docs/API.md
pierrezimmermannbam: This part has been replaced by your suggestion

pierrezimmermannbam:
@MattAgn @mdjastrzebski thanks very much for the feedbacks ! I believe I responded to all your comments

website/docs/API.md
MattAgn: much better!

website/docs/API.md
MattAgn: maybe worth mentionning that in a real life scenario, you should use findBy?

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
-});
+Since `waitFor` is likely to run `expectation()` multiple times, it is [highly recommended not to have any side effects](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#performing-side-effects-in-waitfor) in `waitFor`.
+
+```jsx
+await waitFor(() => {
+  // button will be pressed on each waitFor iteration
+  fireEvent.press(screen.getByText('press me'))
+  expect(mockOnPress).toHaveBeenCalled()
+})
+```
```
MattAgn, line 556: you could also put a ❌ emoji around here to indicate it's a bad pattern

website/docs/API.md
MattAgn: I agree with Pierre about "approximately", maybe we could add a little note at the end saying: "Under the hood, waitFor uses setTimeout and setInterval so the `interval` and `timeout` values are "at least" amounts. They might be run later if the JS thread is busy. More on that [here](https://www.youtube.com/watch?v=8aGhZQkoFbQ)" (I pointed to one ressource but have no strong preference)

website/docs/API.md
pierrezimmermannbam: done !

website/docs/API.md
```diff
@@ -30,6 +30,8 @@ title: API
     - [On a `ScrollView`](#on-a-scrollview)
     - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
+    - [`With jest fake timers`](#with-jest-fake-timers)
+      - [`with modern fake timers`](#with-modern-fake-timers)
```
mdjastrzebski, line 34: Why are the titles in backquotes here? 
```suggestion
    - [With jest fake timers](#with-jest-fake-timers)
      - [With modern fake timers](#with-modern-fake-timers)
```

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+
+:::note
+Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
+:::
+
+It is also recommended to have a [single expectation per waitFor](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) for more consistency and faster failing tests. If you want to make several assertions, then they should be in seperated `waitFor`, or maybe the second assertion won't even require a `waitFor`.
+
+### With jest fake timers
+
+`waitFor` checks wether jest fake timers are enabled and adapts its behavior when it is the case. The following snippet is a simplified version of how it behaves when fake timers are enabled:
```
mdjastrzebski, line 566:
```suggestion
`waitFor` checks whether Jest fake timers are enabled and adapts its behavior in such case. The following snippet is a simplified version of how it behaves when fake timers are enabled:
```

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+// in component
+setTimeout(() => {
+  someFunction();
+}, 10000)
+
+
+// in test
+jest.useFakeTimers();
+
+await waitFor(() => {
```
mdjastrzebski, line 595: The example does not seem to be related to the text around. @pierrezimmermannbam what is your intent here? Maybe you can add some text or code comment to explain it.

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+  expect(someFunction).toHaveBeenCalledWith();
+}, 10000)
 ```
 
+:::note
+If you want to wait for a long period of time, it can result in a very long test, especially when it is costly to run `expectation()`. In those cases, you can increase the `interval` as well so that you don't perform as many checks.
```
mdjastrzebski, line 601: This comments seems confusing, not sure it relates to waiting under real or fake timers? It also seems quite technical and low level. Pls clarify it or consider removing it.

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+:::
+
+#### With modern fake timers
```
mdjastrzebski, line 604: Since modern fake timers are the default ones, and the remarks seems important as without this step, the timers won't work, I think we should move it up to the begging of the fake timer section, and wrap it in :::caution block (without #### header)

website/docs/API.md
mdjastrzebski: I think it's good to update this part not to repeat the first paragraphy. How about something like this:

`waitFor` function will be executing `expectation` callback every `interval` (default: every 50 ms) until `timeout` (default: 1000 ms) is reached.  The repeated execution of callback is stopped as soon as it does not throw an error, in such case the value returned by the callback is returned to `waitFor` caller. Otherwise, when it reaches the timeout, the final error thrown by `expectation` will be re-thrown by `waitFor` to the calling code.

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+
+```tsx
+await waitFor(() => expect(mockFunction).toHaveBeenCalledWith()))
+```
+
+What `waitFor` does is executing `expectation()` at intervals of duration equal to the optional `interval` param (default is 50ms) until it doesn't throw anymore. It will throw if it reaches its timeout, defined by the optional `timeout` param (default is 1000ms). If `expectation()` does not throw then `waitFor` will resolve right away.
+
+```tsx
+// ❌ won't throw because callback never throws
+await waitFor(() => false);
```
mdjastrzebski, line 534:
```suggestion
// ❌ `waitFor` will return immediately because callback does not throw
await waitFor(() => false);
```

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
+```
+
+`waitFor` only rejects when it reaches its timeout so if you don't await it it will never throw.
```
mdjastrzebski, line 537:
```suggestion
`waitFor` as an async function so you need to `await` the result to pause test execution.
```

website/docs/API.md
```diff
@@ -515,22 +517,94 @@ Defined as:
...
 
 ```jsx
-import { render, screen, waitFor } from '@testing-library/react-native';
+// ❌ won't throw because waitFor is not awaited
+waitFor(() => expect(1).toBe(2))
```
mdjastrzebski, line 541:
```suggestion
// ❌ missing `await`: `waitFor` will just return Promise that will be rejected when the timeout is reached
waitFor(() => expect(1).toBe(2))
```

website/docs/API.md
MattAgn: I don't see many cases where expectation was long to run, have you encountered such a case before?
if not, I vote for removing it as well

website/docs/API.md
MattAgn:
```suggestion
`waitFor` is an async function so you need to `await` the result to pause test execution.
```

website/docs/API.md
pierrezimmermannbam: Juste some bad copy paste, fixed it

website/docs/API.md
pierrezimmermannbam: Actually it is related, I want to show that waitFor can be used when testing timers when fake timers aren't enabled, it must have been to implicit, I rewrote this part so it's clearer

website/docs/API.md
pierrezimmermannbam: It relates to having fake timers, since when using real timers it will depend on real time. I have encountered this situation before yes, if you're testing let's say a timeout of 10 minutes or maybe more and keep the default 50ms interval you're going to run quite a lot of queries without it being useful. It's more of an edge case though so I removed  it, maybe it should be mentioned in some advanced section but I'll leave it out for now

website/docs/API.md
pierrezimmermannbam: Maybe this should be detailed in a section about waitFor but I wouldn't add so much precision in this part of the doc, wdyt ?

pierrezimmermannbam: Looks good to me !

website/docs/API.md
pierrezimmermannbam: If feel like it would rather belong in findBy doc, wdyt ?

website/docs/API.md
MattAgn: Agreed

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+  try {
+    expectation();
+    // resolve
+  }
+}
+
+// reject
+```
+
+This implementation doesn't depend on real time and automatically advances fake timers so you can use `waitFor` when testing timeouts without using Jest fake timers'API.
```
MattAgn, line 586: I would be a little more precise here, wdyt?
```suggestion
This implementation doesn't depend on real time and automatically advances fake timers so you can use `waitFor` when testing timeouts without using Jest fake timers'API to manually advance time.
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
 ```jsx
 function waitFor<T>(
   expectation: () => T,
-  { timeout: number = 4500, interval: number = 50 }
+  { timeout: number = 1000, interval: number = 50 }
 ): Promise<T> {}
 ```
 
-Waits for non-deterministic periods of time until your element appears or times out. `waitFor` periodically calls `expectation` every `interval` milliseconds to determine whether the element appeared or not.
+Waits for a period of time for the `expectation` callback to pass. `waitFor` may run the callback a number of times until timeout is reached, as specified by the `timeout` and `interval` options. The callback must throw an error when the expectation is not met. Returning any value, including a falsy one, will be treated as meeting the expectation, and the callback result will be returned to the caller of `waitFor` function.
```
MattAgn, line 523: Not sure about the wording "waits ...for the callback to pass", is it the term to use in our use case? I don't have a much better solution though, maybe:
```suggestion
Waits for a period of time for the `expectation` callback to run without throwing. `waitFor` may run the callback a number of times until timeout is reached, as specified by the `timeout` and `interval` options. The callback must throw an error when the expectation is not met. Returning any value, including a falsy one, will be treated as meeting the expectation, and the callback result will be returned to the caller of `waitFor` function.
```

pierrezimmermannbam: It is specified in the same paragraph that the callback must throw an error and it is also further detailed with an example so I don't think there is a need for more precision here, wdyt ?

website/docs/API.md
pierrezimmermannbam: Agreed, I applied your suggestion

website/docs/API.md
MattAgn: agreed, it was not meant to add precision, i was just not sure about the use of the verb "pass" (I doubted my english programming vocabulary ^^) but overall yes it's very clear :) we can leave it as is

MattAgn (approved):

website/docs/API.md
```diff
@@ -30,6 +30,7 @@ title: API
     - [On a `ScrollView`](#on-a-scrollview)
     - [On a `FlatList`](#on-a-flatlist)
 - [`waitFor`](#waitfor)
+    - [With jest fake timers](#with-jest-fake-timers)
```
mdjastrzebski, line 33:
```suggestion
    - [Using Jest fake timers](#with-jest-fake-timers)
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
 ```jsx
-import { render, screen, waitFor } from '@testing-library/react-native';
+// ❌ missing `await`: `waitFor` will just return Promise that will be rejected when the timeout is reached
+waitFor(() => expect(1).toBe(2))
+```
 
-test('waiting for an Banana to be ready', async () => {
-  render(<Banana />);
+:::note
+You can enforce awaiting waitFor by using the [await-async-utils](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-utils.md) rule from [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library).
```
mdjastrzebski, line 544:
```suggestion
You can enforce awaiting `waitFor` by using the [await-async-utils](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-utils.md) rule from [eslint-plugin-testing-library](https://github.com/testing-library/eslint-plugin-testing-library).
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+```jsx
+await waitFor(() => {
+  // ❌ button will be pressed on each waitFor iteration
+  fireEvent.press(screen.getByText('press me'))
+  expect(mockOnPress).toHaveBeenCalled()
+})
+```
+
+:::note
+Not using fireEvent in `waitFor` can be enforced with the rule [no-wait-for-side-effects](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
```
mdjastrzebski, line 558:
```suggestion
Avoiding side effects in `expectation` callback can be partially enforced with the [`no-wait-for-side-effects` rule](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-wait-for-side-effects.md).
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+:::
+
+It is also recommended to have a [single expectation per waitFor](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) for more consistency and faster failing tests. If you want to make several assertions, then they should be in seperated `waitFor`, or maybe the second assertion won't even require a `waitFor`.
```
mdjastrzebski, line 561:
```suggestion
It is also recommended to have a [single assertion per each `waitFor`](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#having-multiple-assertions-in-a-single-waitfor-callback) for more consistency and faster failing tests. If you want to make several assertions, then they should be in separate `waitFor` calls. In many cases you won't actually need to wrap the second assertion in `waitFor` since the first one will do the waiting required for asynchronous change to happen.
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+
+### With jest fake timers
```
mdjastrzebski, line 563:
```suggestion
### Using Jest fake timers
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+
+:::caution
+When using modern fake timers (the default for jest >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the custom [@testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset).
```
mdjastrzebski, line 566:
```suggestion
When using modern fake timers (the default for `jest` >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the custom [@testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+```tsx
+let fakeTimeRemaining = timeout;
+
+while(fakeTimeRemaining > 0) {
+  fakeTimeRemaining = fakeTimeRemaining - interval;
+  jest.advanceTimersByTime(interval);
+  try {
+    expectation();
+    // resolve
+  }
```
mdjastrzebski, line 580:
```suggestion
  } catch (error) { }
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+`waitFor` checks whether Jest fake timers are enabled and adapts its behavior in such case. The following snippet is a simplified version of how it behaves when fake timers are enabled:
+
+```tsx
+let fakeTimeRemaining = timeout;
+
+while(fakeTimeRemaining > 0) {
+  fakeTimeRemaining = fakeTimeRemaining - interval;
+  jest.advanceTimersByTime(interval);
+  try {
+    expectation();
```
mdjastrzebski, line 578:
```suggestion
    return expectation();
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+### With jest fake timers
+
+:::caution
+When using modern fake timers (the default for jest >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the custom [@testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
+:::
+
+`waitFor` checks whether Jest fake timers are enabled and adapts its behavior in such case. The following snippet is a simplified version of how it behaves when fake timers are enabled:
+
+```tsx
+let fakeTimeRemaining = timeout;
```
mdjastrzebski, line 572:
```suggestion
let fakeTimeRemaining = timeout;
let lastError;
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+while(fakeTimeRemaining > 0) {
+  fakeTimeRemaining = fakeTimeRemaining - interval;
+  jest.advanceTimersByTime(interval);
+  try {
+    expectation();
+    // resolve
+  }
+}
+
+// reject
```
mdjastrzebski, line 583:
```suggestion
// reject
throw lastError;
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+
+```tsx
+let fakeTimeRemaining = timeout;
+
+while(fakeTimeRemaining > 0) {
+  fakeTimeRemaining = fakeTimeRemaining - interval;
+  jest.advanceTimersByTime(interval);
+  try {
+    expectation();
+    // resolve
```
mdjastrzebski, line 579:
```suggestion
    // resolve
    return expectation();
```

website/docs/API.md
```diff
@@ -515,20 +516,88 @@ Defined as:
...
+  }
+}
+
+// reject
+```
+
+This implementation doesn't depend on real time and automatically advances fake timers so you can use `waitFor` when testing timeouts without using Jest fake timers'API to manually advance time.
```
mdjastrzebski, line 586:
```suggestion
This implementation doesn't depend on real time passing and automatically advances fake timers so you can use `waitFor` when testing timeouts without explicitly calling Jest fake timers API to manually advance time.
```

website/docs/API.md
mdjastrzebski: I still have some trouble understanding this use case. Maybe we could prefix it with something like:

In the below example, `setTimeout(..., 10000)` is used to simulate async event that resolves after 10 seconds. However, when using fake timers, the actual time to run the test will be much faster, since it does not rely on real time passing.

mdjastrzebski (approved):
Awesome job @pierrezimmermannbam! This document needed an update, and I think you've done such a great job here 🥇

Issue #1184, opened 2022-10-18
RFC: simplify `Text` and `TextInput` related queries


mdjastrzebski:
## Describe the Feature
Currently when looking for `Text` and `TextInput` instances we compare to tree elements type with components exported from React Native. This is problematic because exported components are composite ones, while we try to work primarily with host components. It leads to methods like: `isHostElementForType` which try to navigate from host to composite parent of certain type to verify whether this is a host `Text` or `TextInput`.

There is an easy alternative, which is to compare `ReactTestInstance.type` field with string values of `"Text"` and `"TextInput"` which are currently used by these host components. The issue here is that the host types are not "standardised" and have used slightly different names in the past, e.g. `RCTText` instead of `Text`, so these could potentially change again. 

In order to get the best of both worlds: robust code and being future proof I propose migrating to `type` matching against configurable host component names.

### 1. `hostComponentNames` config option
Extend config with `hostComponentNames` field. User will be able to pass custom values in case they use older version of RN (which we probably do not support) or if RN changes the naming in the future without need for release:
```ts
configure({
  hostComponentNames: {
    text: "Text",
    textInput: "TextInput",
  }
})
```
The default values are matching the current RN naming scheme. 

### 2. Update queries and event handling
Modify queries and `fireEvent` code that looks for `Text` and `TextInput` to use `hostComponentNames`:
```ts
element.type === getConfig().hostComponentNames.text;
element.type === getConfig().hostComponentNames.textInput;
```

### (Optional) 3. add `hostComponentNames: "autodetect"` option
Add `autodetect` option to `configure` that would run a simple auto-detection code:
```ts
configure({
  hostComponentNames: "autodetect";
})
```

Simple auto-detection code:
```ts
function autodetectHostComponentNames(): HostComponentNames {
  const { getByTestId } = render(
    <View>
      <Text testID="text">Hello</Text>
      <TextInput tystID="textInput" />
    </View>
  );

  return {
    text: getByTestId("text").props.testID,
    textInput: getByTestId("textInput").props.testID,
  }
``` 
This would allow users to quickly resolve any potential issues regarding host component name changes.

### (Optional) 4. Suggest proper config setting for `hostComponentNames` option
Suggest proper values for `hostComponentNames` option when using auto-detect mode:
```
You are using host component name auto-detection mode. This might cause a slightly longer test execution time.
To make you tests run faster use the following code in your test setup file:

configure({
  hostComponentNames: {
    text: "New_Text",
[...]

Issue #1180, opened 2022-10-17
*ByRole: return only elements with implicit or explicit `accessible={true}` prop


mdjastrzebski:
## Describe the Feature

This is a proposed tweak to `*ByRole` query. It can be treated as either a fix or a (breaking) change.

Currently `getByRole` query is based on explicit role set in `accessibilityRole` prop. Other props are not taken into account (at least until #1064 is merged). This behaviour while basic can be problematic as according to screen readers, elements with `accessible={false}` (either explicit or implicit) are not focusable by screen reader.

This manifests in relatively common RN pattern:
```ts
<Pressable accessibilityRole="button" accessible={false}>
  <Text>Cell title</Text>
  <Pressable accessibilityRole="button">
    <Text>Action</Text>
  </Pressable>
</Pressable>
```

Under current implementation `getByRole('button', { name: "Action" })` would select both outer and inner pressables. After implementing the proposed change it would select only the inner one.

## Possible Implementations
* element can be matched by `getByRole` using the same criteria as now (host component, equal `accessibilityRole`) but also needs to be considered `accessibility element` (i.e. be focusable). 
* each host view can be made focusable by explicitly passing `accessible={true}`
* some host view types are considered focusable without the `accessible` prop (e.g. `Text`, `TextInput`), but can have their "focusability" disabled by setting `accessible={false}`
* basic host `View` is not consider focusable
* some RN built in components, e.g. `Pressable`, `TouchableOpacity` which render to host `View` actually set explicit `accessible={true}` prop on the view

In more practical terms we could use something like `isAccessibilityElement` to check for focusable state:
```ts
export function isAccessibilityElement(
  element: ReactTestInstance | null
): boolean {
  if (element == null) {
    return false;
  }

  if (element.props.accessible !== undefined) {
    return element.props.accessible;
  }

  return (
    isHostElementForType(element, Text) ||
    isHostElementForType(element, TextInput)
  );
}
```

Then this criteria should be added to `*ByRole` query.

This change should not affect other general purpose queries as `*ByText` or `*ByTestId`. It may be worth considering applying it to some other a11y-related queries, which would make only sense if selected view is accessible: `*ByA11yValue`, `*ByA11State`.

## Related Issues
#1152 - initial source for issue
#1163 - related discussion
#1064 - hidden query option

CC: @AugustinLF @pierrezimmermannbam @MattAgn @thymikee

MattAgn:
I'm willing to work on that :)

[...]

PR #1234, opened 2022-11-20
feat(breaking): Return host component for all queries

### LINKED ISSUE #1132: feature: make all (safe) queries return host components
## Describe the Feature

Currently different types of queries can return either host or composite elements:

> Queries returning host components (`typeof node.type === 'string'`):
> * `testId`: host, matched by normalised `testID` prop
> * `role`: host, matched by `accessibilityRole` prop
> * `hintText`: host, matched by `accessibilityHint` prop
> * `labelText`: host, matched by `accessibilityLabel` prop
> * `a11yState`: host, matched by `accessibilityState` prop
> * `a11yValue`: host, matched by `accessibilityValue` prop
> 
> Queries returning composite components:
> * `text`: composite `Text` from RN, matched by normalised and flattened `children` prop
> * `displayValue`: composite `TextInput` from RN, matched by normalised `value` or `defaultValue` prop
> * `placeholderText`: composite `TextInput` from RN, matched by normalised `placeholder` prop
> 
> Queries returning either host or composite components:
> * `UNSAFE_byType`: either host or composite, matched by `element.type` which can be either `string` or component type
> * `UNSAFE_byProps`: either host or composite, matched by `element.props` overlap

The goal of this feature request is to make `text`,  `displayValue` & `placeholderText` return host components. This is probably a breaking change for our users so it should warrant a major release.

## Possible Implementations
Use `component-tree.ts` functions from #1128 to get single host child of found composite `Text` and `TextInput` components from React Native.

## Related Issues
#1092 

@thymikee @AugustinLF @pierrezimmermannbam wdyt?

pierrezimmermannbam:
### Summary

Closes #1132 #1184 

Changes behavior of byText, byDisplayValue and byPlaceholderText queries so that they return host components instead of composite components.
This new behavior is activated by default but can be disabled either globally through the config useLegacyQueries option or per query through the legacy option.

When returning host components, the queries will read the component names from the config. The defaults are 'Text' and 'TextInput'. If a newer RN version changes these names we could support the newer version by changing the default based on the RN version used by the user but for now all RN versions supported by RNTL have the same name for Text and TextInput components. 

Users are able to change the host component names by using the configure API. When a getBy query fails, we detect the true host component names used by RN and if they do not match the ones in the config suggest the user to either update RNTL or modify its config with the names we found.

This is a breaking change so it should be included in some next major release. I haven't done it yet but a migration guide should be written before releasing the next major version. Also the documentation on testing environment will need to be updated as it states that those queries currently return composite components.

### Test plan

I added test cases for each impacted queries with all combinations of options and tested as well that they correctly use the host component names from the config and that they throw the expected error message when they fail and the names used do not match the ones from React Native.


src/__tests__/config.test.ts
```diff
@@ -16,20 +16,24 @@ test('configure() overrides existing config values', () => {
     asyncUtilTimeout: 5000,
     defaultDebugOptions: { message: 'debug message' },
     defaultIncludeHiddenElements: true,
+    useLegacyQueries: false,
```
pierrezimmermannbam, line 15: I'm not really convinced by this option's name but I couldn't find a good one and since it's temporary I thought it's not that important but if you find something better I'll change it

src/queries/options.ts
```diff
@@ -12,3 +13,15 @@ export type TextMatchOptions = {
   exact?: boolean;
   normalizer?: NormalizerFn;
 };
+
+export type LegacyQueryOptions = {
+  legacy?: boolean;
```
pierrezimmermannbam, line 18: Same remark for this option, not a huge fan of the name but again not sure either it really matters since it should be documented in the migration guide

src/helpers/getReactNativeHostComponentNames.tsx
```diff
@@ -0,0 +1,18 @@
...
+export function getReactNativeHostComponentNames(): HostComponentNames {
+  const { getByTestId } = render(
+    <View>
+      <Text testID="text">Hello</Text>
+      <TextInput testID="textInput" />
+    </View>
+  );
+
+  return {
+    text: getByTestId('text').type as string,
```
pierrezimmermannbam, line 15: Maybe we could type the queries to return HostReactTestinstance = ReactTestInstance & { type: string} when they do return host components, it could be pretty convenient to know if they return composite or host just by looking at the type. I'll look into it a bit more

mdjastrzebski:
@pierrezimmermannbam awesome job, I will try to review it at the beginning of the week!

mdjastrzebski (commented):
I was thinking about how can we reconcile semver breaking changes vs moving forward in anticipation of upcoming major release. My idea would be to create a global `configure` option called `allowBreakingChanges` that when set explicitly to true would run code that might introduce breaking changes. Then on major release we would remove all `!allowBreakingChanges` code paths.

This is similar to what you propose in this PR however in a bit more generalised way, that would accommodate other breaking changes we want to make soon like #1180. @pierrezimmermannbam @MattAgn @AugustinLF wdyt?

src/queries/__tests__/displayValue.test.tsx
```diff
@@ -119,3 +119,29 @@ test('byDisplayValue queries support hidden option', () => {
     `"Unable to find an element with displayValue: hidden"`
   );
 });
+
+test('byDisplayValue should return composite component if useLegacyQueries option is true', () => {
```
mdjastrzebski, line 123: Agree that we should test both available behaviours, but probably should treat the legacy one as default one, and the duplicate the existing tests and make necessary changes when the are non-legacy. It might make sense just to duplicate the whole file and add some suffix to it: `displayValue.test.tsx` => `displayValue.test.tsx ` + `displayValue.breaking.test.tsx`. Wdyt? It would be easier to clean up after major release when the breaking change will become the new default.

src/queries/makeQueries.ts
```diff
@@ -83,6 +85,26 @@ Example:
...
+  const reactNativeHostComponentNames = getReactNativeHostComponentNames();
+  const configHostComponentNames = getConfig().hostComponentNames;
+  if (
+    reactNativeHostComponentNames.text !== configHostComponentNames.text ||
+    reactNativeHostComponentNames.textInput !==
+      configHostComponentNames.textInput
+  ) {
+    return `
+
+The wrong host component names are used. This can happen if you use a version of react native that is not compatible with your version of @testing-library/react-native
```
mdjastrzebski, line 98:
```suggestion
You configuration contains invalid host component names. This can happen if you use a version of React Native that is not compatible with your version of @testing-library/react-native
```

src/queries/makeQueries.ts
```diff
@@ -83,6 +85,26 @@ Example:
...
+
+To fix this, either upgrade to a version that supports your version of react native or add the following line in a setup file :
+
+configure({ hostComponentNames: { text: '${reactNativeHostComponentNames.text}', textInput: '${reactNativeHostComponentNames.textInput}' } });`;
+  }
```
mdjastrzebski, line 103:
```suggestion
To fix this add following line in a Jest setup file :

configure({ 
  hostComponentNames: { 
    text: '${reactNativeHostComponentNames.text}',
    textInput: '${reactNativeHostComponentNames.textInput}',
  }
});`;
  }
```

src/helpers/getReactNativeHostComponentNames.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Text, TextInput, View } from 'react-native';
+import { HostComponentNames } from '../config';
+import render from '../render';
+
+export function getReactNativeHostComponentNames(): HostComponentNames {
```
mdjastrzebski, line 6: I think we can use simpler name since we are *React Native* TL ;-)
```suggestion
export function getHostComponentNames(): HostComponentNames {
```

src/helpers/getReactNativeHostComponentNames.tsx
```diff
@@ -0,0 +1,18 @@
...
+  const { getByTestId } = render(
```
mdjastrzebski, line 7: BTW this will mess the current `screen` output, which might be problematic if user is using `screen` + `findBy*` queries which are essientially just `waitFor` + `getBy*`. In order to avoid it we need to expose some `internalRender` function that would skip setting `screen`. You may event want to render using `TestRenderer` and bind queries to the returned `root` using `getQueriesForElement`.

src/helpers/getReactNativeHostComponentNames.tsx
mdjastrzebski: Could we handle this using some actual type checking (`typeof x === "string"`) and not just type casting?

src/fireEvent.ts
```diff
@@ -14,8 +14,7 @@ const isTextInput = (element?: ReactTestInstance) => {
   // We have to test if the element type is either the TextInput component
   // (which would if it is a composite component) or the string
   // TextInput (which would be true if it is a host component)
-  // All queries but the one by testID return composite component and event
-  // if all queries returned host components, since fireEvent bubbles up
+  // All queries return host but since fireEvent bubbles up
```
MattAgn, line 17:
```suggestion
  // All queries return host components but since fireEvent bubbles up
```

src/config.ts
```diff
@@ -9,6 +14,12 @@ export type Config = {
 
   /** Default options for `debug` helper. */
   defaultDebugOptions?: Partial<DebugOptions>;
+
+  /** Option to return composite components */
```
MattAgn, line 18:
```suggestion
  /** Option for byText, byDisplayValue and byPlaceholderText queries to return composite components */
```

src/helpers/matchers/matchTextContent.ts
```diff
@@ -7,9 +7,10 @@ import { getTextContent } from '../getTextContent';
 export function matchTextContent(
   node: ReactTestInstance,
   text: TextMatch,
+  typeToMatch: ElementType | string,
   options: TextMatchOptions = {}
```
MattAgn, line 11: we could maybe use an object for the params here wdyt?

src/queries/__tests__/host-component-names.test.tsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+import { Text, TextInput } from 'react-native';
+import { render, screen, configure } from '../../pure';
+
+describe('host component names', () => {
```
MattAgn, line 5: in this repo, we don't use top level describe except if there is several of them at the root of the file

src/queries/__tests__/host-component-names.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    expect(screen.getByPlaceholderText('email')).toBeTruthy();
+
+    configure({
+      hostComponentNames: { text: 'wrongName', textInput: 'wrongName' },
+    });
+
+    expect(screen.queryByPlaceholderText('email')).toBeFalsy();
+  });
+
+  it('should suggest appropriated config when a getByText query fails and wrong component names are used', () => {
```
MattAgn, line 42:
```suggestion
  test('suggest appropriate config when a getByText query fails and wrong component names are used', () => {
```

src/queries/__tests__/host-component-names.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+    });
+
+    render(<Text>hello</Text>);
+
+    expect(() => screen.getByText('hello')).toThrowErrorMatchingInlineSnapshot(`
+      "Unable to find an element with text: hello
+
+      The wrong host component names are used. This can happen if you use a version of react native that is not compatible with your version of @testing-library/react-native
+
+      To fix this, either upgrade to a version that supports your version of react native or add the following line in a setup file :
```
MattAgn, line 54:
```suggestion
      To fix this, either upgrade to a version that supports your version of react native or add the following line in a setup file:
```

src/queries/makeQueries.ts
```diff
@@ -83,6 +85,26 @@ Example:
...
+  if (
+    reactNativeHostComponentNames.text !== configHostComponentNames.text ||
+    reactNativeHostComponentNames.textInput !==
+      configHostComponentNames.textInput
+  ) {
+    return `
+
+The wrong host component names are used. This can happen if you use a version of react native that is not compatible with your version of @testing-library/react-native
+
+To fix this, either upgrade to a version that supports your version of react native or add the following line in a setup file :
```
MattAgn, line 100:
```suggestion
To fix this, either upgrade to a version that supports your version of react native or add the following line in a setup file:
```

src/queries/__tests__/host-component-names.test.tsx
MattAgn: same comment for the other "it" ;)

src/queries/__tests__/placeholderText.test.tsx
```diff
@@ -78,3 +78,41 @@ test('byPlaceholderText queries support hidden option', () => {
...
+  configure({ useLegacyQueries: true });
+  const { getByPlaceholderText } = render(
+    <TextInput placeholder="placeholder" />
+  );
+
+  expect(getByPlaceholderText('placeholder').type).toBe(TextInput);
+});
+
+test('byPlaceHolderText should return host component if useLegacyQueries option is false', () => {
+  const { getByPlaceholderText } = render(
```
MattAgn, line 92: to not depend on the default option which we're not currently testing, I would specify: 
```suggestion
    configure({ useLegacyQueries: false });
  const { getByPlaceholderText } = render(
```
(same comment for the other queries tests)

src/queries/makeQueries.ts
```diff
@@ -83,6 +85,26 @@ Example:
   return waitForOptions;
 }
 
+function getWrongHostComponentNamesError(): string {
+  const reactNativeHostComponentNames = getReactNativeHostComponentNames();
+  const configHostComponentNames = getConfig().hostComponentNames;
```
MattAgn, line 90:
```suggestion
  const hostComponentNamesConfig = getConfig().hostComponentNames;
```

src/__tests__/config.test.ts
mdjastrzebski: @pierrezimmermannbam pls check this idea how to tackle the breaking change in this and other PRs: #1241

src/queries/options.ts
MattAgn: maybe legacyMode? and you can add some jsdoc to explain the meaning a bit more

src/queries/options.ts
```diff
@@ -12,3 +13,15 @@ export type TextMatchOptions = {
...
+};
+
+export const shouldReturnCompositeComponent = (option: LegacyQueryOptions) => {
+  if (option.legacy === false) {
+    return false;
+  }
+
+  return option.legacy || getConfig().useLegacyQueries;
+};
```
MattAgn, line 27: you can write it like that as well:
```suggestion
export const shouldReturnCompositeComponent = (option: LegacyQueryOptions) => option.legacy ?? getConfig().useLegacyQueries;
```

typings/index.flow.js
```diff
@@ -465,10 +474,17 @@ declare module '@testing-library/react-native' {
...
+    text: string;
+    textInput: string;
+  }
+
   declare interface Config {
     asyncUtilTimeout: number;
     defaultIncludeHiddenElements: boolean;
     defaultDebugOptions?: $Shape<DebugOptions>;
+    useLegacyQueries: boolean;
+    HostComponentNames: HostComponentNames;
```
MattAgn, line 487:
```suggestion
    hostComponentNames: HostComponentNames;
```

website/docs/API.md
```diff
@@ -803,6 +805,15 @@ This option is also available as `defaultHidden` alias for compatibility with [R
...
 Default [debug options](#debug) to be used when calling `debug()`. These default options will be overridden by the ones you specify directly when calling `debug()`.
 
+
+#### `useLegacyQueries` option
+
+`byText`, `byPlaceholderText` and `byDisplayValue` queries used to return composite elements (see the doc on [testing environment](./TestingEnvironment.md)) for more information on composite and host components) but they now return host components. This can break tests if you use the element returned by these queries, for instance the props may be slightly different. To keep the legacy behavior, set this option to true. 
+
+#### `hostComponentNames` option
+
+The internal names of `Text` and `TextInput` components used by `byText`, `byDisplayValue` and `byPlaceholderText` queries. This should only be changed if you're using a version of React Native that is not compatible with the version of @testing-library/react-native. If that's the case, then failed getBy queries should suggest the appropriate config to use.
```
MattAgn, line 815:
```suggestion
The internal names of `Text` and `TextInput` components used by `byText`, `byDisplayValue` and `byPlaceholderText` queries. This should only be changed if you're using a version of React Native that is not compatible with the version of @testing-library/react-native. If that's the case, then failed getBy and findBy queries should suggest the appropriate config to use.
```

src/queries/options.ts
pierrezimmermannbam: I removed this option since we'll make this mandatory from next release

src/queries/__tests__/displayValue.test.tsx
pierrezimmermannbam: Yes I like the idea, i'll change it

src/helpers/getReactNativeHostComponentNames.tsx
pierrezimmermannbam: My intention here was to make it clear that this function returns the host component names of the React native version of the user and not the names from the rntl config for instance. Imo getHostComponentNames is too implicit and it could very well return the names from the config. I think the distinction is important but it seems my naming has been misunderstood so we should find another one. Maybe something like `getRealHostComponentNames` or `getAutodetectedHostComponentNames` ?

src/helpers/getReactNativeHostComponentNames.tsx
pierrezimmermannbam: Oh you're right I hadn't thought of this, good catch ! I like very much the idea of just using TestRenderer and binding queries, if there is further need for an internal render that would skip screen maybe it should get refactored but for now I think it would bring unnecessary complexity

src/config.ts
pierrezimmermannbam: This will be removed as well once the `allowBreakingChanges` flag is available

src/helpers/matchers/matchTextContent.ts
pierrezimmermannbam: Yes absolutely

src/queries/makeQueries.ts
pierrezimmermannbam: This has been updated with maciej proposal

src/helpers/getReactNativeHostComponentNames.tsx
mdjastrzebski: Maybe detectHostComponentNames?

mdjastrzebski: Maybe detectHostComponentNames?

src/queries/__tests__/placeholderText.test.tsx
pierrezimmermannbam: What I do now is set useLegacyQueries option to false in .breaking tests files, is that ok for you ?

src/queries/options.ts
pierrezimmermannbam: It is indeed way more concise ! That part has been updated however since I removed legacy option

src/queries/makeQueries.ts
pierrezimmermannbam: Not sure I like this change, I think of this rather as the names in the config over the config for names so `configHostComponentNames ` is better in that way to me, not sure if you see what I mean

website/docs/API.md
pierrezimmermannbam: I like having components before but I don't see much value in adding the precision on host here, wdyt ?

src/queries/makeQueries.ts
MattAgn: yes i understand, i misread it

website/docs/API.md
MattAgn: Not that important, it's just so that people don't confuse it with composite components which don't have editable names. But then it's written in your paragraph title so I guess it's fine

src/queries/__tests__/displayValue.breaking.test.tsx
```diff
@@ -0,0 +1,129 @@
+import * as React from 'react';
```
MattAgn, line 1: should we document somewhere what the .breaking tests are? it might not be clear for someone else without the context to figure it out. It might be global documentation or specific documentation for each .breaking test not sure

src/queries/__tests__/host-component-names.breaking.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+});
+
+test('does not update screen object when running detection', async () => {
+  const TestComponent = () => {
+    const [shouldDisplayHello, setShouldDisplayHello] = useState(false);
+
+    useEffect(() => {
+      setTimeout(() => {
+        setShouldDisplayHello(true);
+      }, 500);
```
MattAgn, line 129: fake timers are not enabled by default in this repo so your test will take quite some time to run. We could either reduce the time in the timeout and/or active fake timers for this test

src/queries/__tests__/displayValue.breaking.test.tsx
pierrezimmermannbam: Yes that's a very good idea ! We'd need to document as well the useBreakingChange flag and the way we handle breaking changes in general. I'm thinking of a documentation in codebase that would be linked in the contributing doc wdyt ? 

However I'd rather write this doc in another pr since this one is already pretty huge

src/queries/__tests__/host-component-names.breaking.test.tsx
pierrezimmermannbam: Oh good catch ! I've activated fake timers for this test only

src/helpers/getReactNativeHostComponentNames.tsx
pierrezimmermannbam: Updating the queries typing doesn't look trivial at all so for now I just replaced this part with type checking and if we don't correctly detect the host component names then the user won't get a warning

pierrezimmermannbam:
@mdjastrzebski @MattAgn I reworked the pr to use the useBreakingChange flag and treated all your comments

src/queries/__tests__/displayValue.breaking.test.tsx
MattAgn: Great idea!

website/docs/Queries.md
```diff
@@ -127,7 +127,7 @@ getByPlaceholderText(
   options?: {
     exact?: boolean;
     normalizer?: (text: string) => string;
-    includeHiddenElements?: boolean;
+    hidden?: boolean;
```
MattAgn, line 130: why this change? includeHiddenElements is the naming we want to promote. But maybe we should put both in the doc?

MattAgn (approved):

src/helpers/detectHostComponentNames.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Text, TextInput, View } from 'react-native';
+import TestRenderer from 'react-test-renderer';
+import { HostComponentNames } from '../config';
+import { getQueriesForElement } from '../within';
+
+export function detectHostComponentNames():
```
mdjastrzebski, line 7: Wouldn't be actually simpler to throw if we do not find relevant names?

src/helpers/matchers/matchTextContent.ts
```diff
@@ -1,15 +1,23 @@
...
+  options?: TextMatchOptions;
+};
+
+export function matchTextContent({
+  node,
+  text,
+  typeToMatch,
+  options = {},
+}: MatchTextContentParams) {
+  if (!filterNodeByType(node, typeToMatch)) {
```
mdjastrzebski, line 20: Can the matcher get it from `getConfig()`?

src/queries/__tests__/displayValue.breaking.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+  expect(queryByDisplayValue(/custom/i)).toBe(input);
+  expect(queryByDisplayValue('no value')).toBeNull();
+  expect(() => queryByDisplayValue(/fresh/i)).toThrow(
+    'Found multiple elements with display value: /fresh/i'
+  );
+});
+
+test('getByDisplayValue, queryByDisplayValue get element by default value only when value is undefined', () => {
+  const { getByDisplayValue, queryByDisplayValue } = render(<Banana />);
+  expect(() => getByDisplayValue(DEFAULT_INPUT_CHEF)).toThrow();
```
mdjastrzebski, line 56:
```suggestion
  expect(() => getByDisplayValue(DEFAULT_INPUT_CHEF)).toThrowErrorMatchingInlineSnapshot();
```

src/queries/__tests__/displayValue.breaking.test.tsx
```diff
@@ -0,0 +1,130 @@
...
+  expect(queryByDisplayValue(DEFAULT_INPUT_CHEF)).toBeNull();
+
+  expect(() => getByDisplayValue('hello')).toThrow();
```
mdjastrzebski, line 59:
```suggestion
  expect(() => getByDisplayValue('hello')). toThrowErrorMatchingInlineSnapshot();
```

src/queries/__tests__/displayValue.test.tsx
```diff
@@ -119,3 +119,9 @@ test('byDisplayValue queries support hidden option', () => {
     `"Unable to find an element with displayValue: hidden"`
   );
 });
+
+test('byDisplayValue should return composite component', () => {
```
mdjastrzebski, line 123:
```suggestion
test('byDisplayValue should return composite TextInput', () => {
```

src/queries/__tests__/host-component-names.breaking.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+  configure({
+    hostComponentNames: { text: 'wrongName', textInput: 'wrongName' },
+  });
+
+  render(<Text>hello</Text>);
+
+  expect(() => screen.getByText('hello')).toThrowErrorMatchingInlineSnapshot(`
+    "Unable to find an element with text: hello
+
+    Your configuration contains invalid host component names. This can happen if you use a version of React Native that is not compatible with your version of @testing-library/react-native
```
mdjastrzebski, line 54: Wow. Awesome error message 🚀

src/queries/__tests__/text.test.tsx
```diff
@@ -502,3 +502,9 @@ test('byText support hidden option', () => {
     `"Unable to find an element with text: /hidden/i"`
   );
 });
+
+test('byText should return composite component', () => {
```
mdjastrzebski, line 506:
```suggestion
test('byText should return composite Text', () => {
```

src/queries/options.ts
```diff
@@ -12,3 +13,7 @@ export type TextMatchOptions = {
   exact?: boolean;
   normalizer?: NormalizerFn;
 };
+
+export const shouldReturnCompositeComponent = () => {
```
mdjastrzebski, line 17: The naming is a bit confusing, as it might suggest that this fuction is relevant to all queries. While in reality it only applied to some.

Instead of that I would suggest adding following to relevant queries (byText, ...):
`const shouldReturnHostElement = getConfig().useBreakingChanges`

website/docs/API.md
```diff
@@ -803,6 +805,10 @@ This option is also available as `defaultHidden` alias for compatibility with [R
 
 Default [debug options](#debug) to be used when calling `debug()`. These default options will be overridden by the ones you specify directly when calling `debug()`.
 
+#### `hostComponentNames` option
+
+The internal names of the components `Text` and `TextInput` used by `byText`, `byDisplayValue` and `byPlaceholderText` queries. This should only be changed if you're using a version of React Native that is not compatible with the version of @testing-library/react-native. If that's the case, then failed getBy and findBy queries should suggest the appropriate config to use.
```
mdjastrzebski, line 810: How about tweaking it a bit to be less about implemeantion details but more from user's POV

```suggestion
Specifies the host component names RNTL expects React Native to use when rendering some built-in component types (e.g. `Text` and `TextInput`).

You normally won't need to change this option. In the rare occasion when React Native would change the relevant host components names, RNTL will show you a warning suggesting proper values to use.
```

mdjastrzebski (commented):
Looks very good 🔥 Nice to see `useBreakingChanges in action. I've left some minor comments.

Regarding naming I think we should not use `shouldReturnCompositeComponent`, as the returned things are actually elements and not components, but rather `shouldReturnHostElmements` as more accurate. BTW I suggest negating the condition (composite => host) as more precise. This does not however affect `hostComponentNames` as there are indeed are names of components not elements.

Pls also address code coverage warnings of missing coverage.

website/docs/Queries.md
pierrezimmermannbam: Oops unintended change due to rebase, fixed it

jestSetupFileAfterEnv.ts
```diff
@@ -0,0 +1,5 @@
+import { resetToDefaults } from './src/pure';
+
+beforeEach(() => {
```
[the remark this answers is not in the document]
pierrezimmermannbam: It is indeed not needed for every test file but we can't be sure that we'll call it in each test file, this is like a safety net that ensures our tests don't depend on each other. Also I don't see a reason why we wouldn't want to reset between each test so I think it's better if it's global

src/helpers/detectHostComponentNames.tsx
pierrezimmermannbam: I thought about it but I'm not a huge fan of throwing functions, it's not explicit that a function may throw and it may not be handled somewhere and I don't think we'd want the user to see this error so I like better returning null so that typescript forces to handle the case

src/helpers/matchers/matchTextContent.ts
pierrezimmermannbam: This should use `getConfig().hostComponentNames.text` when we allow breaking changes but Text component when it's not the case and this is handled in text.ts file so I feel like it would be redundant to have the same checks in `matchTextContent`. Once we prepare for the next major it will always use `getConfig().hostComponentNames.text` and from this point I think it will be a good refactor but I wouldn't make the change now, wdyt ?

src/queries/options.ts
pierrezimmermannbam: Definitely agree, I made the change

website/docs/API.md
pierrezimmermannbam: Agreed, this is way clearer !

src/helpers/detectHostComponentNames.tsx
MattAgn: One way to show your function is throwing is with JS doc, but I agree it's not as visible as null typing

src/helpers/matchers/matchTextContent.ts
mdjastrzebski: I would rather go back to the previous positional signature for consistency as all matchers use that pattern now. 

re passing text type vs calculating it: I think it would be better to calculate text component type inside this method as well, even at the cost of little duplication. It will make code cleanup on release easier, as refactors will be contained in single files vs tracing signatures across different files.

src/queries/makeQueries.ts
```diff
@@ -128,7 +163,13 @@ export function makeQueries<Predicate, Options>(
       }
 
       if (results.length === 0) {
-        throw new ErrorWithStack(getMissingError(predicate, options), getFn);
+        throw new ErrorWithStack(
+          `${getMissingError(
+            predicate,
+            options
+          )}${getWrongHostComponentNamesError()}`,
```
mdjastrzebski, line 170: I wonder if it would rather be useful to put `getWrongHostComponentNamesError` inside relevant `getMissingError` calls for `*ByText`, etc queries.

src/queries/makeQueries.ts
```diff
@@ -83,6 +85,39 @@ Example:
   return waitForOptions;
 }
 
+function getWrongHostComponentNamesError(): string {
```
mdjastrzebski, line 88: naming, nit: wrong -> incorrect:
```suggestion
function formatIncorrectHostComponentNamesError(): string {
```

mdjastrzebski: BTW in case the error is not relevant I would rather return `null` instead of empty string, as `null` feels like more natural representation of not having error than empty string.

src/helpers/matchers/matchTextContent.ts
pierrezimmermannbam: I made the change so that typeToMatch is computed within matchTextContent

src/queries/makeQueries.ts
pierrezimmermannbam: The reason I did it like this is it looked simpler to do it here because I don't have to call this at several places. It is true that this cannot be related to a failed byTestId query but even if it's not the actual issue users should be aware that their configuration is broken anyway no ? The only downside I see to this approach is running autodetect more often than we need to so from a performance perspective it's not ideal but it only happens on failed tests so it's not very impactful and also it's not very costly

src/helpers/matchers/matchTextContent.ts
```diff
@@ -1,15 +1,25 @@
...
-  options: TextMatchOptions = {}
-) {
-  if (!filterNodeByType(node, Text)) {
+type MatchTextContentParams = {
+  node: ReactTestInstance;
+  text: TextMatch;
+  options?: TextMatchOptions;
+};
+
+export function matchTextContent({
```
mdjastrzebski, line 14: All other matchers user positional arguments. What arguments are they for going with objects as args in this case?

src/queries/text.ts
```diff
@@ -23,22 +24,39 @@ type ByTextOptions = CommonQueryOptions & TextMatchOptions;
...
-    if (!baseInstance) {
-      return [];
+      const results = findAll(
+        baseInstance,
+        (node) => matchTextContent({ node, text, options }),
+        { ...options, matchDeepestOnly: true }
+      );
+
+      return results;
     }
```
mdjastrzebski, line 46:
```suggestion
// vNext version: returns host Text
```

src/queries/text.ts
```diff
@@ -23,22 +24,39 @@ type ByTextOptions = CommonQueryOptions & TextMatchOptions;
 const queryAllByText = (
   instance: ReactTestInstance
 ): ((text: TextMatch, options?: ByTextOptions) => Array<ReactTestInstance>) =>
-  function queryAllByTextFn(text, options) {
-    const baseInstance = isHostElementForType(instance, Text)
-      ? getCompositeParentOfType(instance, Text)
-      : instance;
+  function queryAllByTextFn(text, options = {}) {
+    const shouldReturnHostText = getConfig().useBreakingChanges;
+    if (!shouldReturnHostText) {
```
mdjastrzebski, line 29:
```suggestion
    // Legacy version: return composite Text
    if (!shouldReturnHostText) {
```

jestSetupFileAfterEnv.ts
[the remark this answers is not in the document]
mdjastrzebski: Maybe it our case it makes sense, I've based my comment on that in a typical user app using RNTL this would not make sense as it would override any `configure` changes made in a global Jest setup file.

src/helpers/getReactNativeHostComponentNames.tsx
mdjastrzebski: I mean we could just do something like
```
const textType = getByTestId('text').type;
if (typeof textType !== 'string') {
  throw new Error("Expected to find string type of host Text");
}
```

src/helpers/detectHostComponentNames.tsx
mdjastrzebski: I would insist here on either return `null` instead of `{ text: null; textInput: null }` or throw error to be caught in the calling function. The current solution with `null` fields seems strange.

src/queries/makeQueries.ts
mdjastrzebski: The numer of places is limited to 3 calls sites. I think it would make sense from structural POV to run this check only in relevant queries. Especially that it should be also applied to `getAllBy` variant.

‼️ BTW what is the impact on having it run repeatedly in `findBy*` and `findAllBy`?

mdjastrzebski (commented):
Pls check what is the impact of running the `getBy` checks when users uses `findBy` which wraps `getBy` in `waitFor` so that the check might be run repeatedly on each interval.

src/queries/makeQueries.ts
```diff
@@ -83,6 +85,39 @@ Example:
...
+  if (!getConfig().useBreakingChanges) {
+    return '';
+  }
+
+  const reactNativeHostComponentNames = detectHostComponentNames();
+  if (!reactNativeHostComponentNames.text) {
```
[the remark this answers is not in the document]
pierrezimmermannbam: If changed the implementation to add the error encountered during the auto detection to the error message displayed to the user, it looks like this : 

Capture d’écran 2022-12-09 à 13 38 36

src/helpers/detectHostComponentNames.tsx
pierrezimmermannbam: I'm now returning an error message or the host component names. I threw at first but it's more complicated to manipulate errors since they aren't typed and since I wanted to display the error of the query this seemed to be the best way

src/helpers/matchers/matchTextContent.ts
pierrezimmermannbam: I refactored it when it had typeToMatch arg as well so it was a bit too much arguments for positional arguments. I personally am not a huge fan of positional arguments, I find the calls to be more explicit with objects and also when there are multiple optional params it becomes quite hard to handle with positional arguments

src/helpers/host-component-names.tsx
```diff
@@ -4,9 +4,12 @@ import TestRenderer from 'react-test-renderer';
...
 import { getQueriesForElement } from '../within';
 
+const defaultErrorMessage =
+  'There seems to be an issue with your configuration that prevents the library from working correctly. Please ensure that you have a version of react native compatible with the library';
+
 export function detectHostComponentNames():
-  | HostComponentNames
-  | { text: null; textInput: null } {
+  | (HostComponentNames & { errorMessage?: never })
+  | { errorMessage: string; text?: never; textInput?: never } {
```
mdjastrzebski, line 12: If we want to signal an error while returning error message, the idiomatic JS is throwing error with proper message. 

@pierrezimmermannbam You correctly indicated that thrown errors are not declared in return type (unlike e.g. Swift and Java) but they still are idiomatic way of implementing error handling in JS, and are generally understood by JS devs. Proposed solution is handling error in a type safe way but the tradeoff here is that we are using quite a weird day that decreases readability of the code. 

Since we know our codebase, and we have a wide test coverage that would prevent us from forgetting to handle a thrown error, I think that a better tradeoff is to go with throwing errors.

pierrezimmermannbam:
There is indeed an important impact on findBy performance, I'll try to make it so that we don't run the check until timeout is reached

mdjastrzebski:
@pierrezimmermannbam can we you check what is typical duration for executing `detectHostComponentNames`? This should be pretty staightforward:
```ts
const timeStart = Date.now(); 
detectHostComponentNames()
const timeEnd = Date.now();
const diff = timeEnd - timeStart
```
Maybe run it in a loop 100 or 1000 times and average the result to reduce randomness. If the cost of running this would actually be something like 0-1ms then we could simply run it inside each `render` and would not need to run it for failed assertions. Or even better, we could run do something like `detectHostComponentNamesIfNeeded()` and cache the result if available. 

That way we could actually could get rid of the config option altogether, as RNTL would autodetect these names itself and do not bother the user.

Wdyt @pierrezimmermannbam?

pierrezimmermannbam:
@mdjastrzebski I've done some additional testing and it takes between 100 and 200 ms the first time it is ran and then 0-1 ms. I ran it in a beforeEach on my current project's codebase and there's no noticeable difference, it seems to be cached correctly by default. For a whole test suite it's negligible, for a single test it's not really but it's not much either and it would hide this complexity for the user and also allow to remove lots of code from RNTL codebase that's hurting its maintainability so overall I think it's the better tradeoff, wdyt ?

src/fireEvent.ts
```diff
@@ -14,8 +14,7 @@ const isTextInput = (element?: ReactTestInstance) => {
   // We have to test if the element type is either the TextInput component
   // (which would if it is a composite component) or the string
   // TextInput (which would be true if it is a host component)
-  // All queries but the one by testID return composite component and event
-  // if all queries returned host components, since fireEvent bubbles up
+  // All queries return host components but since fireEvent bubbles up
   // it would trigger the parent prop without the composite component check
   return (
     filterNodeByType(element, TextInput) ||
```
mdjastrzebski, line 20: 1. Shouldn't this PR allow us to reduce this condition to only checking for `filterNodeByType(element, 'TextInput')`
2. Looks like this check should use detected host component name instead of hardcoded `TextInput`.

src/helpers/host-component-names.tsx
```diff
@@ -0,0 +1,52 @@
...
+    );
+    const { getByTestId } = getQueriesForElement(renderer.root);
+    const textHostName = getByTestId('text').type;
+    const textInputHostName = getByTestId('textInput').type;
+
+    if (
+      typeof textHostName !== 'string' ||
+      typeof textInputHostName !== 'string'
+    ) {
+      throw new Error(defaultErrorMessage);
```
mdjastrzebski, line 30: I wonder if we can actually cover this code path. In order to do so, we would have to make `getByTestId` return non-host component. Therefore, seems like we could remove these check from here. @pierrezimmermannbam wdyt?

src/helpers/host-component-names.tsx
```diff
@@ -0,0 +1,52 @@
...
+        textInput: textInputHostName,
+      },
+    });
+  } catch (error) {
+    const errorMessage =
+      error && typeof error === 'object' && 'message' in error
+        ? error.message
+        : null;
+
+    throw new Error(`Trying to detect host component names triggered the following error:
```
mdjastrzebski, line 45: nit: Could we improve formatting of these line by replacing multiline string literal with good old `\n` characters inside string literal or `\n` with string concatenation using `+`. Whichever of these looks cleaner?

src/queries/displayValue.ts
```diff
@@ -24,8 +25,13 @@ const getTextInputNodeByDisplayValue = (
   const { exact, normalizer } = options;
   const nodeValue =
     node.props.value !== undefined ? node.props.value : node.props.defaultValue;
+  const shouldReturnHostTextInput = getConfig().useBreakingChanges;
+  const typeToMatch = shouldReturnHostTextInput
+    ? getConfig().hostComponentNames?.textInput || 'TextInput'
```
mdjastrzebski, line 30: I wonder if we could do something better than hardcoding the default value here? 

Generally at this stage detectHostComponentNames should have been run. So perhaps we should throw an error if not? That should only happen in case when user did not use our `render` but rather bound functions to Test Renderer result using `within`. 

Or perhaps we could call `detectHostNameComponentsIfNeeded`. That would get the host component names even in such edge case scenario.

@pierrezimmermannbam wdyt?

src/render.tsx
```diff
@@ -42,6 +43,8 @@ export default function render<T>(
     });
   }
 
+  detectHostComponentNamesIfNeeded();
```
mdjastrzebski, line 46: I thin we should also call it in case when `unstable_validateStringsRenderedWithinText`. This can be probably solved by moving it above the `if (unstable_validateStringsRenderedWithinText)` on L39.

src/queries/makeQueries.ts
```diff
@@ -151,7 +151,7 @@ export function makeQueries<Predicate, Options>(
   }
 
   function findByQuery(instance: ReactTestInstance) {
-    return function findFn(
+    return async function findFn(
```
mdjastrzebski, line 154: Is this change needed? Does it affects the inferred typing? The inner function does not use `await`.

If the goal here was to clarify that the function returns `Promise` then explicity typing of return type would be more readable.

src/queries/displayValue.ts
```diff
@@ -24,8 +25,13 @@ const getTextInputNodeByDisplayValue = (
   const { exact, normalizer } = options;
   const nodeValue =
     node.props.value !== undefined ? node.props.value : node.props.defaultValue;
+  const shouldReturnHostTextInput = getConfig().useBreakingChanges;
+  const typeToMatch = shouldReturnHostTextInput
```
mdjastrzebski, line 29: readability:
```suggestion
  const textInputType = shouldReturnHostTextInput
```

src/helpers/host-component-names.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { Text, TextInput, View } from 'react-native';
+import TestRenderer from 'react-test-renderer';
+import { configureInternal, getConfig } from '../config';
+import { getQueriesForElement } from '../within';
+
+const defaultErrorMessage = `There seems to be an issue with your configuration that prevents the library from working correctly.
+Please ensure that you have a version of react native compatible with the library`;
```
mdjastrzebski, line 8: wording:
```suggestion
Please check if you are using compatible versions of React Native and React Native Testing Library.`;
```

src/helpers/host-component-names.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import { Text, TextInput, View } from 'react-native';
+import TestRenderer from 'react-test-renderer';
+import { configureInternal, getConfig } from '../config';
+import { getQueriesForElement } from '../within';
+
+const defaultErrorMessage = `There seems to be an issue with your configuration that prevents the library from working correctly.
```
mdjastrzebski, line 7:
```suggestion
const defaultErrorMessage = `There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
```

mdjastrzebski (changes requested):
Awesome refactor @pierrezimmermannbam 🚀

I really like the simplified code structure. I think that the key ingredients of this PR are already done. There are some relatively minor issues that need to be addressed before merging though. Looking forward to getting this key refactor merged which will bring us close to vNext. 🎸

src/fireEvent.ts
pierrezimmermannbam: Yes it should look for the detected host name for host textinput, I fixed this. Regarding removing the check on composite TextInput type I think it should remain since although all queries will now return host elements, fireEvent still triggers props on composite components by bubbling up and if we remove this check then the event will be triggered on the composite component TextInput even if it is non editable

src/helpers/host-component-names.tsx
pierrezimmermannbam: This path shouldn't happen because getByTestId will always return host elements but for now typescript isn't aware of that. It's not easy to change the queries typing right now because some still return composite components so what I had in mind was to update the queries return type to HostReactInstance once we prepare v12 and remove legacy code, does that seem reasonable to you ?

src/queries/displayValue.ts
pierrezimmermannbam: I hadn't found a proper way to deal with this but I just got an idea: what about we use `detectHostNameComponentsIfNeeded ` in the queries directly rather than on render ? That way we don't run autodetection for tests that don't need it and we make sure that it has been called when those queries are called. I just realized that `detectHostNameComponentsIfNeeded ` is called when using renderHook for instance which is something we don't want

src/render.tsx
pierrezimmermannbam: Oh absolutely, I fixed this by moving calls to `detectHostComponentNamesIfNeeded ` to queries

src/queries/makeQueries.ts
pierrezimmermannbam: Nope it's not needed I forgot to revert this change

pierrezimmermannbam: I removed this

src/helpers/host-component-names.tsx
pierrezimmermannbam: You're I used \n it's looks better

src/__tests__/host-component-names.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+    detectHostComponentNamesIfNeeded();
+
+    expect(getConfig().hostComponentNames).toEqual({
+      text: 'banana',
+      textInput: 'banana',
+    });
+  });
+
+  test('throw an error when autodetection throws', () => {
```
mdjastrzebski, line 33:
```suggestion
  test('throw an error when autodetection fails', () => {
```

src/helpers/host-component-names.tsx
mdjastrzebski: Let's just add short comment saying that this code like
```
// this code path should not happen as getByTestId should always return a host element
```

src/helpers/matchers/matchTextContent.ts
mdjastrzebski: Bringing up this point again. If we want to check the positional arguments to JS names arguments, lets do it for all queries for consistency, and in a separate PR.

src/helpers/host-component-names.tsx
```diff
@@ -0,0 +1,54 @@
...
+      hostComponentNames: {
+        text: textHostName,
+        textInput: textInputHostName,
+      },
+    });
+
+    return {
+      text: textHostName,
+      textInput: textInputHostName,
+    };
```
mdjastrzebski, line 44: nit:

```suggestion
    const hostComponentNames = {
      text: textHostName,
      textInput: textInputHostName,
    };
    
    configureInternal({ hostComponentNames });
    return hostComponentNames;
```

src/helpers/host-component-names.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { Text, TextInput, View } from 'react-native';
+import TestRenderer from 'react-test-renderer';
+import { configureInternal, getConfig, HostComponentNames } from '../config';
+import { getQueriesForElement } from '../within';
+
+const defaultErrorMessage = `There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
+Please check if you are using compatible versions of React Native and React Native Testing Library.`;
+
+export function detectHostComponentNamesIfNeeded(): HostComponentNames {
```
mdjastrzebski, line 10: Looking at the call-sites of this function, maybe we can simplify it to:
```suggestion
export function getHostComponentNames(): HostComponentNames {
```

src/fireEvent.ts
```diff
@@ -14,12 +15,14 @@ const isTextInput = (element?: ReactTestInstance) => {
...
-    filterNodeByType(element, 'TextInput')
+    filterNodeByType(
+      element,
+      getConfig().hostComponentNames?.textInput || 'TextInput'
```
mdjastrzebski, line 24: Shouldn't this be using `detectHostComponentsNames`? Otherwise it will fail if, user did not trigger detection using some query.

src/fireEvent.ts
mdjastrzebski: Makes sense.

I think we need to revisit our event handling system and figure out whether we can make it more robust, as there have been some issue with somewhat related code: #1261

mdjastrzebski (commented):
I like the new lazy detection and removing it from `render` fn. The code is almost there, I've left some final comments before merging:
1) Let's move matcher positional vs JS named arguments change to a separate PR.

src/fireEvent.ts
pierrezimmermannbam: It may cause to run detection when it's not needed but there should be relatively few cases where that would happen so it's not really an issue and it makes the code simpler and less bug prone so I think it's a good idea

pierrezimmermannbam:
@mdjastrzebski I made the final modifications, it should be ready to merge!

mdjastrzebski (approved):
Looks good! 🚀

PR #1237, opened 2022-11-22
feature: update react nav example

mdjastrzebski:
### Summary

Update React Navigation example app to use newest features: `includeHiddenElements`, `*ByRole` queries, etc.

### Test plan

All new tests pass.


examples/react-navigation/src/NativeStackNavigator.test.js
```diff
@@ -0,0 +1,30 @@
...
+  expect(
+    screen.queryByRole('header', { name: /Details for item/i })
+  ).toBeFalsy();
+});
+
+test('Pressing an item takes user to the details screen', async () => {
+  renderWithNavigation(<NativeStackNavigator />);
+
+  const item5 = screen.getByRole('button', { name: 'Item 5' });
+  fireEvent(item5, 'press');
```
MattAgn, line 21:
```suggestion
  fireEvent.press(item5);
```

examples/react-navigation/src/StackNavigator.test.js
```diff
@@ -0,0 +1,30 @@
...
+  expect(
+    screen.queryByRole('header', { name: /Details for item/i })
+  ).toBeFalsy();
+});
+
+test('Pressing an item takes user to the details screen', async () => {
+  renderWithNavigation(<StackNavigator />);
+
+  const item5 = screen.getByRole('button', { name: 'Item 5' });
+  fireEvent(item5, 'press');
```
MattAgn, line 21:
```suggestion
  fireEvent.press(item5);
```

examples/react-navigation/src/StackNavigator.test.js
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
+import { screen, fireEvent } from '@testing-library/react-native';
+import { renderWithNavigation } from './test-utils';
+import StackNavigator from './StackNavigator';
+
+test('Home screen contains the header and list of items', async () => {
```
MattAgn, line 6:
```suggestion
test('Home screen contains the header and list of items', () => {
```
there is no await in the test

examples/react-navigation/src/StackNavigator.test.js
```diff
@@ -0,0 +1,30 @@
...
+
+  expect(screen.getByRole('header', { name: 'Home screen' })).toBeTruthy();
+  expect(screen.getAllByRole('button', { name: /Item/ })).toHaveLength(10);
+
+  expect(
+    screen.queryByRole('header', { name: /Details for item/i })
+  ).toBeFalsy();
+});
+
+test('Pressing an item takes user to the details screen', async () => {
```
MattAgn, line 17:
```suggestion
test('Pressing an item takes user to the details screen', () => {
```

examples/react-navigation/src/StackNavigator.test.js
```diff
@@ -0,0 +1,30 @@
...
+  const item5 = screen.getByRole('button', { name: 'Item 5' });
+  fireEvent(item5, 'press');
+
+  expect(
+    screen.getByRole('header', { name: 'Details for Item 5' })
+  ).toBeTruthy();
+  expect(screen.getByText('The number you have chosen is 5.')).toBeTruthy();
+
+  // Home screen is still in the element tree but it is hidden from accessibility
+  expect(screen.queryByRole('header', { name: 'Home screen' })).toBeFalsy();
```
MattAgn, line 29: so nice to be able to do that 🤩

examples/react-navigation/src/screens/HomeScreen.js
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import { Text, View, FlatList, Pressable, StyleSheet } from 'react-native';
+
+const items = new Array(20).fill(null).map((_, idx) => ({
```
MattAgn, line 4: nit: i would rename `idx` into `index` for clarity but not very improtant

examples/react-navigation/jest-setup.js
```diff
@@ -1,2 +1,7 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import { configure } from '@testing-library/react-native';
+
 // Silence the warning: Animated: `useNativeDriver` is not supported because the native animated module is missing
 jest.mock('react-native/Libraries/Animated/NativeAnimatedHelper');
+
+configure({ defaultIncludeHiddenElements: false });
```
thymikee, line 7: worth to mention turning that off in the readme

mdjastrzebski: Done

examples/react-navigation/src/TabNavigator.test.js
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import { screen, fireEvent } from '@testing-library/react-native';
+import { renderWithNavigation } from './test-utils';
+import TabNavigator from './TabNavigator';
+
+test('Changing tabs', () => {
+  renderWithNavigation(<TabNavigator />);
+  expect(screen.getByRole('header', { name: 'Home screen' })).toBeTruthy();
+
+  // Note: React Navigation uses `button` role for tab buttons. Cannocially should be `tab` role.
```
MattAgn, line 10:
```suggestion
  // Note: React Navigation uses `button` role for tab buttons. Canonically should be `tab` role.
```

MattAgn (approved):

examples/react-navigation/README.md
```diff
@@ -0,0 +1,5 @@
+# RNTL example app for React Navigation
+
+This example is shows how to use React Native Testing Library when testing apps using React Navigation.
```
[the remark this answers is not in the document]
MattAgn: btw, i think it would be interesting to have a guide showing the 2 ways of testing navigation (with or without mocking it) wdyt? (and showing the pros and cons of each approach)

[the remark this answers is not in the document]
mdjastrzebski: @MattAgn what do you mean by mocking it?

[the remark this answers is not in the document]
MattAgn: you render your component without the Navigation around it, then you mock `useNavigation` and you check the `navigate` function from useNavigation is called with the proper argument

examples/react-navigation/src/TabNavigator.test.js
```diff
@@ -0,0 +1,17 @@
...
+import { screen, fireEvent } from '@testing-library/react-native';
+import { renderWithNavigation } from './test-utils';
+import TabNavigator from './TabNavigator';
+
+test('Changing tabs', () => {
+  renderWithNavigation(<TabNavigator />);
+  expect(screen.getByRole('header', { name: 'Home screen' })).toBeTruthy();
+
+  // Note: React Navigation uses `button` role for tab buttons as workaround.
+  // It should be actually be `tab` role.
```
MattAgn, line 11:
```suggestion
  // It should actually be `tab` role.
```

MattAgn: I'm not sure what you mean when you say it should be `tab` and not `button`, do you mean compared to what's happening on the native side? and why did React Navigation need a workaround?

examples/react-navigation/README.md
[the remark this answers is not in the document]
mdjastrzebski: I think this is what majority of users are doing, not sure if it worth providing example for, since you mock various `useNavigation` props in different use-cases. Also I would like to encourage our users to leverage the full power of RNTL by building integration tests using the real libraries rather than doing shallow, hook-level, mocking.

What might be worth doing is showing how user can render a single screen using something like custom `renderScreen(ui, navigationParams)` which would leverage real `NavigatorContainer` + e.g. `StackNavigator`. Wdyt @MattAgn?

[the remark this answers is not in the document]
pierrezimmermannbam: What would renderScreen look like ? I'm not a huge fan of creating navigators in tests. Ideally you want to test your production code, writing navigators in tests doesn't test the navigator and also adds more code that needs to be maintained.  What I usually do when I want to test some flow that involves several pages is render the navigator containing the pages I need (can be the whole app or just a StackNavigator) and wrap with a NavigatorContainer if needed. That way I don't have to use any setup function or create false navigators that are equivalent to mocks, wdyt about this approach @mdjastrzebski?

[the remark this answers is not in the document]
mdjastrzebski: I've added that section, pls check the latest `test-utils.js`: `renderScreenWithParams`. 

re testing navigators vs screen: this PR contains both approaches `renderNavigator` renders passed JSX and just wraps it in `NavigatorContainer` while `renderScreenWithParams` allows you to pass route params to a given screen without using mocks, but using test setup involving `NavigatorContainer`, and simple `StackNavigator`. The goal here is to test single screen with different route params. Which might be difficult to do if that screen is buried deep in true navigators.

I think both approaches have their pros and cons, basically easy of test setup vs how well the test conditions resemble real life app. But I think both are still better than mocking `useNavigation` since they both work with the real React Navigation.

@pierrezimmermannbam let me know if that make sense to you or would you rather propose some other solution for single screen testing.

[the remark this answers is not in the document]
pierrezimmermannbam: I'm not sure I entirely understand your approach, how do you use renderScreenWithParams ? If react-navigation is not mocked then calls to navigate will throw if the screen we want to navigate to isn't mounted no ? Is it intended to test route params ?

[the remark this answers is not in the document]
mdjastrzebski: Yes, `navigate` calls will throw unless mocked e.g. with `jest.spyOn(...).mockImplementation(...)`. The main purpose of this kind of tests is to test a single screen that uses route params.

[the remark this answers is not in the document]
pierrezimmermannbam: It's very a interesting approach ! How do you manage to mock the navigation if it's passed down to the screen as a prop ?

On projects I've been working on so far we're using a different approach, namely because we mock globally @react-navigation/native in our tests and only unmock it for tests involving multiple pages. So when we have to test route params we basically render our component and pass the route prop directly to it

```tsx
render(<Page route={getMockRoute(params)} navigation={mockNavigation} />);
```

Where getMockRoute is a test util that returns an object that ressembles what route must look like with the provided params. This allows to easily test calls to mockNavigation methods.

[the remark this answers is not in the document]
mdjastrzebski: @pierrezimmermannbam I wonder is your approach working with components using `useNavigation` to get route params?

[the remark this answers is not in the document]
mdjastrzebski: I think you are right that in my `renderScreenWithParams` approach it's actually not possible to handle `navigation` mocking for `navigation` passed as a prop.

examples/react-navigation/README.md
```diff
@@ -0,0 +1,9 @@
+# RNTL example app for React Navigation
+
+This example shows how to write integration tests using React Navigation without mocking it.
+
+There are two types of tests:
+* integration tests operating on whole navigators, they should use `renderNavigator` helper
```
MattAgn, line 6:
```suggestion
* integration tests operating on whole navigators, they should use `renderNavigator` helper. It is useful when you want to test navigation between pages.
```

examples/react-navigation/src/test-utils.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import * as React from 'react';
+import { NavigationContainer } from '@react-navigation/native';
+import { createStackNavigator } from '@react-navigation/stack';
+import { render } from '@testing-library/react-native';
+
+/**
+ * Render given JSX inside Naviagation container.
```
MattAgn, line 8:
```suggestion
 * Render given JSX inside Navigation container.
```

examples/react-navigation/README.md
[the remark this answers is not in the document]
MattAgn: No @pierrezimmermannbam approach won't handle calls to useNavigation since his Page is not wrapped with a NavigationContainer, providing the context used by `useNavigation`. I feel like each of your approach works for one scenario but not the other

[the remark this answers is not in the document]
MattAgn: Imo, if we have to make a compromise anyway on how much our tests are realistic and can handle different scenarios, we might as well go all the way with @pierrezimmermannbam's approach which will be more performant wdyt?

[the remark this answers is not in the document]
pierrezimmermannbam: It does work with useNavigation since it's globally mocked but if components use useRoute to retrieve route params than it should be mocked as well. A way to to this would be that getRouteMock also mocks useRoute so that it returns the same object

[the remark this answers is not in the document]
mdjastrzebski: Yeah, it's kind of problematic that there is no global `navigation` object that can be easily accessed by testing code but rather it is available only inside components, which makes is hard to apply mocking using `spyOn`.

mdjastrzebski:
@MattAgn @pierrezimmermannbam I've reworked single screen test scenario based on you feedback, pls take a look at [`DetailsScreen.test.ts`](https://github.com/callstack/react-native-testing-library/pull/1237/files#diff-333f680f93f9c9e540cc434d071c773b9c774369a9ffe70706ce0c224d80f7bc) and tell me what do you think.

MattAgn:
Very nice! Maybe we could also add a note somewhere explaining that `useNavigation` and `useRoute` won't work for this test unless they're mocked globally (or is it too obvious?)

mdjastrzebski:
@MattAgn I've actually added mocking for `useNavigation` hook to work with `navigation` prop. Pls take a look at current `DetailsScreen.test.ts`.

examples/react-navigation/src/screens/DetailsScreen.test.js
```diff
@@ -1,23 +1,42 @@
...
+  // Passing both navigation and route to the screen as props
   render(<DetailsScreen navigation={navigation} route={{ params }} />);
 
   expect(
     screen.getByRole('header', { name: 'Details for Item 100' })
   ).toBeTruthy();
   expect(screen.getByText('The number you have chosen is 100.')).toBeTruthy();
 
-  fireEvent.press(screen.getByRole('button', { name: 'Go back' }));
+  // Note: Go Back button get navigation from `useNavigation` hook
```
MattAgn, line 39:
```suggestion
  // Note: Go Back button gets navigation from `useNavigation` hook
```

MattAgn (approved):

MattAgn:
Awesome documentation 😍

pierrezimmermannbam (approved):
Awesome work ! I see that there is a section regarding testing with react navigation on the website, it would be great to update it as well !

thymikee (approved):
very nice. cc @satya164 if you'd like to have a peek — and maybe include this in React Navigation docs

examples/react-navigation/README.md
```diff
@@ -0,0 +1,9 @@
+# RNTL example app for React Navigation
+
+This example shows how to write integration tests using React Navigation without mocking it.
+
+There are two types of tests:
+1. integration tests operating on whole navigators, they should use `renderNavigator` helper to render a navigator component used in the app. It is useful when you want to test a scenario that includes multiple screens.
+2. single screen tests where you would pass mock `navigation` prop, built using `buildNavigationMock()` helper, and `route` prop to the screen component using regular `render` function.
```
satya164, line 7: I wouldn't show this approach since this will only work for very simple cases. For more involved use cases, mocking the navigation object can quickly get complex ad the available methods are different based on the navigator and nesting. There are also other hooks that would fail when not used in a navigator.

Instead, I'd suggest writing components that don't use anything React Navigation specific and instead accept explicit callbacks and props for what they need, so that it's easy to test them in isolation.

mdjastrzebski: @satya164 to understand you correctly, instead of mocking React Nav props/hooks for single screen test, you would rather refactor the code of e.g. `DetailsScreen` (that accepts `route` & `navigation` props) by
* extracting `DetailsScreenContent` that accepts callbacks like `onNavigateBack`, `onNavigateToSettings(...)`, etc
* make `DetailsScreen` a thin wrapper over `DetailsScreenContent` so that it only provides the above callbacks and connects them to navigation actions.

Is that correct or would you do it in some other way?

If that is the case, I think it would be painful for using wanting to test apps using React Nav hooks, since it will require them to refactor each of these components into navigation-less `Content` and navigation `Wrapper` components.

mdjastrzebski: I think testing using navigators is not always feasible, e.g. in case you want to test various scenarios that are on a screen "late" in the stack navigator. I think that in this case user would still be content being able to verify that `navigation.navigate(...)` or other relevant methods were/weren't called with relevant route and params.

@satya164 Would it be possible/feasible to have some mock `TestNavigator` that could that could be used to verify assertion. It could be used in following setup:

```jsx
const Test = createTestNavigator();

test("...", () => {
  render(
    <NavigationContainer>
      <Test.Navigator>
        <Test.Screen component={ScreenUnderTest} initialParams={{ ... }} />
      </Test.Navigator>
    <NavigationContainer>
  );

  fireEvent.press(getByText('Some button'));  
  expect(Test.navigation.navigate).toHaveBeenCalledWith(['Details', { id: 5 }]);
}
```

Such mock would be a full navigator but it would render only a single screen and expose `navigation` object so that user would be able to verify if they were called with proper params.

satya164: There's a bit of a disconnect here, for me, there are 2 areas for testing:

1. Testing screens: Screens can often use navigator-specific methods, rely on navigation lifecycle, navigator-specific behavior, etc. There's no good way to test a screen properly without rendering the navigator. You can mock some stuff, but different screens will need different stuff mocked and in the end, you end up replicating React Navigation's API but it may not have the same behavior. These are integration tests - so we need to test them in an appropriate environment that doesn't behave too differently.

And we're only talking about navigation-related stuff. A screen would usually contain data fetching and other things as well which will need to be mocked.

If we're testing whether `navigation.navigate` is called or not, it's not the right thing to test for. If we want to test that user goes to `Details` screen, then we should check if we're on that screen. The `navigation.navigate` method is an implementation detail - if someone changes its signature or uses a different method like `navigation.jumpTo`, it may do the exact same thing but will break the tests.

2. Testing components: These are unit tests for components that might use React Navigation's API. Mocking is less problematic here, but it's still difficult. One approach would be to write a test navigator (react-navigation can't provide one since all navigators can behave differently and can have different methods). Due to this, I would really recommend dependency injection - pass the callbacks it needs. It'd simplify the tests a lot and ensure that we're only testing the public API of the component, i.e. what the component exposes.

> I think that in this case user would still be content being able to verify that navigation.navigate(...) or other relevant methods were/weren't called with relevant route and params.

People may be content with doing them, but imo it's important to suggest best practices and discourage testing implementation details in official docs. Testing the actual thing is harder, yes, but it also ensures that the tests actually help identify issues and don't fail due to a refactor. I am pretty sure people would still mock and test implementation details, but at least they'd have an idea of what a good test should look like by reading official docs. After all, these are just guidelines, people don't have to do it the exact same way - but they'll get more value from their tests if they do.

I'd also bring up the guiding principles of testing library which are in-line with how I see this: [The more your tests resemble the way your software is used, the more confidence they can give you.](https://testing-library.com/docs/guiding-principles) and Avoid testing [implementation details](https://kentcdodds.com/blog/testing-implementation-details).

PR #1243, opened 2022-11-27
fix: typing of `act`

### LINKED ISSUE #1207: Typing of `act` broken
## Describe the bug

The `act` function exported by `@testing-library/react-native` is always of type `(callback: () => void) => void`. However, the underlying function coming from react-test-renderer also has overloads for promises which are incorrectly suppressed.

This error in typing leads to linter errors when one tries to pass promise-returning callbacks to `act` or tries to await a promise-returning `act`.

## Expected behavior

The `act` function should have the same typing as the underlying function from react-test-renderer.

## Steps to Reproduce

The following snippet will produce linter errors such as [this one](https://typescript-eslint.io/rules/no-misused-promises/), depending on one's linter configuration.
```
import { act } from '@testing-library/react-native'

await act(async () => {
  await Promise.resolve();
});
```

## Versions

    @testing-library/react-native: ^11.3.0 => 11.3.0 
    react: 18.0.0 => 18.0.0 
    react-native: 0.69.6 => 0.69.6 

mstykow:
### Summary

- cast customized act function back to react-test-renderer type
- flow types can be ignored since they already use `Thenable`

Fixes #1207

### DIFF
--- a/src/act.ts
+++ b/src/act.ts
@@ -86,7 +86,7 @@ const getAct = () => {
     ? withGlobalActEnvironment(reactTestRendererAct)
     : reactTestRendererAct;
 };
-const act = getAct();
+const act = getAct() as Act;
 
 export default act;
 export {


mdjastrzebski:
@mstykow It seems like your PR just force casts `getAct()` result as `Act`. But when refactored as `const act: Act = getAct();` it actually displays type errors.

Could you adjust your PR to work the actual types to make it work with the types instead of force casting? The complex part seems to be adjusting `withGlobalActEnvironment` wrapper.

mstykow:
> @mstykow It seems like your PR just force casts `getAct()` result as `Act`. But when refactored as `const act: Act = getAct();` it actually displays type errors.
> 
> Could you adjust your PR to work the actual types to make it work with the types instead of force casting? The complex part seems to be adjusting `withGlobalActEnvironment` wrapper.

@mdjastrzebski I tried that at first but this wrapper is so poorly written (full of typecheck and lint disablers) that instead of fixing it we'd have to completely re-write it, and that's not a time investment I'm willing to make when a simple recast does the job.

Feel free to reject my PR if you don't want this quick fix.

mdjastrzebski:
I am not convinced that updated typing of the `act` is correct as it's done overriding the type checking instead of using it verify the types. Therefore I'll be closing this PR in the current form. If you need to use improved typing in your code base you can re-cast exported `act` or use Patch Package tool.

PR #1244, opened 2022-11-28
byRole queries return only accessibility elements

MattAgn:
### Summary

Solves #1180. It makes byRole queries return only implicit or explicit accessibility elements. Explicit accessibility elements are those with `accessibility` prop set to true. Implicit accessibility elements are `Text`, `TextInput`, `Pressable` and `Switch` (tested with VoiceOver for `Switch` to check it was indeed accessible even without setting the accessible prop and I checked through a snapshot, behing the scene, Switch resolves to a host component RCTSwitch).

### Test plan
- [x] Unit tests for `isAccessibilityElement` helper
- [x] One test to check that byRole only returns accessibility elements

### DIFF
--- a/src/helpers/__tests__/accessiblity.test.tsx
+++ b/src/helpers/__tests__/accessiblity.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { View, Text, TextInput, Pressable } from 'react-native';
+import { View, Text, TextInput, Pressable, Switch } from 'react-native';
 import { render, isHiddenFromAccessibility, isInaccessible } from '../..';
 import { isAccessibilityElement } from '../accessiblity';
 
@@ -199,6 +199,11 @@ describe('isAccessibilityElement', () => {
       const element = render(<Pressable testID="pressable" />).getByTestId(
         'pressable'
       );
+      expect(isAccessibilityElement(element)).toEqual(true);
+    });
+
+    test('returns true for Switch component', () => {
+      const element = render(<Switch testID="switch" />).getByTestId('switch');
       expect(isAccessibilityElement(element)).toEqual(true);
     });
 
--- a/src/helpers/__tests__/accessiblity.test.tsx
+++ b/src/helpers/__tests__/accessiblity.test.tsx
@@ -2,174 +2,180 @@ import React from 'react';
 import { View, Text, TextInput } from 'react-native';
 import { render, isHiddenFromAccessibility, isInaccessible } from '../..';
 
-test('returns false for accessible elements', () => {
-  expect(
-    isHiddenFromAccessibility(
-      render(<View testID="subject" />).getByTestId('subject')
-    )
-  ).toBe(false);
+describe('isHiddenFromAccessibility', () => {
+  test('returns false for accessible elements', () => {
+    expect(
+      isHiddenFromAccessibility(
+        render(<View testID="subject" />).getByTestId('subject')
+      )
+    ).toBe(false);
 
-  expect(
-    isHiddenFromAccessibility(
-      render(<Text testID="subject">Hello</Text>).getByTestId('subject')
-    )
-  ).toBe(false);
+    expect(
+      isHiddenFromAccessibility(
+        render(<Text testID="subject">Hello</Text>).getByTestId('subject')
+      )
+    ).toBe(false);
 
-  expect(
-    isHiddenFromAccessibility(
-      render(<TextInput testID="subject" />).getByTestId('subject')
-    )
-  ).toBe(false);
-});
+    expect(
+      isHiddenFromAccessibility(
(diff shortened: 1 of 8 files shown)


src/helpers/__tests__/accessiblity.test.tsx
```diff
@@ -1,175 +1,237 @@
...
+});
+
+describe('isAccessibilityElement', () => {
+  describe('when accessible prop is not defined', () => {
+    test('returns true for Text component', () => {
+      const element = render(<Text testID="text">Hey</Text>).getByTestId(
+        'text'
+      );
+      expect(isAccessibilityElement(element)).toEqual(true);
+    });
```
mdjastrzebski, line 191: How about structuring the tests a bit different form, by grouping default/true/false tests for given component type together. These tests will be both exhaustive and take less LOCs.

```suggestion
  test('TextInput matching', () => {
    const { getByTestId } = render(<View>
      <Text testID="default">Default</Text>
      <Text testID="true">True</Text>
      <Text testID="false">False</Text>
    </View>);
    expect(isAccessibilityElement(getByTestId("default"))).toBeTrue();
    expect(isAccessibilityElement(getByTestId("true"))).toBeTrue();
    expect(isAccessibilityElement(getByTestId("false"))).toBeFalse();
  });
```

mdjastrzebski: We could have such tests for key views that render to host views: View, Text, TextInput, Switch.

Maybe we can also add tests for composite components that render to View with accessible prop: Pressable, TouchableOpacity.

src/queries/role.ts
```diff
@@ -63,8 +64,9 @@ const queryAllByRole = (
     return findAll(
       instance,
       (node) =>
-        // run the cheapest checks first, and early exit too avoid unneeded computations
+        // run the cheapest checks first, and early exit to avoid unneeded computations
         typeof node.type === 'string' &&
+        isAccessibilityElement(node) &&
```
mdjastrzebski, line 69: This would be a breaking change so we should apply it only when `useBreakingChanges` config is set.

mdjastrzebski (changes requested):
So far looks quite good. Good catch with `Switch` component! This PR should use `useBreakingChanges` config option so that we will be able to turn it on for vNext.

src/helpers/__tests__/accessiblity.test.tsx
MattAgn: good idea! the only downside of this method is that tests title are less explicit

MattAgn: Now : CleanShot 2023-01-17 at 11 57 10
Before : 
CleanShot 2023-01-17 at 11 58 26

MattAgn: what do you prefer? both are fine for me

src/queries/__tests__/role.test.tsx
```diff
@@ -733,3 +733,12 @@ test('byRole queries support hidden option', () => {
     `"Unable to find an element with role: "button""`
   );
 });
+
+test('takes accessible prop into account', () => {
```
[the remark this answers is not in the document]
MattAgn: done

MattAgn:
@mdjastrzebski thanks for the review! I've fixed everything you mentioned

src/queries/role.ts
```diff
@@ -60,11 +62,15 @@ const queryAllByRole = (
   instance: ReactTestInstance
 ): ((role: TextMatch, options?: ByRoleOptions) => Array<ReactTestInstance>) =>
   function queryAllByRoleFn(role, options) {
+    const shouldMatchOnlyAccessibilityElements = (node: ReactTestInstance) =>
+      !getConfig().useBreakingChanges || isAccessibilityElement(node);
```
mdjastrzebski:
```suggestion
    const shouldMatchOnlyAccessibilityElements = getConfig().useBreakingChanges;

File: src/queries/role.ts [hunk 5]
```diff
@@ -60,11 +62,15 @@ const queryAllByRole = (
...
+
     return findAll(
       instance,
       (node) =>
-        // run the cheapest checks first, and early exit too avoid unneeded computations
+        // run the cheapest checks first, and early exit to avoid unneeded computations
         typeof node.type === 'string' &&
+        shouldMatchOnlyAccessibilityElements(node) &&
```
mdjastrzebski:
```suggestion
        (!shouldMatchOnlyAccessibilityElements || isAccessibilityElement(node) &&
```

mdjastrzebski: This will get simplified in vNext.

mdjastrzebski (approved):
All good. Let's merge this 🔥

PR #1258, opened 2022-12-13
[WIP] feature: fireEvent default event object

### LINKED ISSUE #714: feature: `fireEvent` should provide generated event object
## Describe the Feature

If I understand the behavior correctly, fireEven looks for properties like "onPress" / "onScroll" in the DOM and calls these functions, but **does not create an event object.** This is a strong difference from the actual behavior of the application.

I understand the potential difficulties in creating these objects in nodejs environment, but it is possible for example to create objects at least partially similar to real events. It's also possible to document which event properties are relevant and which are not.

Even the very fact that a callback will receive an object, will be closer to real situations.

## Possible Implementations
- Generate an event object and pass it to the fired event.
- Fill the object with data from this event. All fields should be presented as in real event.
- Fields for which no valid values can be created must be populated with fake values and documented

## Problem example

```tsx
// short example component 
interface Props {
    /* Button's value to return */
    value?: any;
    /* Press callback with passed event and value */
    onPress?: (value: any, event: GestureResponderEvent) => void;
}

const Button: FunctionComponent<Props> = ({ value, onPress }) => {
    const onButtonPress = useCallback(
        (event: GestureResponderEvent): void => {
            onPress?.(value, event);
        },
        [onPress, value],
    );

    return <TouchableWithoutFeedback onPress={onButtonPress} />
};

// short tests example
fireEvent(button, 'press');
expect(typeof onPress.mock.calls[0][1]).toBe(typeof 'object');
```

- The last test will fail because the event is undefined

mdjastrzebski:
### Summary

Resolves #714 

### Test plan

TODO

### DIFF
--- a/src/__tests__/fireEvent.test.tsx
+++ b/src/__tests__/fireEvent.test.tsx
@@ -107,19 +107,30 @@ describe('fireEvent', () => {
 test('fireEvent.press', () => {
   const onPressMock = jest.fn();
   const text = 'Fireevent press';
-  const eventData = {
+  const event = {
     nativeEvent: {
       pageX: 20,
       pageY: 30,
     },
   };
+
   const { getByText } = render(
     <OnPressComponent onPress={onPressMock} text={text} />
   );
 
-  fireEvent.press(getByText(text), eventData);
+  fireEvent.press(getByText(text), event);
+  expect(onPressMock).toHaveBeenCalledWith(event);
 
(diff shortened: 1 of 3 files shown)


mdjastrzebski:
Closing as stale. Subject will be addressed in `UserEvent` PR (#1405, #1386 #1396), and can be back-ported to `fireEvent` if  needed.

Issue #1209, opened 2022-11-03
RFC: deprecate some a11y queries


mdjastrzebski:
## Problem

Some of our a11y queries are underwhelming:

- `getByA11yState({ disabled: true })` - does not give any control over which control we are trying to find. Can be problematic if there are 2+ controls with matching state, we also might accidentally match another disabled control leading to false negative result.
- `getByA11yValue({ min: 100 })` - similar issues

Note: these queries have no counterpart in RTL.

Other queries are somewhat more useful, but still less powerful than `getByRole`:

- `getByLabelText`
- `getByHintText`

Users are generally better server by using `getByRole` with `name` and state props like `disabled`, `selected`, etc. 

Note: this queries have counterpart in the RTL world (LabelText) or have similar semantic (HintText).

## Proposed solution

1. Add a11y value & hint props support to `getByRole` in order to make it more versatile
2. Deprecate `*ByA11yState` and `*ByA11yValue` queries, and display message to use `getByRole` with appropriate parameters.
3. Keep `*ByLabelText` and `*ByHintText` text as lightweight queries, but mention in the docs that we rather suggest using `getByRole` with proper additional attributes.

## Impact
Soft deprecating, i.e. displaying deprecating warning to console, is not a breaking change, so we might do with minor/feature update.

We then should remove `*ByA11yState` and `*ByA11yValue` in the next or next^2 release.

pierrezimmermannbam:
I think this is a really good idea ! However I wonder if the `byLabelText` and `byA11yValue` make sense on their own. It makes sense to keep them if they have a counterpart in RTL, but I'm not sure about keeping queries, not mark them as deprecated and only indicate in documentation that we recommend using getByRole. Most users won't read the documentation so if we really think that using `byRole` queries is better (which I'm not sure of but I'm not an expert when it comes to accessibility), I wonder if these queries shouldn't be deprecated as well

thymikee:
If we want to discourage usage, a good start would be to leverage that most of the users use TS, and use the deprecated comment. 

To further improve adoption we could create a codemod, ideally applicable through npx so it's a one-liner.

mdjastrzebski:
@thymikee I think there might be problem with both ideas:
1. @deprecated annotation would be easy for explicit function definitions but we have `makeQueries` and also do query binding to root instance, hence all function information is inferred.
2. Codemode will not be possible, as user will have to figure out relevant role values for each affected query.

[...]

PR #1269, opened 2022-12-30
fix: add deprecation notice for *ByA11yState and *ByA11yValue queries

mdjastrzebski:
### Summary

Related to #1209 

### Test plan

Read the relevant changes in the docs.

### DIFF
--- a/website/docs/Queries.md
+++ b/website/docs/Queries.md
@@ -369,7 +369,7 @@ The difference in handling default values is made to reflect observed accessibil
 ### `ByA11Value`, `ByAccessibilityValue`
 
 :::caution
-This query uses a predicate that is typically to general to give meaningful results. Therefore, it's better to use one of following options:
+This query uses a predicate that is typically too general to give meaningful results. Therefore, it's better to use one of the following options:
 * use one of [`ByRole`](#byrole) queries with `value` option
 * use [`toHaveAccessibilityValue()`](https://github.com/testing-library/jest-native#tohaveaccessibilityvalue) matcher to check the state of element found using some other query
 :::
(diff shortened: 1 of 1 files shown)


website/docs/Queries.md
```diff
@@ -305,6 +305,12 @@ const element3 = screen.getByRole('button', { name: "Hello", disabled: true });
 
 ### `ByA11yState`, `ByAccessibilityState`
 
+:::caution
+This query uses a predicate that is typically to general to give meaningful results. Therefore, it's better to use one of following options:
```
MattAgn, line 309:
```suggestion
This query uses a predicate that is typically to general too give meaningful results. Therefore, it's better to use one of following options:
```

website/docs/Queries.md
```diff
@@ -362,6 +368,13 @@ The difference in handling default values is made to reflect observed accessibil
 
 ### `ByA11Value`, `ByAccessibilityValue`
 
+:::caution
+This query uses a predicate that is typically to general to give meaningful results. Therefore, it's better to use one of following options:
```
MattAgn, line 372:
```suggestion
This query uses a predicate that is typically too general to give meaningful results. Therefore, it's better to use one of the following options:
```

MattAgn (approved):

website/docs/Queries.md
```diff
@@ -305,6 +305,12 @@ const element3 = screen.getByRole('button', { name: "Hello", disabled: true });
 
 ### `ByA11yState`, `ByAccessibilityState`
 
+:::caution
+This query uses a predicate that is typically to general too give meaningful results. Therefore, it's better to use one of the following options:
```
mdjastrzebski, line 309:
```suggestion
This query uses a predicate that is typically too general to give meaningful results. Therefore, it's better to use one of the following options:
```

PR #1293, opened 2023-01-26
fix: force cast await to proper type

### LINKED ISSUE #1207: Typing of `act` broken

### LINKED ISSUE #1276: Typescript act() got warning `'await' has no effect on the type of this expression.`
## Describe the bug

when i use act with typescript , got warning like 👇

image
image

## Steps to Reproduce

`@testing-library/react-native/build/act.d.ts`

```diff
declare global {
    var IS_REACT_ACT_ENVIRONMENT: boolean | undefined;
}
declare function setIsReactActEnvironment(isReactActEnvironment: boolean | undefined): void;
declare function getIsReactActEnvironment(): boolean | undefined;
- declare const act: (callback: () => void) => void;
+ declare const act: <T>(callback: () => T) => T;
export default act;
export { setIsReactActEnvironment as setReactActEnvironment, getIsReactActEnvironment, };

```

## Versions

image

mdjastrzebski:
### Summary

Resolve typing issues with `act` by allowing to invoke both `act()` and `await act()`. React `act` has uses function overloading which is hard to work with, so I've resorted to casting.

Dropping support for mock `act` and increasing peer dep req to [React 16.8](https://github.com/facebook/react/releases/tag/v16.8.0) (earliest version to expose act in Test Renderer). I think we can avoid calling this a breaking change, as we already specified minimal RN version of 0.59, which uses React 16.8.3 ([see here](https://docs.google.com/presentation/d/1Z8mNkRw5CJ93qSQqESrcXbgia1dA8V1olnpKT0lsygk/edit#slide=id.gdc26277501_0_39))

Resolves #1276 
Resolves #1207

### Test plan

Typecheck runs fine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
   },
   "peerDependencies": {
     "jest": ">=28.0.0",
-    "react": ">=16.0.0",
+    "react": ">=16.8.0",
     "react-native": ">=0.59",
     "react-test-renderer": ">=16.8.0"
   },
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "jest": ">=28.0.0",
     "react": ">=16.0.0",
     "react-native": ">=0.59",
-    "react-test-renderer": ">=16.0.0"
+    "react-test-renderer": ">=16.8.0"
   },
   "peerDependenciesMeta": {
     "jest": {
--- a/src/__tests__/act.test.tsx
+++ b/src/__tests__/act.test.tsx
@@ -56,7 +56,7 @@ test('should act even if there is no act in react-test-renderer', () => {
 
 test('should be able to not await act', async () => {
   const result = act(() => {});
-  expect(result).toBe(undefined);
+  expect(result).toHaveProperty('then');
 });
 
 test('should be able to await act', async () => {
--- a/src/__tests__/act.test.tsx
+++ b/src/__tests__/act.test.tsx
@@ -43,7 +43,7 @@ test('fireEvent should trigger useState', () => {
 });
 
 test('should act even if there is no act in react-test-renderer', () => {
-  // @ts-ignore
+  // @ts-expect-error
   ReactTestRenderer.act = undefined;
   const callback = jest.fn();
 
@@ -52,4 +52,14 @@ test('should act even if there is no act in react-test-renderer', () => {
   });
 
   expect(callback).toHaveBeenCalled();
+});
+
+test('should be able to not await act', async () => {
+  const result = act(() => {});
+  expect(result).toBe(undefined);
+});
+
+test('should be able to await act', async () => {
+  const result = await act(async () => {});
+  expect(result).toBe(undefined);
 });
--- a/src/__tests__/act.test.tsx
+++ b/src/__tests__/act.test.tsx
@@ -42,18 +42,6 @@ test('fireEvent should trigger useState', () => {
   expect(counter.props.children).toEqual('Total count: 1');
 });
 
-test('should act even if there is no act in react-test-renderer', () => {
-  // @ts-expect-error
-  ReactTestRenderer.act = undefined;
-  const callback = jest.fn();
-
-  act(() => {
-    callback();
-  });
-
-  expect(callback).toHaveBeenCalled();
-});
-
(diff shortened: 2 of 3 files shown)


src/act.ts
```diff
@@ -77,6 +74,14 @@ function withGlobalActEnvironment(actImplementation: Act) {
     }
   };
 }
+
+type MockAct = (callback: () => void) => void;
+type Act = typeof reactTestRendererAct extends undefined ? MockAct : ReactAct;
+
+const actMock = (callback: () => void) => {
```
mdjastrzebski, line 81: @thymikee do we still need to mock `act` in case RTR does not expose it? Seems like it was added in 2019. https://github.com/facebook/react/commit/267ed981466ef323b1461b013b5db8db31e31772

thymikee: Yeah let's drop it

mdjastrzebski: Dropping support and increasing peer dep req to React 16.8 (earlierst version to expose `act` in Test Renderer). I think we can avoid calling this a breaking change, as we already specified minimal RN version of 0.59, which uses React 16.8.3 ([see here](https://docs.google.com/presentation/d/1Z8mNkRw5CJ93qSQqESrcXbgia1dA8V1olnpKT0lsygk/edit#slide=id.gdc26277501_0_39))

AugustinLF: Sounds fair to me, I'd still suggest being explicit about it in the release notes!

src/__tests__/act.test.tsx
```diff
@@ -43,7 +43,7 @@ test('fireEvent should trigger useState', () => {
 });
 
 test('should act even if there is no act in react-test-renderer', () => {
```
AugustinLF, line 45: I'd assume this test is not needed anymore.

src/act.ts
```diff
@@ -20,10 +18,8 @@ function getIsReactActEnvironment() {
   return globalThis.IS_REACT_ACT_ENVIRONMENT;
 }
 
-type Act = typeof reactTestRendererAct;
-
-function withGlobalActEnvironment(actImplementation: Act) {
-  return (callback: Parameters<Act>[0]) => {
+function withGlobalActEnvironment(actImplementation: ReactAct) {
```
AugustinLF, line 21: Is it even useful to pass it `actImplementation` as a param?

mdjastrzebski: hmmmm, good question, we've got our code from RTL and just sprinkled types and eslint ignores over it. 
I guess it's easier to keep it in sync with their implmentation.

src/__tests__/act.test.tsx
mdjastrzebski: Good catch, I've removed that tests. Funny thing is that it did not fail despite no longer supporting missing `act`.

AugustinLF (approved):

package.json
```diff
@@ -60,7 +60,7 @@
     "jest": ">=28.0.0",
     "react": ">=16.0.0",
     "react-native": ">=0.59",
-    "react-test-renderer": ">=16.0.0"
+    "react-test-renderer": ">=16.8.0"
```
thymikee, line 63: hm, React and React Test Renderer should go in sync to avoid weird issues

thymikee (approved):
Left one comment, other than that LGTM

trajano (commented):
lgtm

PR #1298, opened 2023-01-27
refactor(breaking): rename `container` to `UNSAFE_root`; introduce `root` host element

### LINKED ISSUE #1296: RFC: re-write `container` API
## Summary

Current `container` API has following issues:
* it returns composite element, while we try to avoid return such components
* it is not behaving the same as RTL API of the same name, as RTL returns an element containing the rendered elements

Some users have been trying to use it in their component tests in a following way:
```ts
test('renders correctly', () => {
  const { container } = render(<Test />);
  expect(container).toBeEmptyElement()  // this should fail, but passes
});
```

## Suggested solution

1. Rename `container` to `UNSAFE_root` to allow easy migration for users that might have valid uses for `container` API.
  * Potential valid uses seem to be custom queries atm.
  * Make `container` log deprecation message pointing to `UNSAFE_root`

3. Create a new `root` return from `render` (+ `screen.root`) that would expose root host component. This will allow to simplify writing component test, eg.:
```ts
test('renders correctly', () => {
  render(<Test />);
  expect(screen.root).toBeEmptyElement()
});
```

## Related Issues

This should be released as part of v12.

* Original discussion: https://github.com/callstack/react-native-testing-library/discussions/1289
* Issue that triggered the dicussion https://github.com/testing-library/jest-native/issues/141

mdjastrzebski:
### Summary

This PR contains breaking changes behind `useBreakingChanges` feature flag.

Resolves #1296 

1. Rename `container` to `UNSAFE_root`
2. Display deprecation messages for `container` (behind `useBreakingChanges` )
3. Expose `root` API that returns root host element.

### Test plan

All tests pass.

### DIFF
--- a/src/__tests__/__snapshots__/render.breaking.test.tsx.snap
+++ b/src/__tests__/__snapshots__/render.breaking.test.tsx.snap
@@ -1,0 +1,39 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`toJSON 1`] = `
+<View
+  accessibilityState={
+    {
+      "busy": undefined,
+      "checked": undefined,
+      "disabled": undefined,
+      "expanded": undefined,
+      "selected": undefined,
+    }
+  }
+  accessibilityValue={
+    {
+      "max": undefined,
+      "min": undefined,
+      "now": undefined,
+      "text": undefined,
+    }
+  }
+  accessible={true}
+  collapsable={false}
+  focusable={true}
+  onBlur={[Function]}
+  onClick={[Function]}
+  onFocus={[Function]}
+  onResponderGrant={[Function]}
+  onResponderMove={[Function]}
+  onResponderRelease={[Function]}
+  onResponderTerminate={[Function]}
+  onResponderTerminationRequest={[Function]}
+  onStartShouldSetResponder={[Function]}
+>
+  <Text>
+    press me
+  </Text>
+</View>
+`;
--- a/src/__tests__/__snapshots__/render.breaking.test.tsx.snap
+++ b/src/__tests__/__snapshots__/render.breaking.test.tsx.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`toJSON 1`] = `
+exports[`toJSON renders host output 1`] = `
 <View
   accessibilityState={
     {
--- a/src/__tests__/render.breaking.test.tsx
+++ b/src/__tests__/render.breaking.test.tsx
@@ -219,7 +219,7 @@ test('returns host root', () => {
   expect(root.props.testID).toBe('inner');
 });
 
-test('returns UNSAFE_root', () => {
+test('returns composite UNSAFE_root', () => {
   const { UNSAFE_root } = render(<View testID="inner" />);
 
   expect(UNSAFE_root).toBeDefined();
--- a/src/__tests__/render.breaking.test.tsx
+++ b/src/__tests__/render.breaking.test.tsx
@@ -1,7 +1,14 @@
 /* eslint-disable no-console */
 import * as React from 'react';
 import { View, Text, TextInput, Pressable, SafeAreaView } from 'react-native';
-import { render, fireEvent, RenderAPI } from '..';
+import { render, screen, fireEvent, RenderAPI } from '..';
+import { configureInternal } from '../config';
+
+type ConsoleLogMock = jest.Mock<typeof console.log>;
+
+beforeEach(() => {
+  configureInternal({ useBreakingChanges: true });
+});
 
 const PLACEHOLDER_FRESHNESS = 'Add custom freshness';
 const PLACEHOLDER_CHEF = 'Who inspected freshness?';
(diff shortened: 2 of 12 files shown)


src/__tests__/__snapshots__/render.breaking.test.tsx.snap
```diff
@@ -0,0 +1,39 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`toJSON 1`] = `
```
thymikee, line 3: let's make the description a bit more descriptive, I don't know what this tests 😅

src/__tests__/render.breaking.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+  jest.spyOn(console, 'warn').mockImplementation(() => {});
+  const mockCalls = (console.warn as ConsoleLogMock).mock.calls;
+  const view = render(<View testID="inner" />);
+
+  expect(() => view.container).toThrowErrorMatchingInlineSnapshot(
+    `"'container' property has been renamed to 'UNSAFE_root'"`
+  );
+  expect(mockCalls[0][0]).toMatchInlineSnapshot(
+    `"'container' property has been renamed to 'UNSAFE_root'"`
+  );
```
thymikee, line 240: pick one, these are redundant

src/render.tsx
```diff
@@ -100,10 +101,22 @@ function buildRenderResult(
...
+    },
+    UNSAFE_root: instance,
+    get container() {
+      if (!getConfig().useBreakingChanges) {
+        return instance;
+      }
+
+      // eslint-disable-next-line no-console
+      console.warn(`'container' property has been renamed to 'UNSAFE_root'`);
+      throw new Error("'container' property has been renamed to 'UNSAFE_root'");
```
thymikee, line 118: Why warn and error at the same time? Error is descriptive enough imho

src/screen.ts
```diff
@@ -16,6 +16,12 @@ const defaultScreen: RenderResult = {
   get container(): ReactTestInstance {
     throw new Error(SCREEN_ERROR);
   },
+  get root(): ReactTestInstance {
```
thymikee, line 19: Is this type accurate? Isn't there a narrower type for host elements?

mdjastrzebski: unfortunately no, though we might think about something like that in the future

src/render.tsx
mdjastrzebski: remove `console.warn`?

src/__tests__/__snapshots__/render.breaking.test.tsx.snap
mdjastrzebski: LOL, this is duplicate of existing `render.test.tsx` :-)

src/render.tsx
thymikee: Yeah

src/__tests__/__snapshots__/render.breaking.test.tsx.snap
thymikee: Then we can fix it in to places :D

thymikee (approved):

src/render.tsx
```diff
@@ -100,10 +101,20 @@ function buildRenderResult(
...
     rerender: update, // alias for `update`
     toJSON: renderer.toJSON,
     debug: debug(instance, renderer),
+    get root() {
+      return getHostChildren(instance)[0];
+    },
+    UNSAFE_root: instance,
+    get container() {
+      if (!getConfig().useBreakingChanges) {
+        return instance;
```
MattAgn, line 113: should we log a warning to tell the user `container` will be deprecated in the futur in favor of root?

MattAgn, line 113: should we log a warning to tell the user `container` will be deprecated in the futur in favor of root?

MattAgn (approved):

src/render.tsx
mdjastrzebski: Good idea, added that!

website/docs/API.md
```diff
@@ -251,13 +252,31 @@ toJSON(): ReactTestRendererJSON | null
...
+### `UNSAFE_root`
+
+```ts
+UNSAFE_root: ReactTestInstance;
+```
+
+Returns the rendered [composite root element](testing-env#host-and-composite-components).
+
+:::caution
+This API typically will return a composite view which goes against recommended testing practices. This API is primarily available for legacy test suits that rely on such testing.
```
pierrezimmermannbam, line 274:
```suggestion
This API typically will return a composite view which goes against recommended testing practices. This API is primarily available for legacy test suites that rely on such testing.
```

website/docs/API.md
```diff
@@ -251,13 +252,31 @@ toJSON(): ReactTestRendererJSON | null
...
 
 ```ts
-container: ReactTestInstance;
+root: ReactTestInstance;
 ```
 
-A reference to the rendered root element.
+Returns the rendered root [host element](testing-env#host-and-composite-components).
+
+This API is primarily useful in component tests, as it allows you to access root host view without using `*ByTestId` queries or similar methods.
```
pierrezimmermannbam, line 263: I feel like it could be interesting to give more indications on what we think are good use cases for this api (e.g. asserting an element is not rendered) because I don't think it's very intuitive how to use this

pierrezimmermannbam (approved):

Issue #1302, opened 2023-02-02
[1.15.1] new act implementation is causing actual tests to fail


idrissakhi:
## Describe the bug

When updating to new version, all tests having act will fail

## Expected behavior

update documentation with new way of using act or declare it as breaking change

## Steps to Reproduce

import { PropsWithChildren, ReactElement } from 'react';

import { NavigationContext } from '@react-navigation/native';

import { act } from 'react-test-renderer';

import { fireEvent, navContext, render } from 'test-utils';

import { CloseButton } from 'src/Features/Authentication/EmailSent/SubComponents/CloseButton/CloseButton';
import { routeNames } from 'src/Shared/Constants';

const TestComponent = ({ children }: PropsWithChildren<unknown>): ReactElement => (
  <NavigationContext.Provider value={navContext}>{children}</NavigationContext.Provider>
);

describe('CloseButton component', () => {
  it('on press navigates to login screen', () => {
    const { getByTestId } = render(<CloseButton />, { wrapper: TestComponent });
    const button = getByTestId('CloseButton');

    act(() => {
      fireEvent.press(button);
    });

    expect(navContext.navigate).toHaveBeenCalledTimes(1);
    expect(navContext.navigate).toHaveBeenLastCalledWith(routeNames.Login);
  });
});

report 

  ● CloseButton component › on press navigates to login screen

    Trying to detect host component names triggered the following error:

    Can't access .root on unmounted test renderer

    There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
    Please check if you are using compatible versions of React Native and React Native Testing Library.

      20 |
      21 |     act(() => {
    > 22 |       fireEvent.press(button);
         |                 ^
      23 |     });
      24 |
      25 |     expect(navContext.navigate).toHaveBeenCalledTimes(1);

## Versions

    "@testing-library/react-native": "11.5.1",

MuhmdRaouf:
I have the same issue (+1)

mdjastrzebski:
@MuhmdRaouf thanks for raporting this. We will look into that.

BTW could you try fireEvent without wrapping it into act? fireEvent is already automatically   wrapping it's operation with act.

@pierrezimmermannbam looking at the reported error it is related to detect host components PR.

pierrezimmermannbam:
@idrissakhi I am unable to reproduce the issue, what version of react-test-renderer are you using? From which version of react-native-testing-library did you upgrade?

[...]

PR #1304, opened 2023-02-03
fix: run autodetection on render to prevent detection to be wrapped by act

pierrezimmermannbam:
### Summary

Fixes #1302 by calling getHostComponentNames in render to prevent running auto detection in act which fails.

### Test plan

Added a test case where I call getHostComponentNames wrapped in act after a render and expect it does not throw

PR #1306, opened 2023-02-03
fix: get host component names on render (modified)

mdjastrzebski:
This is based on #1304, but since that was external PR I've moved it as internal repo PR with added proposed changes:
* extract `detectHostElementNames` from `getHostComponents()`
* create `configureHostElementNamesIfNeeded` to be called in `render`
* simplify `getHostElementNames`.

Summary
Fixes https://github.com/callstack/react-native-testing-library/issues/1302 by detecting host component names in `render` to prevent running auto detection in external `act` which fails for some users.

Test plan
Added a test case where I call getHostComponentNames wrapped in act after a render and expect it does not throw

### DIFF
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 import { View } from 'react-native';
 import TestRenderer from 'react-test-renderer';
 import { configureInternal, getConfig } from '../config';
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -1,8 +1,10 @@
+import React from 'react';
 import { View } from 'react-native';
 import TestRenderer from 'react-test-renderer';
 import { configureInternal, getConfig } from '../config';
 import { getHostComponentNames } from '../helpers/host-component-names';
 import * as within from '../within';
+import { act, render } from '..';
 
 const mockCreate = jest.spyOn(TestRenderer, 'create') as jest.Mock;
 const mockGetQueriesForElements = jest.spyOn(
@@ -33,6 +35,15 @@ describe('getHostComponentNames', () => {
       text: 'banana',
       textInput: 'banana',
     });
+  });
+
+  test('does not throw when wrapped in act after render has been called', () => {
+    render(<View />);
+    expect(() =>
+      act(() => {
+        getHostComponentNames();
+      })
+    ).not.toThrow();
   });
 
   test('throw an error when autodetection fails', () => {
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -27,13 +27,16 @@ describe('getHostComponentNames', () => {
     });
   });
 
-  test('throws when names are missing in the internal config', () => {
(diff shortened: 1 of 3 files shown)


src/helpers/host-component-names.tsx
```diff
@@ -1,21 +1,33 @@
...
 export function getHostComponentNames(): HostComponentNames {
+  const configHostComponentNames = getConfig().hostComponentNames;
+  if (!configHostComponentNames) {
+    throw new Error(`Missing host component names.\n\n${defaultErrorMessage}`);
+  }
+
+  return configHostComponentNames;
+}
+
+export function configureHostComponentNamesIfNeeded() {
```
MattAgn, line 18: love this 3 function split, much clearer!

src/__tests__/host-component-names.test.tsx
```diff
@@ -11,10 +16,41 @@ const mockGetQueriesForElements = jest.spyOn(
...
+  test('throws when names are missing in the internal config', () => {
+    expect(() => getHostComponentNames()).toThrowErrorMatchingInlineSnapshot(`
+      "Missing host component names.
+
+      There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
+      Please check if you are using compatible versions of React Native and React Native Testing Library."
+    `);
+  });
+
+  test('does not throw when wrapped in act after render has been called', () => {
```
MattAgn, line 39: maybe we could add some explanation here as to why this test is needed wdyt ? or we could also just link the issue (but will take longer to understand for futur devs)

MattAgn (approved):

src/__tests__/host-component-names.test.tsx
```diff
@@ -11,10 +16,41 @@ const mockGetQueriesForElements = jest.spyOn(
...
+    });
+
+    expect(getHostComponentNames()).toEqual({
+      text: 'banana',
+      textInput: 'banana',
+    });
+  });
+
+  test('throws when names are missing in the internal config', () => {
+    expect(() => getHostComponentNames()).toThrowErrorMatchingInlineSnapshot(`
```
pierrezimmermannbam, line 31: I think the error message for `getHostComponentNames` function should be updated, it will throw if `configureHostComponentNamesIfNeeded` has not been called before but that may not be linked to the configuration and would likely be a bug on our end, wdyt ?

src/__tests__/host-component-names.test.tsx
mdjastrzebski: Added comment and reference

mdjastrzebski:
@pierrezimmermannbam @MattAgn In order to simplify code, reduce error handling I've added back lazy detection to `getHostComponentNames()`. Pls take another look at this PR.

mdjastrzebski:
@pierrezimmermannbam @MattAgn In order to simplify code, reduce error handling I've added back lazy detection to `getHostComponentNames()`. Pls take another look at this PR.

src/__tests__/host-component-names.test.tsx
mdjastrzebski: Resolved by triggering lazy detection inside `getHostComponentNames`

MattAgn (approved):

pierrezimmermannbam (approved):

idrissakhi:
well done 👍

PR #1313, opened 2023-02-13
[v12] refactor: enable all breaking changes

mdjastrzebski:
Note: this PR merges to `v12` branch, for future v12 release.

### Summary

- [x] enable by default all features hidden behind `useBreakingChanges` (without changing the flag value)
- [x] remove dead code & dead tests
- [x] update docs

### Test plan

All tests pass


website/docs/MigrationV12.md
```diff
@@ -0,0 +1,28 @@
+---
+id: migration-v12
+title: Migration to 12.0
+---
+
+Migration to React Native Testing Library version 12 from version 11.x should be a relatively easy task due small amount of breaking changes.
```
thymikee, line 6:
```suggestion
React Native Testing Library 12 introduces a handful of breaking changes compared to 11.x. We believe they were necessary to improve the experience using the library and help the users fall into the pit of success when writing meaningful tests. You will find migration instructions for each and every change below.
```

website/docs/MigrationV12.md
```diff
@@ -0,0 +1,28 @@
...
+
+# Breaking changes
+
+1. `*ByRole` queries now return only accessibility elements, either explicitly marked with `accessible` prop or implicit ones where this status is derrived from component type itself (e.g `Text`, `TextInput`, `Switch`).
```
thymikee, line 10: Needs code example / comparison

website/docs/MigrationV12.md
```diff
@@ -0,0 +1,28 @@
...
+
+You may need to adjust relevant components under test to make sure they pass `isAccessibilityElement` check.
+
+2. `*ByText`, `*ByDisplayValue`, `*ByPlaceholderText` queries now return host components, as all other queries.
```
thymikee, line 14: add comparison pls

website/docs/MigrationV12.md
```diff
@@ -0,0 +1,28 @@
...
+
+While potentially breaking, this should not cause issues in tests.
```
thymikee, line 16: What if it does? Do we know how this can break and guide users on how to do it?

website/docs/MigrationV12.md
```diff
@@ -0,0 +1,28 @@
...
+
+3. `container` API has been renamed to `UNSAFE_root`.
+
+Try using `root` API as a self alternative. The difference between these two propertie is that `root` will always represent a host component, while `UNSAFE_root` will typically represent a composite component.
```
thymikee, line 20:
```suggestion
Historically `container` was supposed to mimic the [RTL's container](https://testing-library.com/docs/react-testing-library/api/#container). However it turned out not so relevant in RNTL's environment, where we proxied it to the underlying React Test Renderer's instance. In v12 RNTL introduces `root` API as an alternative that returns a host element instead of the full instance. The difference between these two propertie is that `root` will always represent a host component, while `UNSAFE_root` will typically represent a composite component of ReactTestInstance.
```

thymikee (commented):
👍🏼 I think we need to start writing blog posts where migration guide is only an extension :D

website/docs/MigrationV12.md
mdjastrzebski: Not sure what do you mean by adding comparison here. I guess that in most cases this will not affect users tests, so there is nothing to show here.

website/docs/MigrationV12.md
mdjastrzebski: Hontestly, we do not have a clear case when it breaks, as composite & host `Text` / `TextInputs` are very simillar in terms of props. For the types of test we recommend (query + jest matchers) it should work the same. The case where it might break is when user did e.g. explicit type checks, or inspects children, etc

website/docs/MigrationV12.md
mdjastrzebski: "help users fall into the pit of success when ..." 😂

website/docs/MigrationV12.md
thymikee: Maybe that `returnedValue.type` is now always a string instead of a function for example?

website/docs/MigrationV12.md
mdjastrzebski: @thymikee Is this the pit you are referring to: https://blog.codinghorror.com/falling-into-the-pit-of-success/

website/docs/MigrationV12.md
thymikee: > The case where it might break is when user did e.g. explicit type checks, or inspects children, etc

To me it's something that's worth mentioning

website/docs/MigrationV12.md
thymikee: c'mon

website/docs/MigrationV12.md
mdjastrzebski: Yes, this is true, however we never encouraged users to inspect `instance.type`, as this is largely an implementation detail for the users.

thymikee: We never did, but what's possible, is gonna be used eventually. https://www.hyrumslaw.com

website/docs/MigrationV12.md
mdjastrzebski: Added

src/queries/__tests__/role.test.tsx
```diff
@@ -8,6 +8,11 @@ import {
   Button as RNButton,
 } from 'react-native';
 import { render } from '../..';
+import { configureInternal } from '../../config';
+
+beforeEach(() => {
+  configureInternal({ useBreakingChanges: true });
```
pierrezimmermannbam: this can be removed I think

src/queries/__tests__/text.test.tsx
```diff
@@ -8,6 +8,11 @@ import {
   TextInput,
 } from 'react-native';
 import { render, getDefaultNormalizer, within } from '../..';
+import { configureInternal } from '../../config';
+
+beforeEach(() => {
+  configureInternal({ useBreakingChanges: true });
```
pierrezimmermannbam: this is no longer needed

src/queries/placeholderText.ts
```diff
@@ -24,10 +24,7 @@ const getTextInputNodeByPlaceholderText = (
...
 ) => {
   const { exact, normalizer } = options;
-  const shouldReturnHostTextInput = getConfig().useBreakingChanges;
-  const textInputType = shouldReturnHostTextInput
-    ? getHostComponentNames().textInput
-    : TextInput;
+  const textInputType = getHostComponentNames().textInput;
 
   return (
     filterNodeByType(node, textInputType) &&
```
pierrezimmermannbam, line 28: I think getHostComponentNames().textInput could be inlined

src/queries/displayValue.ts
```diff
@@ -26,10 +26,7 @@ const getTextInputNodeByDisplayValue = (
...
   const nodeValue =
     node.props.value !== undefined ? node.props.value : node.props.defaultValue;
-  const shouldReturnHostTextInput = getConfig().useBreakingChanges;
-  const textInputType = shouldReturnHostTextInput
-    ? getHostComponentNames().textInput
-    : TextInput;
+  const textInputType = getHostComponentNames().textInput;
 
   return (
     filterNodeByType(node, textInputType) &&
```
pierrezimmermannbam, line 30: we could use getHostComponentNames().textInput directly here

src/queries/__tests__/role.test.tsx
pierrezimmermannbam: ok this was removed in a ulterior commit

src/queries/__tests__/text.test.tsx
pierrezimmermannbam: this also has been removed

website/docs/MigrationV12.md
```diff
@@ -0,0 +1,60 @@
...
+// Explicit "accessible={false}" for Text,which is implicitly accessible element
+<Text accessible={false} role="button">Button</Text>
+```
+
+## 2. `*ByText`, `*ByDisplayValue`, `*ByPlaceholderText` queries now return host components
+`*ByText`, `*ByDisplayValue`, `*ByPlaceholderText` queries now return host components, as all other queries.
+
+While potentially breaking, this should not cause issues in tests if you are using recommended queries and Jest Matchers from Jest Native package. 
+
+Problematic cases may include: examining `type` property of `ReactTestInstance`, referencing other nodes using `parent` or `children` props, etc.
```
pierrezimmermannbam, line 46: I feel like the most common case breaking will be reading element prop, which is not listed here and it's not clear that this particular case should be solved through usage of jest native matchers imo

pierrezimmermannbam: Also it could be useful to mention that the parent of the element that is now returned is the composite element that was returned before v12 as it provides an easy way to fix broken tests at least temporarily

website/docs/MigrationV12.md
```diff
@@ -0,0 +1,60 @@
...
+
+// Explicit "accessible={false}" prop for View
+<View accessible={false} role="button" />
+
+// Explicit "accessible={false}" for Text,which is implicitly accessible element
+<Text accessible={false} role="button">Button</Text>
+```
+
+## 2. `*ByText`, `*ByDisplayValue`, `*ByPlaceholderText` queries now return host components
+`*ByText`, `*ByDisplayValue`, `*ByPlaceholderText` queries now return host components, as all other queries.
```
pierrezimmermannbam, line 42: I'd have linked the testing environment doc that explains what host components are

pierrezimmermannbam (changes requested):
Very nice! We're very close to the release now! The documentation on testing environment mentions that some queries return composite components which is no longer true so it should be updated as well. I've left some other comments, mainly regarding the migration guide

src/config.ts
```diff
@@ -26,15 +26,11 @@ export type HostComponentNames = {
 };
 
 export type InternalConfig = Config & {
-  /** Whether to use breaking changes intended for next major version release. */
-  useBreakingChanges: boolean;
```
pierrezimmermannbam, line 30: Don't we want to keep this flag for the next time we introduce breaking changes?

mdjastrzebski: I was thinking about pros and cons, and it seems like YAGNI approach makes more sense.

It was created in a particilar project situation in mind. Future cases might be simillar or might be better addressed e.g. by version branches.

We can re-introduce it fairly easily if needed.

website/docs/MigrationV12.md
mdjastrzebski: Updated comments. I would not hint users about `parent` fix, as I think users are better by writing these tests in a recommended way.

pierrezimmermannbam (approved):

Issue #1307, opened 2023-02-08
Should the custom jest preset be deprecated?


roryabraham:
## Ask your Question

I see that the [custom jest preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset) which does not override promise was added in response to [this upstream issue](https://github.com/facebook/react-native/issues/29303). Now that that issue has been resolved upstream in https://github.com/facebook/react-native/commit/f1fdc8b9b636e010ae5d5e737fc81a8da8f536d4 and released in [v0.71.2](https://github.com/facebook/react-native/releases/tag/v0.71.2), should we update documentation in this repo to deprecate the custom jest preset and only recommend it to developers using ["modern fake timers"](https://jestjs.io/blog/2020/05/05/jest-26#new-fake-timers) _and_ RN < `0.71.2` ?

mdjastrzebski:
@roryabraham that's a good question. We will have to review it. However, it seems fairly low prio, so it might take a will to handle it.

pierrezimmermannbam:
Oh I hadn't seen that promise was no longer polyfilled that's great news! I think it's worth mentioning in the readme that it's no longer needed for RN >= 0.71.2, I'll open a PR to update the readme

thymikee:
I wouldn't deprecate it, but treat as a proxy to original preset. When we need to add something custom, we can do it without much work from the users

mdjastrzebski:
Resolved in #1316

PR #1316, opened 2023-02-14
docs: update docs on custom jest preset

pierrezimmermannbam:
### Summary

Based on #1307

Updates docs to specify that jest preset is no longer required for users that have a version of react native >= 0.71.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 
 ### Custom Jest Preset
 
-> **important** if you use "modern" Fake Timers
+> **important** if you use "modern" Fake Timers and React Native < 0.71
 
 We generally advise to use the "react-native" preset when testing with this library. However, if you use ["modern" Fake Timers](https://jestjs.io/blog/2020/05/05/jest-26#new-fake-timers) (default since Jest 27), you'll need to apply our custom Jest preset or awaiting promises, like `waitFor`, will timeout.
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 }
 ```
 
-### Custom Jest Preset
+### Custom Jest Preset (React Native before 0.71)
 
 We generally advise to use the "react-native" preset when testing with this library.
 
--- a/README.md
+++ b/README.md
@@ -74,9 +74,9 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 
 ### Custom Jest Preset
 
-> **important** if you use "modern" Fake Timers and React Native < 0.71
+We generally advise to use the "react-native" preset when testing with this library.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -74,7 +74,7 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 
 ### Custom Jest Preset
 
-> **important** if you use "modern" Fake Timers
+> **important** if you use "modern" Fake Timers and React Native < 0.71.2
```
thymikee, line 77: This commit is available in all 0.71 variants
image

pierrezimmermannbam: Oh I hadn't seen this, I updated the docs accordingly, thanks!

website/docs/API.md
```diff
@@ -563,7 +563,7 @@ It is also recommended to have a [single assertion per each `waitFor`](https://k
 ### Using Jest fake timers
 
 :::caution
-When using modern fake timers (the default for `Jest` >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the custom [@testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset). 
+When using modern fake timers (the default for `Jest` >= 27), `waitFor` won't work (it will always timeout even if `expectation()` doesn't throw) unless you use the custom [@testing-library/react-native preset](https://github.com/callstack/react-native-testing-library#custom-jest-preset) or you use a version of React Native >= 0.71.
```
mdjastrzebski, line 566: Can we flip the order of conditions explained in the docs? I would rather explain it in a way that:
1. if you use React Native earlier that 0.71
2. and you use modern Jest fake timers which are default for Jest 27+
3. then you should use custom preset or `waitFor` + fake timers will not work

README.md
```diff
@@ -74,7 +74,7 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 
 ### Custom Jest Preset
 
-> **important** if you use "modern" Fake Timers
+> **important** if you use "modern" Fake Timers and React Native < 0.71
 
 We generally advise to use the "react-native" preset when testing with this library. However, if you use ["modern" Fake Timers](https://jestjs.io/blog/2020/05/05/jest-26#new-fake-timers) (default since Jest 27), you'll need to apply our custom Jest preset or awaiting promises, like `waitFor`, will timeout.
```
mdjastrzebski, line 79:
```suggestion
We generally advise to use the "react-native" preset when testing with this library.

However, if you use React Native version earlier than 0.71 with [modern Jest fake timers](https://jestjs.io/blog/2020/05/05/jest-26#new-fake-timers) (default since Jest 27), you'll need to apply this custom Jest preset or otherwise awaiting promises, like using `waitFor` or `findBy*`, queries will fail with timeout.
```

mdjastrzebski (changes requested):
I would suggest re-ordering of our preset info to mention RN version application first, this will allow users on RN 0.71+ to quickly establish that they do not need to use it.

README.md
```diff
@@ -74,7 +74,7 @@ Then automatically add it to your jest tests by using `setupFilesAfterEnv` optio
 
 ### Custom Jest Preset
```
mdjastrzebski, line 75:
```suggestion
### Custom Jest Preset (React Native before 0.71)
```

website/docs/API.md
pierrezimmermannbam: Agreed, I changed it and also the title

mdjastrzebski (approved):

PR #1317, opened 2023-02-14
[v12] refactor: exclude hidden elements by defaults

mdjastrzebski:
### Summary

- [x] Switch `defaultIncludeHiddenElements: false`
- [x] Update tests
- [x] Update docs

### Test plan

All checks should pass. Pls review testing code.

### DIFF
--- a/src/helpers/accessiblity.ts
+++ b/src/helpers/accessiblity.ts
@@ -59,6 +59,11 @@ export function isHiddenFromAccessibility(
 export const isInaccessible = isHiddenFromAccessibility;
 
 function isSubtreeInaccessible(element: ReactTestInstance): boolean {
+  // Null props can happen for React.Fragments
+  if (element.props == null) {
+    return false;
+  }
+
   // iOS: accessibilityElementsHidden
   // See: https://reactnative.dev/docs/accessibility#accessibilityelementshidden-ios
   if (element.props.accessibilityElementsHidden) {
--- a/website/docs/API.md
+++ b/website/docs/API.md
@@ -884,7 +884,7 @@ Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementTo
 
 #### `defaultIncludeHiddenElements` option
 
-Default value for [includeHiddenElements](Queries.md#includehiddenelements-option) query option for all queries. Defaults to `false`, which means that queries will not match [elements hidden from accessibility](#ishiddenfromaccessibility) by default.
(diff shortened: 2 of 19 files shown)


pierrezimmermannbam (approved):

src/helpers/accessiblity.ts
```diff
@@ -59,6 +59,11 @@ export function isHiddenFromAccessibility(
 export const isInaccessible = isHiddenFromAccessibility;
 
 function isSubtreeInaccessible(element: ReactTestInstance): boolean {
+  // Null props can happen for React.Fragments
+  if (element.props == null) {
```
thymikee, line 63: can it be undefined as well? if so, let's go with `===`

website/docs/API.md
```diff
@@ -884,9 +884,7 @@ Default timeout, in ms, for async helper functions (`waitFor`, `waitForElementTo
 
 #### `defaultIncludeHiddenElements` option
 
-Default value for [includeHiddenElements](Queries.md#includehiddenelements-option) query option for all queries. Defaults to `true`, which means that queries will match [elements hidden from accessibility](#ishiddenfromaccessibility) by default.
-
-Currently this option is set to `true` which means that queries will also match hidden elements. This is done to avoid breaking changes. However, we plan to change the default behavior to exclude hidden elements in the next major release.
+Default value for [includeHiddenElements](Queries.md#includehiddenelements-option) query option for all queries. Defaults to `false`, which means that queries will not match [elements hidden from accessibility](#ishiddenfromaccessibility) by default.
```
thymikee, line 887:
```suggestion
Default value for [includeHiddenElements](Queries.md#includehiddenelements-option) query option for all queries. Defaults to `false`, which means that queries will not match [elements hidden from accessibility](#ishiddenfromaccessibility) by default. Similar to the users that can't interact with those.
```

src/helpers/accessiblity.ts
mdjastrzebski: `== null` is a safer check, as it matches exactly `undefined` and `null`.

mdjastrzebski: BTW according to official type info this cannot be `undefined` but is for `React.Fragments`

thymikee (approved):

src/helpers/accessiblity.ts
thymikee: Yeah this is what I asked for, just a bit indirectly. Let's leave it as is then

Issue #1347, opened 2023-02-24
Test with suspending component times out with `fakeTimers: { enableGlobally: true }`


robingullo:
## Describe the bug

I've set up a simple test with a suspending component (no timers, just a return in async / await) that eventually renders a text. The test waits for the text with `await findByText`.
With  `jest.useFakeTimers()`, the test pass.
With `fakeTimers: { enableGlobally: true }` in `jest.config.js`, the test fails.

## Expected behavior

The test should pass in both cases.

## Steps to Reproduce

- Clone the examples/basic directory from the repo
- Add this test:
```
import * as React from 'react';
import { render, screen } from '@testing-library/react-native';
import { Text, View } from 'react-native';

beforeEach(() => {
  jest.useFakeTimers();

  // Logs a warning if modern fake timers are not set
  jest.setSystemTime(Date.now());
});

test('should work', async () => {
  let loading = true;
  let data: string | undefined = undefined;

  const Suspending = () => {
    if (loading)
      throw (async () => {
        return await 'result';
      })().then((result) => {
        data = result;
        loading = false;
      });
    return <Text>{data}</Text>;
  };

  render(
    <View>
      <React.Suspense fallback={<Text>Loading</Text>}>
        <Suspending />
      </React.Suspense>
    </View>
  );

  await screen.findByText('result');
});
```
- Run the test: it passes ✅
- In `jest.config.js`, add the line `fakeTimers: { enableGlobally: true }`
- Run the test: it fails ❌

## Versions

  npmPackages:
    @testing-library/react-native: ^11.4.0 => 11.5.2 
    react: 18.1.0 => 18.1.0 
    react-native: 0.70.5 => 0.70.5 
    react-test-renderer: 18.1.0 => 18.1.0

Note: 
I have the same result with RN 0.71.3 and the standard 'react-native' preset.

mdjastrzebski:
Not sure how well RNTL handles React Suspense, I will have to spend some time to check it out.

BTW why are you throwing a function returning a promise? Also `await 'result'` seems to be unnecessary as awaiting non-promise value just returns it. Shouldn't the `Suspending` component throw a promise?

robingullo:
I think it throws a promise, the arrow function is called immediately. I admit it might not be the clearest syntax.

robingullo:
I encountered the issue initally with React Query btw

pierrezimmermannbam:
[...]

PR #1366, opened 2023-03-15
Flush all pending microtasks and updates before returning from waitFor

jsnajdr:
It can happen that a `waitFor` call like
```js
await waitFor(() => getByText('green'));
```
successfully returns because the `"green"` text has been rendered, but after returning, the effects or state updates scheduled by the `"green"` render haven't been executed yet.

That's because `waitFor` internally leaves the `act` environment for a moment, lets a render to be performed and to schedule updates outside the `act` queue, and then returns before these updates were flushed.

I'm submitting a test that exposes this bug, using a series of cascaded effects and state updates, and also a fix.

Accidentally, @eps1lon landed almost exactly the same fix to `@testing-library/react` a month ago: https://github.com/testing-library/react-testing-library/pull/1137. But I discovered the React Native bug independently, while investigating why [our own tests are failing](https://github.com/WordPress/gutenberg/pull/46735#issuecomment-1469931650). 🙂

### DIFF
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -281,9 +281,7 @@ test.each([
 
       // On mount, set the color to "red" in a promise microtask
       React.useEffect(() => {
-        // we want to use a Promise, as async functions are not very ergonomic as effects,
-        // and we're sure we don't need a catch handler.
-        // eslint-disable-next-line , promise/prefer-await-to-then, promise/catch-or-return
+        // eslint-disable-next-line promise/prefer-await-to-then, promise/catch-or-return
         Promise.resolve('red').then((c) => setColor(c));
       }, []);
 
@@ -302,7 +300,7 @@ test.each([
       );
     }
 
-    const onPress = jest.fn<void, [string]>();
+    const onPress = jest.fn();
     const view = render(<Apple onPress={onPress} />);
 
     // Required: this `waitFor` will succeed on first check, because the "root" view is there
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -277,6 +277,9 @@ test.each([false, true])(
 
       // On mount, set the color to "red" in a promise microtask
       React.useEffect(() => {
+        // we want to use a Promise, as async functions are not very ergonomic as effects,
+        // and we're sure we don't need a catch handler.
+        // eslint-disable-next-line , promise/prefer-await-to-then, promise/catch-or-return
         Promise.resolve('red').then((c) => setColor(c));
       }, []);
 
--- a/src/__tests__/waitFor.test.tsx
+++ b/src/__tests__/waitFor.test.tsx
@@ -263,3 +263,52 @@ test.each([false, true])(
     expect(mockFn).toHaveBeenCalledTimes(3);
   }
 );
+
+test.each([false, true])(
+  'flushes scheduled updates before returning (fakeTimers = %s)',
+  async (fakeTimers) => {
+    if (fakeTimers) {
+      jest.useFakeTimers();
+    }
+
+    function Apple({ onPress }: { onPress: (color: string) => void }) {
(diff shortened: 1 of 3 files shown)


src/waitFor.ts
```diff
@@ -196,7 +196,10 @@ export default async function waitFor<T>(
     setReactActEnvironment(false);
 
     try {
-      return await waitForInternal(expectation, optionsWithStackTrace);
+      const result = await waitForInternal(expectation, optionsWithStackTrace);
+      // drain the microtask queue before restoring the `act` environment
+      await new Promise((resolve) => setImmediate(resolve));
+      return result;
```
AugustinLF, line 202: I can see in the reference PR that there is
```ts
if (jestFakeTimersAreEnabled()) {
  jest.advanceTimersByTime(0)
}
```

Do you know why this is not needed here?

Did your added test failed with both fake and real timers?

AugustinLF (commented):
Nice catch, I've had this problem in several of my tests, looking forward for this fix!

src/waitFor.ts
AugustinLF: Oh, that's a really good catch, this `setImmediate` problem explains the issue we've had in #1347, it makes sense to me.

src/__tests__/waitFor.test.tsx
```diff
@@ -263,3 +263,52 @@ test.each([false, true])(
...
+      jest.useFakeTimers();
+    }
+
+    function Apple({ onPress }: { onPress: (color: string) => void }) {
+      const [color, setColor] = React.useState('green');
+      const [syncedColor, setSyncedColor] = React.useState(color);
+
+      // On mount, set the color to "red" in a promise microtask
+      React.useEffect(() => {
+        Promise.resolve('red').then((c) => setColor(c));
```
mdjastrzebski, line 280: Pls add // eslint-ignore-next with a brief comment to fix lint

jsnajdr: Done ✅

mdjastrzebski: @jsnajdr Huge thanks for submitting both microtask-related PRs! This is complex subject, so I plan to reserve some time in the upocoming week to wrap my head around these.

mdjastrzebski (approved):
@jsnajdr Awesome work and research. Thanks for contributing this fix!

src/flushMicroTasks.ts
```diff
@@ -2,7 +2,7 @@ import { setImmediate } from './helpers/timers';
 
 type Thenable<T> = { then: (callback: () => T) => unknown };
 
-export function flushMicroTasks<T>(): Thenable<T> {
+export function flushMicroTasks(): Thenable<void> {
   return {
     // using "thenable" instead of a Promise, because otherwise it breaks when
     // using "modern" fake timers
```
jsnajdr, line 8: This is a weird thenable because the `then` function doesn't return anything and is therefore not chainable. Also, the `setImmediate` timer starts running at the time when the promise chain is _constructed_, not when it _runs_.
```js
flushMicroTasks().then( flushMicroTasks ).then( flushMicroTasks )
```
should run there `setImmediate`s after each other, but instead it will crash, and even after fixing the crash, it will run all three timers in parallel.

The way how `await` works doesn't expose these bugs, but outside `await`, the helper won't work.

My intuition is that the really correct version will look differently 🙂

mdjastrzebski: My bad, I've missed that comment.

So when used this in instead of inline `Promise` it passed the test correctly. @jsnajdr would you be able to submit a PR with better implementation. You seem to now the Promise stuff very well, so help in this area would be very welcomed.

mdjastrzebski: Should the implementation be just:

```
import { setImmediate } from './helpers/timers';

export function flushMicroTasks() {
  return new Promise((resolve) => setImmediate(resolve));
}
```

PR #1371, opened 2023-03-17
detectHostComponentNames: render test tree inside act to avoid timer leaks

jsnajdr:
Fixes a bug that I discovered while working on #1366: when first calling `render` in a test suite, the `render` function calls `detectHostComponentNames`, which mounts a testing tree with a few components. This tree is rendered outside the `act` environment, and therefore schedules a `setImmediate` callback to process effects. This scheduled callback then has an unwanted interaction with my actual testing code. If there is a `waitFor` inside my test, the effects scheduled by _my_ testing code will not be processed by _my_ `setImmediate`, but by the one scheduled by `detectHostComponentNames`. That makes the precise order of events inside my test non-deterministic and unpredictable.

Look at this call stack:

Screenshot 2023-03-17 at 11 55 24

and note how it starts in my `waitFor.test.ts` test, in the `render` function, then calls `detectHostComponentNames`, schedules `setImmediate` callback, and how that `setImmediate` later processes an update scheduled by my test code (the `Apple` component).

This PR avoids that by wrapping the render with `act`. That way, all effects scheduled by the render will come into the `actQueue`, and will be flushed before returning. No leaks.

In the code, I had to do some silly IIFE magic to make TypeScript happy: it doesn't know that `act` will synchronously and reliably initialize the `result` value.

### DIFF
--- a/src/helpers/host-component-names.tsx
+++ b/src/helpers/host-component-names.tsx
@@ -1,8 +1,7 @@
-import React from 'react';
+import * as React from 'react';
 import { Text, TextInput, View } from 'react-native';
-import TestRenderer from 'react-test-renderer';
-import type { ReactTestRenderer } from 'react-test-renderer';
 import { configureInternal, getConfig, HostComponentNames } from '../config';
+import { renderWithAct } from '../render-act';
 import { getQueriesForElement } from '../within';
 
 const userConfigErrorMessage = `There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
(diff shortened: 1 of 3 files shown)


src/helpers/host-component-names.tsx
```diff
@@ -29,12 +30,20 @@ export function configureHostComponentNamesIfNeeded() {
...
+      TestRenderer.act(() => {
+        result = TestRenderer.create(
+          <View>
+            <Text testID="text">Hello</Text>
+            <TextInput testID="textInput" />
+          </View>
+        );
+      });
+
+      // @ts-ignore act is syncronous, so renderer will already be initialised here
```
jsnajdr, line 44: There's a typo, missing 'h' in 'synchronous'

mdjastrzebski: Spotted it and fixed! :-)

mdjastrzebski:
Released in [v12.0.1](https://github.com/callstack/react-native-testing-library/releases/tag/v12.0.1) 🎉

PR #1373, opened 2023-03-20
Fixing disabled TextInput keyPress  event in fireEvent

An631:
### Summary

When trying to use `fireEvent(element, 'keyPress', event)` for a TextInput element that has been set to the disabled state the firing of the event ignored if the element was a TextInput that had been disabled. This check makes sense for onChangeText events, but it doesn't make sense for onKeyPress events because the developer might choose to do direct handling of the input state through key presses while disabling the default change text behavior of the input.

### Test plan

Added 2 new test cases for TextInput keyPress events: 
1. Text input is editable
2. TextInput is disabled

### DIFF
--- a/src/__tests__/fireEvent.test.tsx
+++ b/src/__tests__/fireEvent.test.tsx
@@ -274,6 +274,52 @@ test('should not fire on non-editable TextInput with nested Text', () => {
   expect(onChangeTextMock).not.toHaveBeenCalled();
 });
 
+test('Should fire keyPress on non-editable TextInput', () => {
+  const placeholder = 'Test placeholder';
+  const onKeyPressMock = jest.fn();
+  const KEY_PRESS_EVENT = {
+    nativeEvent: {
+      key:'1',
+    }
+  };
+
+  const { getByPlaceholderText } = render(
+    <View>
+      <TextInput
+        editable={false}
+        placeholder={placeholder}
+        onKeyPress={onKeyPressMock}
+      />
+    </View>
+  );
+
+  fireEvent(getByPlaceholderText(placeholder), 'keyPress', KEY_PRESS_EVENT);
+  expect(onKeyPressMock).toHaveBeenCalled();
+});
+
+test('Should fire keyPress on editable TextInput', () => {
+  const placeholder = 'Test placeholder';
+  const onKeyPressMock = jest.fn();
+  const KEY_PRESS_EVENT = {
+    nativeEvent: {
+      key:'1',
+    }
+  };
+
+  const { getByPlaceholderText } = render(
+    <View>
+      <TextInput
+        editable={true}
+        placeholder={placeholder}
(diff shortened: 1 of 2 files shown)


mdjastrzebski:
@An631 Could you clarify the intent of this PR. From the description it seems like `keyPress` event blocking has not worked for you, but the change actually narrowed down the disabled event handling to also include `keyChange` event.

The tests on the other hand cover `editable` = `true/false` for `changeText` event, but do not cover the `editable` = `true/false` for other events.

mdjastrzebski:
@An631 I've re-read the description and noticed that you mention two different events: `changeText` vs `keyPress`.

If we want to make changes in this event handling, the we should investigate how RN behaves in real app running on a device or simu/emulator, not inside RNTL, to being `editable={false}` for various events like `change`, `changeText,` `keyPress`, etc. Without having such analysis we risk breaking users` tests while exchanging one inaccurate behavior (ignoring all events) for another inaccurate behavior (ignoring only `changeText`, while other events also might be prevented).

An631:
Thanks @mdjastrzebski, my intent is to narrow down the check for disabled state to only those events that require it or for which it makes sense like onChangeText. I can see how narrowing the behavior of not firing the event only to onChangeText can have unintended consequences for existing users but the current behavior completely blocks the ability to test a component that is handling its own state through keypress events or any other event because all of them are blocked if the TextInput is set to disabled.

Would changing the behavior of the fireEvent code to **only allow the firing of onKeyPress events while the TextInput is disabled** make more sense to avoid a bigger change to the behavior that might affect others?

I found this issue while working on my own component where I want to handle all user interactions with the text and disable direct edits to its value; hence the use of onKeyPress and disabling of the input. It is impossible to test this component using RNTL because I can't fire any event on my component as it is set to 'disabled' by design.

mdjastrzebski:
I've recently worked on area mentioned in this PR. From what I understand, the idea of allowing `keyPress` events to be raised when `TextInput` `editable={false}` is not coherent with observed React Native behaviour.

When `editable` is `false`, the `TextInput` will not be able to acquire focus on neither iOS nor Android, and without being focused there is no way that user could trigger `keyPress` event on it. Hence, the current behaviour when`editable={false}` prevents most of the `TextInput` events seems correct.

@An631 Do you have any scenario, in real world React Native app, when user could be able to trigger `keyPress` on disabled `TextInput`?

An631:
Hi @mdjastrzebski, sorry for the late reply, it is possible to trigger the keypress scenario when running react native in Windows. 

We might want to consider desktop platforms as well which can run react native now given this: https://microsoft.github.io/react-native-windows/

I would understand if the focus of this package is to only support Android and IOs though.

Also, I haven't tested it but can't android run key press events on inputs when using external keyboards? Or is the focus completely denied even when "Tabing" around the UI?

PR #397, opened 2020-06-22
fix: modern jest timers

thymikee:
### Summary

Fixes #391 

### Test plan

Added regression tests.

Issue #391, opened 2020-06-22
Version 2.x not compatible with jest.useFakeTimers('modern');


ccfz:
## Describe the bug

Using `jest.useFakeTimers('modern');` in any spec file, causes the tests to fail due to timeout error:

> Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.Error: Timeout - Async callback was not invoked within the 5000 ms timeout specified by jest.setTimeout.

Downgrading to 1.14 fixes this, therefore I believe this is related to 2.x. 

I am using  `jest.useFakeTimers('modern');` in order to call `jest.setSystemTime` in my test just FYI.

## Expected behavior

Tests should pass

## Steps to Reproduce

Simple component:
```
import React from 'react';
import { Text } from 'react-native';

const TestComponent = () => (
  <Text>hello world!</Text>
);

export default TestComponent;

```

spec file:
```
import React from 'react';
import { render } from 'react-native-testing-library';

import TestComponent from 'src/components/TestComponent';

jest.useFakeTimers('modern');

describe('TestComponent', () => {
  it('should render hello world!', () => {
    const utils = render(<TestComponent />);

    expect(utils.getByText('hello world!')).not.toBeNull();
  });
});
```

## Versions

```
    react: 16.13.1 => 16.13.1 
    react-native: 0.62.2 => 0.62.2 
    react-native-testing-library: ^2.1.0 => 2.1.0 
    react-test-renderer: ^16.13.0 => 16.13.1 
```

thymikee:
Thanks, this is interesting. After a brief investigation, I've identified that this code is to "blame", specifically the `await flushMicroTasks();` call:
https://github.com/callstack/react-native-testing-library/blob/401d3d3daca737a317d906d9d9225dc976ec1db9/src/index.js#L12-L15

`flushMicroTasks` looks like this:
https://github.com/callstack/react-native-testing-library/blob/401d3d3daca737a317d906d9d9225dc976ec1db9/src/flushMicroTasks.js#L13-L15

Interesting thing is that it works fine with legacy timers.

The `@testing-library/react` resolves similar problem by requiring setImmediate from `timers` node module: 
https://github.com/testing-library/react-testing-library/pull/514/files#diff-3ee797e0d5d82fbd2501ceaa19ce170bR16-R18 
and returning a thenable: 
[...]

PR #1374, opened 2023-03-23
refactor: flush microtask queue implementation

mdjastrzebski:
### Summary

**Edit**
This PR renames the old `flushMicroTasks` functions as `flushMicroTasksLegacy`, and adds a proper one under the name of `flushMicroTasks`. Respective implementations are invoked in the sample places as before, so there should be no behavior change for the users. We will delay removal of `flushMicroTasksLegacy` until the next major release, to avoid introducing breaking changes mid-cycle. 

**Original**
PR to fix `flushMicrotaskQueue` implemantion based on @jsnajdr [comment](https://github.com/callstack/react-native-testing-library/pull/1366#discussion_r1145947350):

> This is a weird thenable because the then function doesn't return anything and is therefore not chainable. Also, the setImmediate timer starts running at the time when the promise chain is constructed, not when it runs.
>
> `flushMicroTasks().then( flushMicroTasks ).then( flushMicroTasks )`
> should run there `setImmediate`s after each other, but instead it will crash, and even after fixing the crash, it will run all three timers in parallel.
> 
> The way how `await` works doesn't expose these bugs, but outside `await`, the helper won't work.
> 
> My intuition is that the really correct version will look differently 🙂

This is the implementation in question:

```
export function flushMicroTasks(): Thenable<void> {
  return {
    // using "thenable" instead of a Promise, because otherwise it breaks when
    // using "modern" fake timers
    then(resolve) {
      setImmediate(resolve);
    },
  };
}
```

### Test plan

All current tests should pass.

### Reviewers
@jsnajdr pls review and share you comments.
@mikeduminy, @dcalhoun adding you, as you seem knowledgeable in this area. Could you pls share your comments and/or concerns.


src/flushMicroTasks.ts
```diff
@@ -1,13 +1,5 @@
...
-export function flushMicroTasks(): Thenable<void> {
-  return {
-    // using "thenable" instead of a Promise, because otherwise it breaks when
-    // using "modern" fake timers
-    then(resolve) {
-      setImmediate(resolve);
-    },
-  };
+export function flushMicroTasks() {
+  return new Promise((resolve) => setImmediate(resolve));
```
jsnajdr, line 4: Yes, this is the most straightforward and correct way to wait for `setImmediate`. But the original comment said that this implementation breaks with modern fake timers? That turns out to be not true, doesn't it?

By the way, there's one more `setImmediate` await that could be replaced with `flushMicroTasks`, in the `waitForInternal` function.

mdjastrzebski:
@thymikee could you take a look, the current implementation seems to come from you in PR #397 resolving issues #391. That does not seems to be needed any more as we have various tests with both modern and legacy fake timers, and all pass with more "standard" way of flushing microtask queue.

src/flushMicroTasks.ts
thymikee: Can you test it with older versions of Jest and see where they fixed the implementation?

jsnajdr:
I tried to debug why the `new Promise(r=>setImmediate(r))` implementation was failing with modern timers, and I tracked it down to this React Native PR: https://github.com/facebook/react-native/pull/34659

Before that PR, React Native's `jest/setup.js` file was installing a Promise polyfill from the `promise` NPM package. And that package internally calls `process.nextTick`, which is being hijacked by modern fake timers. Somebody had to advance the fake timers just to resolve promises and call their `then` handlers. And nobody was making that call.

With recent versions of React Native, and the current `react-native-testing-library` depends on a recent-enough version, that polyfill is no longer used, and all promises are native. Modern fake timers no longer break them.

thymikee (approved):
@mdjastrzebski we might need a note somewhere that this is only compatible with newer version of RN

jsnajdr:
That `Promise` polyfill removal was shipped in React Native 0.71, in January 2023, just a few weeks ago. And this package has a peer dependency on React Native >=0.59, keeping back compat back to November 2021. It needs to keep the weird thenable until the peer dep gets updated one day.

Or, alternatively, it could advance the timers, like:
```js
new Promise(resolve => {
  setImmediate(() => {
    resolve();
    if (jestFakeTimersEnabled()) {
      jest.advanceTimersByTime(0);
    }
  });
});
```
The `@testing-library/react` library does something similar, and we discussed the need for `advanceTimersByTime` in https://github.com/callstack/react-native-testing-library/pull/1366#discussion_r1138595374.

Here the situation is a bit different, the timers need to be advanced _after_ `resolve` has been called. The microtask queue needs to be already populated when we're flushing it.

dcalhoun (commented):
>  @dcalhoun adding you, as you seem knowledgeable in this area. Could you pls share your comments and/or concerns.

Thanks for the ping. I do not have any specific concerns at this time. 

@jsnajdr's [original concern](https://github.com/callstack/react-native-testing-library/pull/1366#discussion_r1145947350) makes sense. Also, he has done a fair amount of [in-depth research](https://github.com/WordPress/gutenberg/pull/45235#issuecomment-1300359931) into our [project's tests](https://github.com/WordPress/gutenberg/pull/46735). I believe he has a great understanding of the domain. 

Additionally, [the tests](https://github.com/callstack/react-native-testing-library/blob/7b02f510d6f7ce6c90c44fdc198e6b9fd9d0b0f4/src/__tests__/timers.test.ts) (added in https://github.com/callstack/react-native-testing-library/pull/568) covering one of my [original discoveries](https://github.com/callstack/react-native-testing-library/issues/506) related to `Promise` and fake timers appear to still be present and passing in this PR. 🎉

mdjastrzebski:
I've done some tests:
* RN: 0.71.4, R: 18.2.0, Jest: 29.4 - all tests pass
* RN: 0.71.4, R: 18.2.0, Jest: 28.0 - all tests pass
* RN: 0.71.4, R: 18.2.0, Jest: 27.0 - all tests pass (after setting `testEnvironment: "node"` in Jest config file)

* RN: 0.70.0, R: 18:1.0, Jest: 29.4 - all tests pass
* RN: 0.70.0, R: 18:1.0, Jest: 28.0 - all tests pass
* RN: 0.70.0, R: 18:1.0, Jest: 27.0 - all tests pass
 
* RN: 0.69.10, R: 18.0.0, Jest: 29.4 - all tests pass
* RN: 0.69.10, R: 18.0.0, Jest: 28.0 - all tests pass
* RN: 0.69.10, R: 18.0.0, Jest: 27.0 - all tests pass

jsnajdr (approved):
Looks like a good preparation for future deprecation :+1:

mdjastrzebski:

Issue #1375, opened 2023-03-27
Output is not helpful when ByText query fails


stevehanson:
Thank you for this excellent library!  I'm a longtime user of RNTL, and something I have struggled with is that when a `ByText` query fails, the output does not help me determine the cause of the failure:

```js
expect(screen.getByText(/HELLOOO/)).toBeDefined()

> Unable to find an element with text: /HELLOOO/
```

When I encounter a failure like the above, my usual next step is to insert a `screen.debug()` or some `console.log()` statements and rerun the test until I can understand the cause of the failure. 

If the test uses the async `findByText` query, then adding these `debug` or log statements manually can be a bit more difficult, and I usually do something like this:

```js
await sleep(100) // wait 100ms to simulate what the findBy normally does before debugging
screen.debug()
expect(away screen.findByText(/HELLOOO/)).toBeDefined() // the line that was erroring
```

## Previous workaround (`toHaveTextContent` matcher)

I recently found a workaround that fit my needs, using the [jest-native `toHaveTextContent` matcher](https://github.com/testing-library/jest-native#tohavetextcontent):

```
expect(screen.container).toHaveTextContent(/HELLOOO/)

> Expected element to have text content:
>       /HELLOOO/
>     Received:
>       HELLOWelcome to the AppSign InForgot Password
```

As can be seen above, the `toHaveTextContent` matcher outputs all of the text from the screen along with the error message, which can be quite helpful for diagnosing why the test is failing.

This workaround however no longer seems like a safe approach since in the latest version of RNTL `screen.container` is discouraged and has been [renamed to `UNSAFE_root`](https://callstack.github.io/react-native-testing-library/docs/api#unsafe_root).

Additionally, using the `toHaveTextContent` matcher is non-standard and something we have to train teams to use instead of the `ByText` queries. Ideally, we could solve this problem with the `ByText` queries themselves

## Comparing to React Testing Library (web)

As can be seen in [this Code Sandbox](https://codesandbox.io/s/react-testing-library-demo-9ik86?file=/src/__tests__/hello.js:248-307), React Testing Library helpfully outputs the full DOM when a `byText` query fails:

```
Unable to find an element with the text: /HELLOO/i. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

<body>
  <div>
    <h1>
      Hello 
      Jill
      !
    </h1>
  </div>
</body>
```

## Proposal

[...]

PR #1378, opened 2023-03-31
feat: Render element tree in query error messages

stevehanson:
## Summary

Closes https://github.com/callstack/react-native-testing-library/issues/1375.

This commit updates the error message to the `ByText` query so that it outputs the virtual DOM, which can help the developer pinpoint the error without having to rerun the test with a `debug()` statement.

**Update:** this PR now outputs the element tree for all query errors, not just `ByText`.

### Before

```ts
Unable to find an element with text: /HELLOOO/
```

### After

```ts
Unable to find an element with text: /HELLOOO/

<Text>
  My component
</Text>
```

### Screenshot

Here is an example of an error when an element does not match because it is hidden (in this case, `accessibilityElementsHidden: true`). The relevant prop is included. The virtual DOM also is formatted with syntax highlighting:

screenshot of test output with syntax highlighting of the virtual DOM

## Preserving visibility-related props

By default, the `byText` query only matches visible elements, so any props that hide an element are relevant to the test failure. For that reason, this PR preserves any props that hide an element. I followed the logic from [isSubtreeInaccessible](https://github.com/callstack/react-native-testing-library/blob/main/src/helpers/accessiblity.ts#L61) to determine how to do this:

* Preserve: `accessibilityElementsHidden=true`
  * does not preserve prop if false
* Preserve: `accessibilityViewIsModal=true`
   * does not preserve prop if false
* Preserve: `importantForAccessibility="no-hide-descendants"`
  * does not preserve prop if has other value
* Preserve: `style: { display: 'none' }`
  * flattens the styles first, does not preserve any other styles

## Test plan

I've added unit tests for the `mapVisibilityRelatedProps` function as well as snapshot tests for the output of `getByText` with all of the above conditions. The snapshots are a little ugly since the formatted virtual DOM includes special characters for the syntax highlighting.

* I did not test performance. Since this only runs when the test errors, it seems like any performance penalty would be worth the cost, but let me know if you'd like me to benchmark this.
* I also did not test this on a large virtual DOM tree to see what it would look like. Let me know if that is desired, and I can include the output

### DIFF
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -87,7 +87,7 @@ describe('configureHostComponentNamesIfNeeded', () => {
       .toThrowErrorMatchingInlineSnapshot(`
       "Trying to detect host component names triggered the following error:
 
-      Unable to find an element with testID: text
+      \`render\` method has not been called
 
       There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
       Please check if you are using compatible versions of React Native and React Native Testing Library."
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -79,15 +79,17 @@ describe('configureHostComponentNamesIfNeeded', () => {
   });
 
   test('throw an error when autodetection fails', () => {
+    const renderer = TestRenderer.create(<View />);
+
     mockCreate.mockReturnValue({
-      root: { type: View, children: [], props: {} },
+      root: renderer.root,
     });
 
     expect(() => configureHostComponentNamesIfNeeded())
       .toThrowErrorMatchingInlineSnapshot(`
       "Trying to detect host component names triggered the following error:
 
-      \`render\` method has not been called
+      Unable to find an element with testID: text
 
       There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
       Please check if you are using compatible versions of React Native and React Native Testing Library."
@@ -105,7 +107,7 @@ describe('configureHostComponentNamesIfNeeded', () => {
       .toThrowErrorMatchingInlineSnapshot(`
       "Trying to detect host component names triggered the following error:
 
-      getByTestId returned non-host component
+      Unable to find an element with testID: text
 
       There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
       Please check if you are using compatible versions of React Native and React Native Testing Library."
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -6,14 +6,13 @@ import {
   getHostComponentNames,
   configureHostComponentNamesIfNeeded,
 } from '../helpers/host-component-names';
-import * as within from '../within';
 import { act, render } from '..';
 
 const mockCreate = jest.spyOn(TestRenderer, 'create') as jest.Mock;
-const mockGetQueriesForElements = jest.spyOn(
-  within,
-  'getQueriesForElement'
-) as jest.Mock;
+
+beforeEach(() => {
+  mockCreate.mockReset();
+});
 
 describe('getHostComponentNames', () => {
   test('returns host component names from internal config', () => {
@@ -83,24 +82,6 @@ describe('configureHostComponentNamesIfNeeded', () => {
 
     mockCreate.mockReturnValue({
       root: renderer.root,
-    });
-
-    expect(() => configureHostComponentNamesIfNeeded())
-      .toThrowErrorMatchingInlineSnapshot(`
-      "Trying to detect host component names triggered the following error:
-
-      Unable to find an element with testID: text
-
-      There seems to be an issue with your configuration that prevents React Native Testing Library from working correctly.
-      Please check if you are using compatible versions of React Native and React Native Testing Library."
-    `);
-  });
-
-  test('throw an error when autodetection fails due to getByTestId returning non-host component', () => {
-    mockGetQueriesForElements.mockReturnValue({
-      getByTestId: () => {
-        return { type: View };
-      },
     });
 
     expect(() => configureHostComponentNamesIfNeeded())
--- a/src/helpers/mapProps.ts
+++ b/src/helpers/mapProps.ts
@@ -1,0 +1,26 @@
+import { StyleSheet } from 'react-native';
+import { MapPropsFunction } from './format';
+
+/**
+ * Preserve props that would cause an element to not be visible while stripping
+ * the rest
+ */
+export const mapVisibilityRelatedProps: MapPropsFunction = ({
+  accessibilityElementsHidden,
+  accessibilityViewIsModal,
+  importantForAccessibility,
+  style,
+}) => {
+  const styles = StyleSheet.flatten(style as any) ?? {};
+
+  return {
+    ...(accessibilityElementsHidden
+      ? { accessibilityElementsHidden }
+      : undefined),
+    ...(accessibilityViewIsModal ? { accessibilityViewIsModal } : undefined),
+    ...(importantForAccessibility === 'no-hide-descendants'
+      ? { importantForAccessibility }
+      : undefined),
+    ...(styles.display === 'none' ? { style: { display: 'none' } } : undefined),
+  };
+};
--- a/src/helpers/mapProps.ts
+++ b/src/helpers/mapProps.ts
@@ -2,46 +2,43 @@ import { StyleSheet } from 'react-native';
 import { MapPropsFunction } from './format';
 
 const propsToDisplay = [
+  'testID',
+  'nativeID',
   'accessibilityElementsHidden',
   'accessibilityViewIsModal',
   'importantForAccessibility',
-  'testID',
-  'nativeID',
+  'accessibilityRole',
   'accessibilityLabel',
   'accessibilityLabelledBy',
-  'accessibilityRole',
   'accessibilityHint',
   'placeholder',
   'value',
   'defaultValue',
+  'title',
 ];
 
-function isObject(
-  thing: unknown
-): thing is Record<string | symbol | number, unknown> {
-  return typeof thing === 'object' && !Array.isArray(thing) && thing !== null;
+function isObject(value: unknown): value is Record<string, unknown> {
+  return typeof value === 'object' && value !== null && !Array.isArray(value);
 }
 
 function removeUndefinedKeys(prop: unknown) {
-  if (isObject(prop)) {
-    const object = Object.keys(prop).reduce((acc, propName) => {
-      return {
-        ...acc,
-        ...(prop[propName] === undefined ? {} : { [propName]: prop[propName] }),
-      };
-    }, {});
-
-    const allValuesUndefined =
-      Object.values(object).findIndex((val) => val !== undefined) === -1;
-
-    if (allValuesUndefined) {
-      return undefined;
-    }
-
-    return object;
+  if (!isObject(prop)) {
+    return prop;
   }
 
-  return prop;
+  const result: Record<string, unknown> = {};
+  Object.keys(prop).forEach((propName) => {
+    if (prop[propName] !== undefined) {
+      result[propName] = prop[propName];
+    }
+  });
+
+  // If object does not have any props we will ignore it.
+  if (Object.keys(result).length === 0) {
+    return undefined;
+  }
+
+  return result;
 }
 
 /**
@@ -51,24 +48,20 @@ export const mapPropsForQueryError: MapPropsFunction = (props) => {
   const accessibilityState = removeUndefinedKeys(props.accessibilityState);
   const accessibilityValue = removeUndefinedKeys(props.accessibilityValue);
 
-  const styles = StyleSheet.flatten(props.style as any) ?? {};
+  const styles = StyleSheet.flatten(props.style) as any;
 
   // perform custom prop mappings
-  const mappedProps: Record<string, unknown> = {
-    ...(styles.display === 'none' ? { style: { display: 'none' } } : undefined),
-    ...(accessibilityState === undefined ? {} : { accessibilityState }),
-    ...(accessibilityValue === undefined ? {} : { accessibilityValue }),
+  const result: Record<string, unknown> = {
+    ...(styles?.display === 'none' ? { style: { display: 'none' } } : {}),
+    ...(accessibilityState !== undefined ? { accessibilityState } : {}),
+    ...(accessibilityValue !== undefined ? { accessibilityValue } : {}),
   };
 
-  // add props from propsToDisplay without mapping
-  return propsToDisplay.reduce((acc, propName) => {
+  propsToDisplay.forEach((propName) => {
     if (propName in props) {
-      return {
-        ...acc,
-        [propName]: props[propName],
-      };
+      result[propName] = props[propName];
     }
+  });
 
-    return acc;
-  }, mappedProps);
+  return result;
 };
--- a/src/helpers/mapProps.ts
+++ b/src/helpers/mapProps.ts
@@ -1,26 +1,71 @@
 import { StyleSheet } from 'react-native';
 import { MapPropsFunction } from './format';
 
-/**
- * Preserve props that would cause an element to not be visible while stripping
- * the rest
- */
-export const mapVisibilityRelatedProps: MapPropsFunction = ({
-  accessibilityElementsHidden,
-  accessibilityViewIsModal,
-  importantForAccessibility,
-  style,
-}) => {
-  const styles = StyleSheet.flatten(style as any) ?? {};
+const propsToDisplay = [
+  'accessibilityElementsHidden',
+  'accessibilityViewIsModal',
+  'importantForAccessibility',
+  'testID',
+  'nativeID',
+  'accessibilityLabel',
+  'accessibilityLabelledBy',
+  'accessibilityRole',
+  'accessibilityHint',
+  'placeholder',
+  'value',
+  'defaultValue',
+];
 
-  return {
-    ...(accessibilityElementsHidden
-      ? { accessibilityElementsHidden }
-      : undefined),
-    ...(accessibilityViewIsModal ? { accessibilityViewIsModal } : undefined),
-    ...(importantForAccessibility === 'no-hide-descendants'
-      ? { importantForAccessibility }
-      : undefined),
+function isObject(
+  thing: unknown
+): thing is Record<string | symbol | number, unknown> {
+  return typeof thing === 'object' && !Array.isArray(thing) && thing !== null;
+}
+
+function removeEmptyKeys(prop: unknown) {
+  if (isObject(prop)) {
+    const object = Object.keys(prop).reduce((acc, propName) => {
+      return {
+        ...acc,
+        ...(prop[propName] === undefined ? {} : { [propName]: prop[propName] }),
+      };
+    }, {});
+
+    if (!Object.values(object).find((val) => val !== undefined)) {
+      return undefined;
+    }
+
+    return object;
+  }
+
+  return prop;
+}
+
+/**
+ * Preserve props that are helpful in diagnosing test failures, while stripping rest
+ */
+export const mapPropsForQueryError: MapPropsFunction = (props) => {
+  const accessibilityState = removeEmptyKeys(props.accessibilityState);
+  const accessibilityValue = removeEmptyKeys(props.accessibilityValue);
+
+  const styles = StyleSheet.flatten(props.style as any) ?? {};
+
+  // perform custom prop mappings
+  const mappedProps: Record<string, unknown> = {
     ...(styles.display === 'none' ? { style: { display: 'none' } } : undefined),
+    ...(accessibilityState === undefined ? {} : { accessibilityState }),
+    ...(accessibilityValue === undefined ? {} : { accessibilityValue }),
   };
+
+  // add props from propsToDisplay without mapping
+  return propsToDisplay.reduce((acc, propName) => {
+    if (propName in props) {
+      return {
+        ...acc,
+        [propName]: props[propName],
+      };
+    }
+
+    return acc;
+  }, mappedProps);
 };
--- a/src/helpers/mapProps.ts
+++ b/src/helpers/mapProps.ts
@@ -22,7 +22,7 @@ function isObject(
   return typeof thing === 'object' && !Array.isArray(thing) && thing !== null;
 }
 
-function removeEmptyKeys(prop: unknown) {
+function removeUndefinedKeys(prop: unknown) {
   if (isObject(prop)) {
     const object = Object.keys(prop).reduce((acc, propName) => {
       return {
@@ -31,7 +31,10 @@ function removeEmptyKeys(prop: unknown) {
       };
     }, {});
 
-    if (!Object.values(object).find((val) => val !== undefined)) {
+    const allValuesUndefined =
+      Object.values(object).findIndex((val) => val !== undefined) === -1;
+
+    if (allValuesUndefined) {
       return undefined;
     }
 
@@ -45,8 +48,8 @@ function removeEmptyKeys(prop: unknown) {
  * Preserve props that are helpful in diagnosing test failures, while stripping rest
  */
 export const mapPropsForQueryError: MapPropsFunction = (props) => {
-  const accessibilityState = removeEmptyKeys(props.accessibilityState);
-  const accessibilityValue = removeEmptyKeys(props.accessibilityValue);
+  const accessibilityState = removeUndefinedKeys(props.accessibilityState);
+  const accessibilityValue = removeUndefinedKeys(props.accessibilityValue);
 
   const styles = StyleSheet.flatten(props.style as any) ?? {};
 
--- a/src/queries/__tests__/text.test.tsx
+++ b/src/queries/__tests__/text.test.tsx
@@ -7,7 +7,7 @@ import {
   Button,
   TextInput,
 } from 'react-native';
-import { render, getDefaultNormalizer, within } from '../..';
+import { render, getDefaultNormalizer, within, screen } from '../..';
 
 test('byText matches simple text', () => {
   const { getByText } = render(<Text testID="text">Hello World</Text>);
@@ -593,6 +593,52 @@ describe('error messages', () => {
         [33maccessibilityViewIsModal[39m=[32m{true}[39m
       [36m/>[39m"
     `);
+  });
+
+  test('only renders the DOM on last failure with findBy', async () => {
+    const { findByText } = render(
+      <View accessibilityViewIsModal key="this is filtered" />
+    );
+
+    jest.spyOn(screen, 'toJSON');
+
+    await expect(() => findByText(/foo/)).rejects.toThrow();
+
+    expect(screen.toJSON).toHaveBeenCalledTimes(1);
+  });
+
+  test('can still modify findBy error in custom onTimeout', async () => {
+    const { findByText } = render(
+      <View accessibilityViewIsModal key="this is filtered" />
+    );
+
+    jest.spyOn(screen, 'toJSON');
+    const onTimeout = jest.fn();
+
+    await expect(() =>
+      findByText(/foo/, undefined, {
+        onTimeout,
+      })
+    ).rejects.toThrow();
+
+    expect(screen.toJSON).toHaveBeenCalledTimes(1);
+    expect(onTimeout).toHaveBeenCalledTimes(1);
+  });
+
+  test('only renders the DOM on last failure with findAllBy', async () => {
+    const { findAllByText } = render(
+      <View accessibilityViewIsModal key="this is filtered" />
+    );
+
+    jest.spyOn(screen, 'toJSON');
+    const onTimeout = jest.fn();
+
+    await expect(() =>
+      findAllByText(/foo/, undefined, { onTimeout })
+    ).rejects.toThrow();
+
+    expect(screen.toJSON).toHaveBeenCalledTimes(1);
+    expect(onTimeout).toHaveBeenCalledTimes(1);
   });
 
   test('does not strip display: none from "style" prop, but does strip other styles', () => {
--- a/src/queries/__tests__/text.test.tsx
+++ b/src/queries/__tests__/text.test.tsx
@@ -501,18 +501,111 @@ test('byText support hidden option', () => {
   );
 });
 
-test('byText renders the React DOM without props on failure', async () => {
-  const { getByText } = render(
-    <Text accessibilityLabel="label">Some text</Text>
-  );
+describe('byText error message', () => {
+  test('renders the React DOM with regular props stripped', async () => {
+    const { getByText } = render(
+      <Text accessibilityLabel="label">Some text</Text>
+    );
 
-  expect(() => getByText(/foo/)).toThrowErrorMatchingInlineSnapshot(`
-    "Unable to find an element with text: /foo/
+    expect(() => getByText(/foo/)).toThrowErrorMatchingInlineSnapshot(`
+      "Unable to find an element with text: /foo/
 
-    [36m<Text>[39m
-      [0mSome text[0m
-    [36m</Text>[39m"
-`);
+      [36m<Text>[39m
+        [0mSome text[0m
+      [36m</Text>[39m"
+    `);
+  });
+
+  test('does not strip accessibilityElementsHidden prop if true', () => {
+    const { getByText } = render(
+      <View accessibilityElementsHidden={false}>
+        <Text accessibilityElementsHidden>Some text</Text>
+      </View>
+    );
+
+    expect(() => getByText(/foo/)).toThrowErrorMatchingInlineSnapshot(`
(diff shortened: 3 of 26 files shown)


thymikee:
Love it! I think it would be helpful to also show "testID" prop

stevehanson:
Thanks for taking a look, @thymikee! I'm happy to include `testID` if we think that is useful. I'm curious what criteria we should use to determine which props to include. I was planning to open PRs for the other queries, and my thinking was to only include props that are directly relevant to the failures, since that line seemed like one that users would intuitively understand. So the `byTestID` query would include `testId`, `byLabelText` would include `accessibilityLabel` and `accessibilityLabelledBy`, etc. Are there some props, like `testId`, that you think should always be included when we render the virtual DOM?

mdjastrzebski:
@stevehanson I've support the general idea around this PR. It seems helpful to report element tree to the user on query failure. That being said, the unfiltered element tree can be quite long, especially when all props are included, and some are especially noisy (e.g. style, default touch event handlers, etc). The approach you have taken so far, which allowing only accessibility related ones seems generally good direction. However, I think we should include a little bit more. 

* all a11y props you added so far
* `testID`
* `nativeID`
* `accessibilityLabel`
* `accessbilityLabeledBy`
* `accessibilityRole`
* `accessbilityHint`
* `accessibilityState`
* `accessibilityValue`
* `placeholder`
* `value`
* `defaultValue`

The reason for including these, is that a large tree without additional context might be hard to navigate, it will just be a deeply nested tree of `View`s with occasionals `Text`, `TextInputs`, `RNSafeArea`, etc. Adding the above props, if present, should at a lot of semantic context that would allow user to navigate the element tree better. That would be regardless of the query used, i.e. we should display `testIDs` also for `*ByText` queries, etc.

mdjastrzebski:
Other things to watch out for:
- [ ] `getByText` is invoked multiple times inside `findByText`.  Triggering tree formatting on every failure seem unnecessary and costly, as it's normal that some of the initial `getByText` invokations will be failed. We should explore possibility to generate element on `onTimeout` event for `findBy*` as well as keeping in working for `getBy*` on first failure.
- [ ] Write tests for all query types (`getBy`, `getAllBy`, `queryBy`, ...) to make sure all work as intended.

src/queries/text.ts
```diff
@@ -37,7 +40,9 @@ const getMultipleError = (text: TextMatch) =>
   `Found multiple elements with text: ${String(text)}`;
 
 const getMissingError = (text: TextMatch) =>
-  `Unable to find an element with text: ${String(text)}`;
+  `Unable to find an element with text: ${String(text)}
+
+${format(screen.toJSON() || [], { mapProps: mapVisibilityRelatedProps })}`;
```
mdjastrzebski, line 45: Pls add formatting rather in `makeQueries` function, in order to have it affect all query types.

src/helpers/mapProps.ts
```diff
@@ -0,0 +1,26 @@
...
+export const mapVisibilityRelatedProps: MapPropsFunction = ({
+  accessibilityElementsHidden,
+  accessibilityViewIsModal,
+  importantForAccessibility,
+  style,
+}) => {
+  const styles = StyleSheet.flatten(style as any) ?? {};
+
+  return {
+    ...(accessibilityElementsHidden
```
mdjastrzebski, line 17: Pls log values for other props that might help user navigate the element tree:
* testID
* nativeID
* accessibilityLabel
* accessbilityLabeledBy
* accessibilityRole
* accessbilityHint
* accessibilityState
* accessibilityValue
* placeholder
* value
* defaultValue

src/helpers/mapProps.ts
```diff
@@ -0,0 +1,26 @@
+import { StyleSheet } from 'react-native';
+import { MapPropsFunction } from './format';
+
+/**
+ * Preserve props that would cause an element to not be visible while stripping
+ * the rest
+ */
+export const mapVisibilityRelatedProps: MapPropsFunction = ({
```
mdjastrzebski, line 8: `mapPropsForQueryError` or similar might be more accurate?

stevehanson:
@mdjastrzebski thank you for the review! 

> However, I think we should include a little bit more.

I totally agree with this suggestion, and mostly agree with the props you suggested. I wired up the `mapProps` from this PR in a real project yesterday and found what you mentioned -- since most of the elements end up being `<View />`, stripping too many props can make it hard to make sense of the DOM.

One open question in my mind is if every query will render the same props? `accessibilityState` and `accessibilityValue`, for example, add 82 extra lines to the output in my example below but would never affect a `byText` query passing or failing, unless I'm mistaken. I don't feel like they will aid much in understanding the output, either, since there should be a corresponding `accessibilityRole`, label, or text, that makes those elements understandable, without needing to see the specific state or value.

It does seem like we would need those props for other queries, like maybe `byRole` or `byLabel`, in addition to maybe some other props, like `focusable`, `collapsable`, `accessible`. I'm just trying to think through how the output can be kept as small as possible while including everything relevant or helpful. I'm definitely open to including them for now if you feel strongly about this but wanted to bring this up for discussion. `accessibilityState` and `accessibilityValue` are probably the only ones I personally feel on the fence about for `byText`.

Here are some sample DOM trees from a small real app that uses React Navigation with multiple stacks, including bottom tabs. Just including this as a relevant data point to this discussion:

<details>
<summary>Debug output: all props except style (571 lines)</summary>

```jsx
<RNCSafeAreaView
  edges={
    Array [
      "bottom",
      "left",
      "right",
      "top",
    ]
  }
>
  <RNCSafeAreaProvider
    onInsetsChange={[Function anonymous]}
  >
    <RNSScreenStack>
      <RNSScreen
        collapsable={false}
        forwardedRef={[Function handleRef]}
        gestureResponseDistance={
          Object {
            "bottom": -1,
            "end": -1,
            "start": -1,
            "top": -1,
          }
        }
        nativeBackButtonDismissalEnabled={false}
        onAppear={[Function onAppear]}
        onDisappear={[Function onDisappear]}
        onDismissed={[Function onDismissed]}
        onHeaderBackButtonClicked={[Function onHeaderBackButtonClicked]}
        onNativeDismissCancelled={[Function onNativeDismissCancelled]}
        onTransitionProgress={[Function anonymous]}
        onWillDisappear={[Function onWillDisappear]}
        replaceAnimation="push"
        stackPresentation="push"
        swipeDirection="horizontal"
      >
        <RNSScreenStackHeaderConfig
          backButtonInCustomView={false}
          backgroundColor="rgb(255, 255, 255)"
          color="#111"
          direction="ltr"
          disableBackButtonMenu={false}
          hidden={true}
          hideBackButton={false}
          hideShadow={true}
          largeTitleHideShadow={false}
          title="Tabs"
          titleColor="#282828"
          topInsetEnabled={false}
          translucent={false}
        />
        <View
          accessibilityElementsHidden={false}
          importantForAccessibility="auto"
        >
          <View>
            <View>
              <RNSScreenNavigationContainer>
                <RNSScreen
                  activityState={2}
                  collapsable={false}
                  forwardedRef={[Function handleRef]}
                  gestureResponseDistance={
                    Object {
                      "bottom": -1,
                      "end": -1,
                      "start": -1,
                      "top": -1,
                    }
                  }
                >
                  <View
                    accessibilityElementsHidden={false}
                    importantForAccessibility="auto"
                  >
                    <View>
                      <View>
                        <RNSScreenStack>
                          <RNSScreen
                            collapsable={false}
                            forwardedRef={[Function handleRef]}
                            gestureResponseDistance={
                              Object {
                                "bottom": -1,
                                "end": -1,
                                "start": -1,
                                "top": -1,
                              }
                            }
                            nativeBackButtonDismissalEnabled={false}
                            onAppear={[Function onAppear]}
                            onDisappear={[Function onDisappear]}
                            onDismissed={[Function onDismissed]}
                            onHeaderBackButtonClicked={[Function onHeaderBackButtonClicked]}
                            onNativeDismissCancelled={[Function onNativeDismissCancelled]}
                            onTransitionProgress={[Function anonymous]}
                            onWillDisappear={[Function onWillDisappear]}
                            replaceAnimation="push"
                            stackPresentation="push"
                            swipeDirection="horizontal"
                          >
                            <RNSScreenStackHeaderConfig
                              backButtonInCustomView={false}
                              backgroundColor="rgb(255, 255, 255)"
                              color="#111"
                              direction="ltr"
                              disableBackButtonMenu={false}
                              hidden={true}
                              hideBackButton={false}
                              largeTitleHideShadow={false}
                              title="Home"
                              titleColor="#282828"
                              topInsetEnabled={false}
                              translucent={false}
                            />
                            <View
                              accessibilityElementsHidden={true}
                              importantForAccessibility="no-hide-descendants"
                            >
                              <View>
                                <RCTScrollView
                                  bounces={true}
                                  contentContainerStyle={
                                    Array [
                                      Object {
                                        "flexGrow": 1,
                                        "paddingVertical": 20,
                                      },
                                      Object {
                                        "paddingHorizontal": 20,
                                      },
                                      undefined,
                                    ]
                                  }
                                  contentInsetAdjustmentBehavior="automatic"
                                  keyboardShouldPersistTaps="handled"
                                  showsVerticalScrollIndicator={false}
                                  testID="HomeScreen"
                                >
                                  <View>
                                    <Text
                                      accessibilityRole="header"
                                    >
                                      Good Morning, 
                                      <Text>
                                        Dale
                                      </Text>
                                    </Text>
                                    <Text>
                                      Welcome to the Example app. Content will be coming soon!
                                    </Text>
                                    <View
                                      accessibilityRole="link"
                                      accessibilityState={
                                        Object {
                                          "busy": undefined,
                                          "checked": undefined,
                                          "disabled": undefined,
                                          "expanded": undefined,
                                          "selected": undefined,
                                        }
                                      }
                                      accessibilityValue={
                                        Object {
                                          "max": undefined,
                                          "min": undefined,
                                          "now": undefined,
                                          "text": undefined,
                                        }
                                      }
                                      accessible={true}
                                      collapsable={false}
                                      focusable={true}
                                      onBlur={[Function onBlur]}
                                      onClick={[Function onClick]}
                                      onFocus={[Function onFocus]}
                                      onResponderGrant={[Function onResponderGrant]}
                                      onResponderMove={[Function onResponderMove]}
                                      onResponderRelease={[Function onResponderRelease]}
                                      onResponderTerminate={[Function onResponderTerminate]}
                                      onResponderTerminationRequest={[Function onResponderTerminationRequest]}
                                      onStartShouldSetResponder={[Function onStartShouldSetResponder]}
                                    >
                                      <Text>
                                        Go to article!
                                      </Text>
                                    </View>
                                    <View>
                                      <Text>
                                        Some text
                                      </Text>
                                    </View>
                                  </View>
                                </RCTScrollView>
                              </View>
                            </View>
                          </RNSScreen>
                          <RNSScreen
                            collapsable={false}
                            forwardedRef={[Function handleRef]}
                            gestureResponseDistance={
                              Object {
                                "bottom": -1,
                                "end": -1,
                                "start": -1,
                                "top": -1,
                              }
                            }
                            nativeBackButtonDismissalEnabled={false}
                            onAppear={[Function onAppear]}
                            onDisappear={[Function onDisappear]}
                            onDismissed={[Function onDismissed]}
                            onHeaderBackButtonClicked={[Function onHeaderBackButtonClicked]}
                            onNativeDismissCancelled={[Function onNativeDismissCancelled]}
                            onTransitionProgress={[Function anonymous]}
                            onWillDisappear={[Function onWillDisappear]}
                            replaceAnimation="push"
                            stackPresentation="push"
                            swipeDirection="horizontal"
                          >
                            <RNSScreenStackHeaderConfig
                              backButtonInCustomView={false}
                              backgroundColor="rgb(255, 255, 255)"
                              color="#111"
                              direction="ltr"
                              disableBackButtonMenu={false}
                              hidden={false}
                              hideBackButton={false}
                              largeTitleHideShadow={false}
                              title=""
                              titleColor="#282828"
                              topInsetEnabled={false}
                              translucent={false}
                            />
                            <View
                              accessibilityElementsHidden={false}
                              importantForAccessibility="auto"
                            >
                              <View>
                                <RCTScrollView
                                  bounces={true}
                                  contentContainerStyle={
                                    Array [
                                      Object {
                                        "flexGrow": 1,
                                        "paddingVertical": 20,
                                      },
                                      Object {
                                        "paddingHorizontal": 20,
                                      },
                                      undefined,
                                    ]
                                  }
                                  contentInsetAdjustmentBehavior="automatic"
                                  keyboardShouldPersistTaps="handled"
                                  showsVerticalScrollIndicator={false}
                                  testID="ArticleScreen"
                                >
                                  <View>
                                    <Text
                                      accessibilityRole="header"
                                    >
                                      Article Heading
                                    </Text>
                                    <Text>
                                      Some text
                                    </Text>
                                  </View>
                                </RCTScrollView>
                              </View>
                            </View>
                          </RNSScreen>
                        </RNSScreenStack>
                      </View>
                    </View>
                  </View>
                </RNSScreen>
              </RNSScreenNavigationContainer>
              <View
                collapsable={false}
                onLayout={[Function handleLayout]}
                pointerEvents="auto"
              >
                <View
                  pointerEvents="none"
                />
                <View
                  accessibilityRole="tablist"
                >
                  <View
                    accessibilityRole="button"
                    accessibilityState={
                      Object {
                        "busy": undefined,
                        "checked": undefined,
                        "disabled": undefined,
                        "expanded": undefined,
                        "selected": true,
                      }
                    }
                    accessibilityStates={
                      Array [
                        "selected",
                      ]
                    }
                    accessibilityValue={
                      Object {
                        "max": undefined,
                        "min": undefined,
                        "now": undefined,
                        "text": undefined,
                      }
                    }
                    accessible={true}
                    collapsable={false}
                    focusable={true}
                    onBlur={[Function onBlur]}
                    onClick={[Function onClick]}
                    onFocus={[Function onFocus]}
                    onResponderGrant={[Function onResponderGrant]}
                    onResponderMove={[Function onResponderMove]}
                    onResponderRelease={[Function onResponderRelease]}
                    onResponderTerminate={[Function onResponderTerminate]}
                    onResponderTerminationRequest={[Function onResponderTerminationRequest]}
                    onStartShouldSetResponder={[Function onStartShouldSetResponder]}
                  >
                    <View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Home
                    </Text>
                  </View>
                  <View
                    accessibilityRole="button"
                    accessibilityState={
                      Object {
                        "busy": undefined,
                        "checked": undefined,
                        "disabled": undefined,
                        "expanded": undefined,
                        "selected": false,
                      }
                    }
                    accessibilityStates={Array []}
                    accessibilityValue={
                      Object {
                        "max": undefined,
                        "min": undefined,
                        "now": undefined,
                        "text": undefined,
                      }
                    }
                    accessible={true}
                    collapsable={false}
                    focusable={true}
                    onBlur={[Function onBlur]}
                    onClick={[Function onClick]}
                    onFocus={[Function onFocus]}
                    onResponderGrant={[Function onResponderGrant]}
                    onResponderMove={[Function onResponderMove]}
                    onResponderRelease={[Function onResponderRelease]}
                    onResponderTerminate={[Function onResponderTerminate]}
                    onResponderTerminationRequest={[Function onResponderTerminationRequest]}
                    onStartShouldSetResponder={[Function onStartShouldSetResponder]}
                  >
                    <View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Search
                    </Text>
                  </View>
                  <View
                    accessibilityRole="button"
                    accessibilityState={
                      Object {
                        "busy": undefined,
                        "checked": undefined,
                        "disabled": undefined,
                        "expanded": undefined,
                        "selected": false,
                      }
                    }
                    accessibilityStates={Array []}
                    accessibilityValue={
                      Object {
                        "max": undefined,
                        "min": undefined,
                        "now": undefined,
                        "text": undefined,
                      }
                    }
                    accessible={true}
                    collapsable={false}
                    focusable={true}
                    onBlur={[Function onBlur]}
                    onClick={[Function onClick]}
                    onFocus={[Function onFocus]}
                    onResponderGrant={[Function onResponderGrant]}
                    onResponderMove={[Function onResponderMove]}
                    onResponderRelease={[Function onResponderRelease]}
                    onResponderTerminate={[Function onResponderTerminate]}
                    onResponderTerminationRequest={[Function onResponderTerminationRequest]}
                    onStartShouldSetResponder={[Function onStartShouldSetResponder]}
                  >
                    <View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Discover
                    </Text>
                  </View>
                  <View
                    accessibilityRole="button"
                    accessibilityState={
                      Object {
                        "busy": undefined,
                        "checked": undefined,
                        "disabled": undefined,
                        "expanded": undefined,
                        "selected": false,
                      }
                    }
                    accessibilityStates={Array []}
                    accessibilityValue={
                      Object {
                        "max": undefined,
                        "min": undefined,
                        "now": undefined,
                        "text": undefined,
                      }
                    }
                    accessible={true}
                    collapsable={false}
                    focusable={true}
                    onBlur={[Function onBlur]}
                    onClick={[Function onClick]}
                    onFocus={[Function onFocus]}
                    onResponderGrant={[Function onResponderGrant]}
                    onResponderMove={[Function onResponderMove]}
                    onResponderRelease={[Function onResponderRelease]}
                    onResponderTerminate={[Function onResponderTerminate]}
                    onResponderTerminationRequest={[Function onResponderTerminationRequest]}
                    onStartShouldSetResponder={[Function onStartShouldSetResponder]}
                  >
                    <View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                      <View>
                        <View
                          accessible={false}
                        >
                          <Image
                            accessibilityIgnoresInvertColors={false}
                            source={
                              Object {
                                "process": [Function process],
                              }
                            }
                          />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Profile
                    </Text>
                  </View>
                </View>
              </View>
            </View>
          </View>
        </View>
      </RNSScreen>
    </RNSScreenStack>
  </RNCSafeAreaProvider>
</RNCSafeAreaView>
```

</details>

<details>
<summary>Debug output: visibility props, testID, some accessibility-related (173 lines)</summary>

This output is a lot easier to parse. It includes most of the props listed in the PR review, except `acessibilityState`, `accessibilityValue`,`placeholder`, `value`, `defaultValue`, `nativeId`.

```jsx
<RNCSafeAreaView>
  <RNCSafeAreaProvider>
    <RNSScreenStack>
      <RNSScreen>
        <RNSScreenStackHeaderConfig />
        <View
          accessibilityElementsHidden={false}
          importantForAccessibility="auto"
        >
          <View>
            <View>
              <RNSScreenNavigationContainer>
                <RNSScreen>
                  <View>
                    <View>
                      <View>
                        <RNSScreenStack>
                          <RNSScreen>
                            <RNSScreenStackHeaderConfig />
                            <View
                              accessibilityElementsHidden={true}
                              importantForAccessibility="no-hide-descendants"
                            >
                              <View>
                                <RCTScrollView
                                  testID="HomeScreen"
                                >
                                  <View>
                                    <Text
                                      accessibilityRole="header"
                                    >
                                      Good Morning, 
                                      <Text>
                                        Dale
                                      </Text>
                                    </Text>
                                    <Text>
                                      Welcome to the Example app. Content will be coming soon!
                                    </Text>
                                    <View
                                      accessibilityRole="link"
                                    >
                                      <Text>
                                        Go to article!
                                      </Text>
                                    </View>
                                    <View>
                                      <Text>
                                        Some text
                                      </Text>
                                    </View>
                                  </View>
                                </RCTScrollView>
                              </View>
                            </View>
                          </RNSScreen>
                          <RNSScreen>
                            <RNSScreenStackHeaderConfig />
                            <View>
                              <View>
                                <RCTScrollView
                                  testID="ArticleScreen"
                                >
                                  <View>
                                    <Text
                                      accessibilityRole="header"
                                    >
                                      Article Heading
                                    </Text>
                                    <Text>
                                      Some text
                                    </Text>
                                  </View>
                                </RCTScrollView>
                              </View>
                            </View>
                          </RNSScreen>
                        </RNSScreenStack>
                      </View>
                    </View>
                  </View>
                </RNSScreen>
              </RNSScreenNavigationContainer>
              <View>
                <View />
                <View
                  accessibilityRole="tablist"
                >
                  <View
                    accessibilityRole="button"
                  >
                    <View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Home
                    </Text>
                  </View>
                  <View
                    accessibilityRole="button"
                  >
                    <View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Search
                    </Text>
                  </View>
                  <View
                    accessibilityRole="button"
                  >
                    <View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Discover
                    </Text>
                  </View>
                  <View
                    accessibilityRole="button"
                  >
                    <View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                      <View>
                        <View>
                          <Image />
                        </View>
                      </View>
                    </View>
                    <Text>
                      Profile
                    </Text>
                  </View>
                </View>
              </View>
            </View>
          </View>
        </View>
      </RNSScreen>
    </RNSScreenStack>
  </RNCSafeAreaProvider>
</RNCSafeAreaView>
```

</details>


With no `mapProps` function, the output was 1272 lines long. VS Code truncated this output, though I was able to see the full output when running the test in my terminal. I didn't include the output here.

mdjastrzebski:
re. `accessibilityState`, `accessibilityValue`: in case of these we could filter out `undefined` props/empty objects.

So a part from your example:
```
accessibilityState={
  Object {
    "busy": undefined,
    "checked": undefined,
    "disabled": undefined,
    "expanded": undefined,
    "selected": false,
  }
}
accessibilityValue={
  Object {
    "max": undefined,
    "min": undefined,
    "now": undefined,
    "text": undefined,
  }
}
```
                    
Could become
```
// all `undefined props has been skipped
accessibilityState={
  Object {
    "selected": false,
  }
}
// accessibilityValue={Object { }} not to be displayed because of all undefined props`
```

stevehanson:
@mdjastrzebski your proposal of filtering out undefined object values and not printing those props altogether if all values are undefined seems like a huge improvement. I'll go ahead and move forward with updating this PR with the comments from your review, including what we just discussed. Thanks again!

src/helpers/mapProps.ts
mdjastrzebski: It terms of actual code we could have a list of allowed props to pass in:

```
const propsToDisplay = ['testID', 'nativeID', ... ];
```

For case when we would want to make some extra processing, like `accessibilityState`/`accessibilityValue` (remove `undefined` entries), we could add manual made `if` statements.

src/helpers/mapProps.ts
stevehanson: Updated to use this naming.

src/helpers/mapProps.ts
stevehanson: Updated to map this logic.

src/queries/__tests__/text.test.tsx
```diff
@@ -496,13 +496,118 @@ test('byText support hidden option', () => {
...
-  ).toThrowErrorMatchingInlineSnapshot(
-    `"Unable to find an element with text: /hidden/i"`
+  expect(() => getByText(/hidden/i, { includeHiddenElements: false })).toThrow(
+    /Unable to find an element with text: \/hidden\/i/
   );
 });
 
+describe('byText error message', () => {
+  test('renders the React DOM with regular props stripped', async () => {
+    const { getByText } = render(
```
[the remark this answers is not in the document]
stevehanson: Updated to add tests for the query variants plus all the other queries (eg. `byLabel`, etc.), now that this PR scope includes those.

src/queries/text.ts
stevehanson: I've now moved this formatting to `makeQueries`.

stevehanson:
@mdjastrzebski I believe I have implemented all of your suggestions, so this should be ready for re-review. Thank you again for all of your guidance on this PR 🙂

src/__tests__/host-component-names.test.tsx
```diff
@@ -87,7 +87,7 @@ describe('configureHostComponentNamesIfNeeded', () => {
       .toThrowErrorMatchingInlineSnapshot(`
       "Trying to detect host component names triggered the following error:
 
-      Unable to find an element with testID: text
+      \`render\` method has not been called
```
stevehanson, line 90: This error message changing is an unintended side effect. Basically when the `getByTestID` call fails in this test, it tries to print the DOM now, and `screen.toJSON()` fails because we haven't actually called render. 

I'm not sure if we would get this error in the real world or if it even matters that the message changed, since it didn't seem like a message that would have made sense to the user in the first place (the `testID=text` code is all internal to this library).

src/waitFor.ts
```diff
@@ -177,7 +177,10 @@ function waitForInternal<T>(
         }
       }
       if (typeof onTimeout === 'function') {
-        onTimeout(error);
+        const result = onTimeout(error);
+        if (result) {
+          error = result;
```
stevehanson, line 182: I could probably use a more thorough eye on this change. When a custom `onTimeout` was passed to `waitFor`, we did not previously allow for this to alter the returned error. This change makes it so the error returned by the custom `onTimeout` is the error that is ultimately reported.

mdjastrzebski:
@stevehanson looks promising, as the PR is quite large I will need to spend some time reviewing it. So far I've reviewed `makeQueries.ts` and refactored slightly your code. 

I wonder if we could find way to avoid these weird numbers popping in the error messages:
```
[36m<Text[39m
```

stevehanson:
@mdjastrzebski it seems the recent commit removes syntax highlighting not just for the tests but for the actual output. I'm not sure if that was an intended side effect, but I find the syntax highlighting to be quite helpful when parsing large DOM trees.

Before | After
--|--
colors-before | colors-after

mdjastrzebski:
@stevehanson That change was intended but you are right, colors might help users analyse the errors. The reason it bothered me was that all error snapshots contained color control characters.

I've thing I've found a good solution here. The coloring of output will depend on `process.env.COLORS !== false`, this is based on RTLs approach. By default the element tree output will be colored, but for our own tests we will set `process.env.COLORS = false` in order to generate clean snapshots.

@stevehanson hope that works for both of us :-)

stevehanson:
I'm noticing a couple interesting things. If I add and run the following failing test:

```js
test('sample', async () => {
  const { findByText } = render(<View />);
  await findByText(/foo/);
});
```

I notice two issues: the output does not include the element tree like expected, and the wrong frame is highlighted in the stack trace:

Screenshot 2023-04-18 at 11 06 35 PM

If I catch the error in the test and call `console.error(error.message)`, the element tree is indeed included, so it just is not printing for some reason:

output of console.error, shows element tree

I think this is why our tests for `findBy` are passing, even though the actual output, which is what we ultimately care about, is incorrect.

The other issue I pointed out is that the stacktrace highlights the frame from `makeQueries` instead of the frame from the test failure (i.e. `findBy` in `makeQueries` is not removed from the stacktrace). This same issue is also present in `main`. I'm not sure if this affects any builds.

## Potential solution

I dug around the `waitFor` code a bit and see that it creates an error prior to any async logic because that apparently causes some issues with the stacktrace. I tried doing something similar in `findByQuery` in `makeQueries`, and it does fix the issue, but I don't yet fully understand _why_. I will explore this more to see if there is a simpler solution and to get a better grasp of this but wanted to surface this in case any more regular contributors have some initial thoughts. 

Here's the solution that worked for me:

```ts
  function findByQuery(instance: ReactTestInstance) {
    let error: Error; // instantiate the error that will ultimately be returned from the timeout

    return function findFn(
      predicate: Predicate,
      queryOptions?: Options & WaitForOptions,
      {
        onTimeout = (e: unknown) => {
          const timeoutError = e as Error;
          if (timeoutError.message) {
            // take just the message from the timeout error and return our error instead
            // we don't care about the timeoutError stacktrace since we called waitFor from here.
            error.message = formatErrorMessage(timeoutError.message, true);
          }

          return error;
        },
        ...waitForOptions
      }: WaitForOptions = {}
    ) {
      error = new ErrorWithStack('FIND_BY_ERROR', findFn); // initialize the error here, removing `findFn` from the trace
      const deprecatedWaitForOptions =
        extractDeprecatedWaitForOptions(queryOptions);

      // append formatted DOM to final error

      return waitFor(
        () =>
          getByQuery(instance, { printElementTree: true })(
            predicate,
            queryOptions
          ),
        {
          ...deprecatedWaitForOptions,
          ...waitForOptions,
          onTimeout,
        }
      );
    };
  }
``` 

Here is the now-correct failure output with the above code:

correct test failure output for a findBy query. Shows element tree and stacktrace

## Other things that did not work

I also tried some other approaches, but nothing else I tried worked. The first thing I tried was manually removing `findBy` from the stacktrace of the error that we catch in our `onTimeout`:

```ts
onTimeout = (e: unknown) => {
  const error = e as Error;

  if (error.message) {
    error.message = formatErrorMessage(error.message, true);
  }

  // remove findFn from stacktrace
  if (Error.captureStackTrace) {
    Error.captureStackTrace(error, findFn);
  }

  return error;
}
```

This had a weird effect. The element tree now prints, but the text is muted. Additionally, no stack trace is shown.

output with element tree muted and not stacktrace

I can push up the solution I shared above but wanted to first leave this comment and gather feedback while I also continue to explore this, because I feel like there might be a better solution.

stevehanson:
I just had a realization and prepended an update to my comment above with more info.

src/waitFor.ts
stevehanson: Resolved in 2a9fb5c. The behavior should now match DOM Testing Library where if a custom `onTimeout` is passed, that function receives the error without the element tree. The element tree is only added in the default `onTimeout` if no custom one is passed.

stevehanson:
> I think your solution looks promising. The output looks good. Apply this as a code change.

@mdjastrzebski great! I can push up the code from my previous comment if that is what you are referring to here, but I wanted to make sure you also saw my commit from yesterday ([2a9fb5c](https://github.com/callstack/react-native-testing-library/pull/1378/commits/2a9fb5c81f5e7bf1b99752171db990c62412af2d#diff-d6b22e35d957ed617d3bc40b9853f053777f7c1e3e850ad4d7c8860e0b87f542)). That commit is overall a simpler solution with less potential for regression -- when a timeout occurs, we just append the element tree to the message and also replace the original message on `error.stack`. I verified that this does fix the error with the DOM not being output when a failure occurs, however it does not fix the error I was seeing with the `findBy` frame still being present in the trace. Since that issue is also present in `main`, it seems preferable to fix it in a separate PR and keep this one focused on printing the element tree.

I did try `findBy` in a real project using the latest RN Testing Library and see that even though the `findFn` frame is still present, my application code is what gets highlighted, so it seems like a smaller issue than what I was seeing when running tests from this codebase. I wonder if Jest or Node has some functionality to not highlight frames from external dependencies or something like that:

findBy | waitFor
--|--
image | Screenshot 2023-04-20 at 4 38 45 PM

stevehanson:
@mdjastrzebski I just wanted to check in to see if there is anything else needed from me to push this PR forward. It was unclear to me from the most recent comment which solution you preferred I go with to resolve the `findBy*` element tree issue. The currently pushed up solution seems to me to be the simplest approach that resolves the issue, but I'm happy to consider alternatives. Thanks!

mdjastrzebski (approved):
🎉 Looks awesome!

@stevehanson thank you kindly for submitting this PR and all the hard work into polishing it. Our users and I really appreciate it.

mdjastrzebski:
This PR has been released in [v12.1.0](https://github.com/callstack/react-native-testing-library/releases/tag/v12.1.0) 🚀

PR #1385, opened 2023-04-12
fix: non-editable wrapped TextInput events

### LINKED ISSUE #1261: Bug: Non-editable TextInputs are updated via fireEvent.changeText() when wrapped within 2 components down
## Describe the bug

I have an issue very similar to https://github.com/callstack/react-native-testing-library/issues/476. However this reproduces once the `<TextInput />` is nested within 2 components.

[I created a Github repo to demonstrate this issue.](https://github.com/ylemberg/text-input-spec-example) 

## Expected behavior

A `<TextInput/>` that is nested within 2 components or more should not have the `fireEvent.changeText()` invoke the `onChangeText()` callback and the test in the above repo `"should not fire on non-editable SecondTextInputWrapper"` should pass

## Steps to Reproduce

1) Clone the repository https://github.com/ylemberg/text-input-spec-example
2) Run `yarn install`
3) Run `yarn test`

## Screenshots

Screen Shot 2022-12-19 at 5 50 27 PM

## Versions

``` 
"devDependencies": {
    "@babel/core": "^7.18.0",
    "@babel/preset-typescript": "^7.17.12",
    "@babel/runtime": "^7.18.0",
    "@testing-library/jest-native": "3.4.3",
    "@testing-library/react-native": "^11.5.0",
    "@types/jest": "^29.2.4",
    "babel-plugin-module-resolver": "^4.1.0",
    "cross-env": "^7.0.3",
    "jest": "26.6.3",
    "react": "17.0.2",
    "react-native": "0.66.5",
    "react-test-renderer": "^17.0.2",
    "typescript": "^4.2.0"
  }
```

https://github.com/ylemberg/text-input-spec-example/blob/main/package.json

### LINKED ISSUE #1384: Wrapped TextInput with editable set to false calls onFocus anyway
## Describe the bug
When `fireEvent` is called with wrapped TextInput(i.e with `styled-components`) that is disabled the event bubbles up.
For example: We want to check if the `TextInput` component is actually disabled we might check whether `onFocus` event handler was called or not.

## Expected behaviour
`onFocus` handler should not be called if the `TextInput` is disabled (aka `editable` prop set to `false`)

## Steps to Reproduce
https://github.com/TMaszko/FireEventBug
Install dependencies and run `yarn test`

## Versions

 npmPackages:
    @testing-library/react-native: ^12.0.1 => 12.0.1 
    react: 18.2.0 => 18.2.0 
    react-native: 0.71.6 => 0.71.6 

TMaszko:
### Summary

[Description by @mdjastrzebski]
This PR modifies the `fireEvent` handling for `TextInputs` by checking the `editable` prop status of the nearest touch responder of given element. The nearest touch responder is the nearest descendant element that is both host element as well as advertises as touch responder. One such element is `TextInput`.

The previous version of the code, checked the `editable` prop status on the element itself, which could lead to situations when `TextInput` wrappers would execute events, e.g. `changeText`, `focus`, etc on non-editable `TextInput`, because the check has been limited to host `TextInput` and wrapping it composite `TextInput` coming from RN package. Any additional `TextInput` wrapper would not have it's `editable` prop checked and would refer to host `TextInput` `editable` prop.

During experiments conducted on iOS and Android, I've discovered that most of the events are blocked by `editable={false}` prop, but some events like `layout`, `contentSizeChange`, etc are being called anyway. This PR includes special exclusion for such events.

Fixes #1384
Fixes #1261 

### Test plan
Green tests

### DIFF
--- a/src/__tests__/fireEvent-textInput.test.tsx
+++ b/src/__tests__/fireEvent-textInput.test.tsx
@@ -1,0 +1,131 @@
+import React from 'react';
+import { Text, TextInput, TextInputProps } from 'react-native';
+import { render, fireEvent } from '..';
+
+function WrappedTextInput(props: TextInputProps) {
+  return <TextInput {...props} />;
+}
+
+function DoubleWrappedTextInput(props: TextInputProps) {
+  return <WrappedTextInput {...props} />;
+}
+
+const layoutEvent = { nativeEvent: { layout: { width: 100, height: 100 } } };
+
+test('should fire only non-touch-related events on non-editable TextInput', () => {
+  const onFocus = jest.fn();
+  const onChangeText = jest.fn();
+  const onSubmitEditing = jest.fn();
+  const onLayout = jest.fn();
+
+  const view = render(
+    <TextInput
+      editable={false}
+      testID="subject"
+      onFocus={onFocus}
+      onChangeText={onChangeText}
+      onSubmitEditing={onSubmitEditing}
+      onLayout={onLayout}
+    />
+  );
+
+  const subject = view.getByTestId('subject');
+  fireEvent(subject, 'focus');
+  fireEvent.changeText(subject, 'Text');
+  fireEvent(subject, 'submitEditing', { nativeEvent: { text: 'Text' } });
+  fireEvent(subject, 'layout', layoutEvent);
+
+  expect(onFocus).not.toHaveBeenCalled();
+  expect(onChangeText).not.toHaveBeenCalled();
+  expect(onSubmitEditing).not.toHaveBeenCalled();
+  expect(onLayout).toHaveBeenCalledWith(layoutEvent);
+});
+
+test('should fire only non-touch-related events on non-editable TextInput with nested Text', () => {
+  const onFocus = jest.fn();
+  const onChangeText = jest.fn();
+  const onSubmitEditing = jest.fn();
+  const onLayout = jest.fn();
+
+  const view = render(
+    <TextInput
+      editable={false}
+      testID="subject"
+      onFocus={onFocus}
+      onChangeText={onChangeText}
+      onSubmitEditing={onSubmitEditing}
+      onLayout={onLayout}
+    >
+      <Text>Nested Text</Text>
+    </TextInput>
+  );
+
+  const subject = view.getByText('Nested Text');
+  fireEvent(subject, 'focus');
+  fireEvent.changeText(subject, 'Text');
+  fireEvent(subject, 'submitEditing', { nativeEvent: { text: 'Text' } });
+  fireEvent(subject, 'layout', layoutEvent);
+
+  expect(onFocus).not.toHaveBeenCalled();
+  expect(onChangeText).not.toHaveBeenCalled();
+  expect(onSubmitEditing).not.toHaveBeenCalled();
+  expect(onLayout).toHaveBeenCalledWith(layoutEvent);
+});
+
+test('should fire only non-touch-related events on non-editable wrapped TextInput', () => {
+  const onFocus = jest.fn();
+  const onChangeText = jest.fn();
+  const onSubmitEditing = jest.fn();
+  const onLayout = jest.fn();
+
+  const view = render(
+    <WrappedTextInput
+      editable={false}
+      testID="subject"
+      onFocus={onFocus}
+      onChangeText={onChangeText}
+      onSubmitEditing={onSubmitEditing}
+      onLayout={onLayout}
+    />
+  );
+
+  const subject = view.getByTestId('subject');
+  fireEvent(subject, 'focus');
+  fireEvent.changeText(subject, 'Text');
+  fireEvent(subject, 'submitEditing', { nativeEvent: { text: 'Text' } });
+  fireEvent(subject, 'layout', layoutEvent);
+
+  expect(onFocus).not.toHaveBeenCalled();
+  expect(onChangeText).not.toHaveBeenCalled();
+  expect(onSubmitEditing).not.toHaveBeenCalled();
+  expect(onLayout).toHaveBeenCalledWith(layoutEvent);
+});
+
+test('should fire only non-touch-related events on non-editable double wrapped TextInput', () => {
+  const onFocus = jest.fn();
+  const onChangeText = jest.fn();
+  const onSubmitEditing = jest.fn();
+  const onLayout = jest.fn();
+
+  const view = render(
+    <DoubleWrappedTextInput
+      editable={false}
+      testID="subject"
+      onFocus={onFocus}
+      onChangeText={onChangeText}
+      onSubmitEditing={onSubmitEditing}
+      onLayout={onLayout}
+    />
+  );
+
+  const subject = view.getByTestId('subject');
+  fireEvent(subject, 'focus');
+  fireEvent.changeText(subject, 'Text');
+  fireEvent(subject, 'submitEditing', { nativeEvent: { text: 'Text' } });
+  fireEvent(subject, 'layout', layoutEvent);
+
+  expect(onFocus).not.toHaveBeenCalled();
+  expect(onChangeText).not.toHaveBeenCalled();
+  expect(onSubmitEditing).not.toHaveBeenCalled();
+  expect(onLayout).toHaveBeenCalledWith(layoutEvent);
+});
--- a/src/__tests__/fireEvent-textInput.test.tsx
+++ b/src/__tests__/fireEvent-textInput.test.tsx
@@ -72,6 +72,22 @@ test('should fire only non-touch-related events on non-editable TextInput with n
   expect(onLayout).toHaveBeenCalledWith(layoutEvent);
 });
 
+/**
+ * Historically there were problems with custom TextInput wrappers, as they
+ * could creat a hierarchy of three or more text input views with very similar
+ * event props.
+ *
+ * Typical hierarchy would be:
+ * - User composite TextInput
+ * - UI library composite TextInput
+ * - RN composite TextInput
+ * - RN host TextInput
+ *
+ * Previous implementation of fireEvent only checked `editable` prop for
+ * RN TextInputs, both host & composite but did not check on the UI library or
+ * user composite TextInput level, hence invoking the event handlers that
+ * should be blocked by `editable={false}` prop.
+ */
 test('should fire only non-touch-related events on non-editable wrapped TextInput', () => {
   const onFocus = jest.fn();
   const onChangeText = jest.fn();
@@ -101,6 +117,9 @@ test('should fire only non-touch-related events on non-editable wrapped TextInpu
   expect(onLayout).toHaveBeenCalledWith(layoutEvent);
 });
 
+/**
+ * Ditto testing for even deeper hierarchy of TextInput wrappers.
+ */
 test('should fire only non-touch-related events on non-editable double wrapped TextInput', () => {
   const onFocus = jest.fn();
   const onChangeText = jest.fn();
--- a/src/__tests__/fireEvent-textInput.test.tsx
+++ b/src/__tests__/fireEvent-textInput.test.tsx
@@ -74,8 +74,8 @@ test('should fire only non-touch-related events on non-editable TextInput with n
 
 /**
  * Historically there were problems with custom TextInput wrappers, as they
- * could creat a hierarchy of three or more text input views with very similar
- * event props.
+ * could creat a hierarchy of three or more composite text input views with
+ * very similar event props.
  *
  * Typical hierarchy would be:
  * - User composite TextInput
--- a/src/__tests__/fireEvent.test.tsx
+++ b/src/__tests__/fireEvent.test.tsx
@@ -216,44 +216,6 @@ test('should not fire on disabled Pressable', () => {
   expect(handlePress).not.toHaveBeenCalled();
 });
 
-test('should not fire on non-editable TextInput', () => {
-  const testID = 'my-text-input';
-  const onChangeTextMock = jest.fn();
-  const NEW_TEXT = 'New text';
-
-  const { getByTestId } = render(
-    <TextInput
-      editable={false}
-      testID={testID}
-      onChangeText={onChangeTextMock}
-    />
-  );
-
-  fireEvent.changeText(getByTestId(testID), NEW_TEXT);
-  expect(onChangeTextMock).not.toHaveBeenCalled();
-});
-
-test('should not fire on non-editable TextInput with nested Text', () => {
-  const placeholder = 'Test placeholder';
-  const onChangeTextMock = jest.fn();
-  const NEW_TEXT = 'New text';
-
-  const { getByPlaceholderText } = render(
-    <View>
-      <TextInput
-        editable={false}
-        placeholder={placeholder}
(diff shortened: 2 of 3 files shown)


src/__tests__/fireEvent.test.tsx
```diff
@@ -274,6 +275,44 @@ test('should not fire on non-editable TextInput with nested Text', () => {
   expect(onChangeTextMock).not.toHaveBeenCalled();
 });
 
+test.only('should not fire onFocus on non-editable TextInput wrapped', () => {
```
mdjastrzebski, line 278: @TMaszko pls remove focussed test, as it disables all other tests in the file.

TMaszko: Yeah that is a leftover !

src/fireEvent.ts
```diff
@@ -56,6 +56,9 @@ const isEventEnabled = (
   eventName?: string
 ) => {
   if (isTextInput(element)) return element?.props.editable !== false;
+  if (isTextInput(touchResponder)) {
```
pierrezimmermannbam, line 59: I'm not too sure about this. It will fix this issue indeed but saying that no event should be fired if at any point we pass through a non editable TextInput is not really correct either. For instance if I did a fireEvent.press on a non editable TextInput that's wrapped by a Pressable the press would not be run while it should. This may be a lesser evil still but there may be some other problems that we haven't thought of

pierrezimmermannbam: In the end fireEvent is just a low level API that calls props and we try to do too much with it so it has many unexpected behavior. How I see it long term is to fix this kind of issues using the new API userEvent but it's still at its early stage. It's not ideal that we can change text of non editable TextInputs but maybe to test those cases the easiest would be to use a custom matcher instead

TMaszko: Tbh if we want to call `fireEvent.press` on a non editable TextInput that's wrapped by a Pressable the press shouldn't be noticed (at least that's the behaviour that user will experience and we already have in the browser)

thymikee: @pierrezimmermannbam could you post some test cases that you suspect wouldn't work so that @TMaszko could include in this PR? With `fireEvent` and handling Responder API we kinda reverse-engineer the event system of RN, which may actually change depending on the platform (especially web), so expect edge cases that would be great to understand and document

pierrezimmermannbam: The following test will fail

```tsx
test('can pres pressable that wraps non editable textinput', () => {
    const mockOnPress = jest.fn();

    const { getByPlaceholderText } = render(
      <Pressable onPress={mockOnPress}>
        <TextInput placeholder="hello" editable={false} />
      </Pressable>
    );

    fireEvent.press(getByPlaceholderText('hello'));

    expect(mockOnPress).toHaveBeenCalled();
  });
```

We consider all text inputs to be touch responders, so when we do a press targeting a TextInput, it is considered to be the touch responder, even though it may be non editable. So we fix a bug but introduce another, possibly more. I'm fine with this pr being merged, it looks like a lesser evil but I feel like we're trying to extend too much the fireEvent API so it will always be flawed and it should in my opinion be considered as just an API that calls a prop (in fact I think it should be renamed to fireProp). For the long term i believe we need the userEvent API to really fix those issues and keep the fireEvent API only as a way to call custom props.

pierrezimmermannbam: The main issue with fireEvent is that is has almost the same behavior no matter what prop we call, so we disable all events on non editable text inputs for instance, but we should still be able to call onPressIn

```tsx
test('can trigger onPressIn on disabled textinput', () => {
    const mockOnPressIn = jest.fn();

    const { getByPlaceholderText } = render(
      <TextInput
        placeholder="hello"
        editable={false}
        onPressIn={mockOnPressIn}
      />
    );

    fireEvent(getByPlaceholderText('hello'), 'pressIn');

    expect(mockOnPressIn).toHaveBeenCalled();
  });
```

This test fails as well but it shouldn't

TMaszko: I don't get why this test should pass. Isn't it a proper behaviour to block events when component is basically `disabled`?

pierrezimmermannbam: The editable prop controls wether the value of the input can be modified or not but it still responds to touch events when it's not editable. You can try to press a non editable text input, the onPressIn prop is called in both cases

mdjastrzebski: I've been curious to get the understanding of how the TexInput events actually work when running in real RN app. I've build a small experiments app to test it, and submitted it as PR #1391.

The relevant experiment code is here: https://github.com/callstack/react-native-testing-library/pull/1391/files#diff-800ebe5060dbf16ec52e1cf9c65e7c2fa0f0cd86af32dcfdf913e13576831a70

For `editable: false` scenario both `onPressIn` and `onPressOut` events are being triggered (at least on iOS):

```
 LOG  Event: pressIn {"changedTouches": [[Circular]], "identifier": 1, "locationX": 57.33332824707031, "locationY": 35, "pageX": 77.33332824707031, "pageY": 146, "target": 117, "timestamp": 707054389.9047917, "touches": [[Circular]]}
 LOG  Event: pressOut {"changedTouches": [[Circular]], "identifier": 1, "locationX": 57.33332824707031, "locationY": 35, "pageX": 77.33332824707031, "pageY": 146, "target": 117, "timestamp": 707054477.4037917, "touches": []}
```

I'm yet to prepare experiments involving `Pressables`.

mdjastrzebski: I've also created a second test case, when non-editable `TextInput` is wrapped in `Pressable`:

```
<Pressable
        onPress={handlePressablePress}
        onPressIn={handlePressablePressIn}
        onPressOut={handlePressablePressOut}
        style={styles.pressable}
      >
        <TextInput
          style={styles.textInput}
          value={value}
          onChangeText={handleChangeText}
          editable={false}
          onPressIn={handlePressIn}
          onPressOut={handlePressOut}
        />
      </Pressable>
```

In this cases, when I press on the `TextInput` I get following events on iOS:
```
 LOG  Event: pressIn {"changedTouches": [[Circular]], "identifier": 1, "locationX": 165.66665649414062, "locationY": 16.333328247070312, "pageX": 205.66665649414062, "pageY": 147.3333282470703, "target": 65, "timestamp": 717273074.8629583, "touches": [[Circular]]}
 LOG  Event: pressOut {"changedTouches": [[Circular]], "identifier": 1, "locationX": 165.66665649414062, "locationY": 16.333328247070312, "pageX": 205.66665649414062, "pageY": 147.3333282470703, "target": 65, "timestamp": 717273155.2179583, "touches": []}
```
So non-editable `TextInput` normally receives `pressIn` and `pressOut` events, which seems to "consume" them, and prevent the press from reaching the wrapping `Pressable`. Running the same test with editable `TextInput` similarly "consumes" the touch event.

mdjastrzebski:
Based on the experiments described [here](https://github.com/callstack/react-native-testing-library/pull/1385#discussion_r1174593126), and [here](https://github.com/callstack/react-native-testing-library/pull/1385#discussion_r1174631810) I've came to the conclusion that state of `TextInputs` `editable` prop does seem to prevent some events on the `TextInput` (like `focus`) but not others (like `pressIn`).

I hope that will help us focus the review dicussion. Regarding the idea of checking touch responder `editable` prop in case of touch responder being `TextInput` that could still be a valid case but deserves more consideration, and probably some more tests.

As per @pierrezimmermannbam [comment](https://github.com/callstack/react-native-testing-library/pull/1385/files#r1165674065), the touch responder variable is being reset when walking up the element tree a new touch responder is encountered. So in the `Pressable` example given by Pierre, the host `View` rendered by `Pressable` would become the nearest touch responder when found. However, as shown above, this does not happen because even non-editable `TextInput` is still a valid touch responder for `pressIn` event.

mdjastrzebski:
@TMaszko Since `pressIn`/`pressOut` events are still being called for non-editable `TextInput` we should prepare a list of event names that will be ignored when `TextInput` is non-editable. Initially such events should include: `focus`, `blur`, `changeText`, `change`. Other elements may be added to the list if show in experimental manner using real RN app that they are ignored.

mdjastrzebski:
@TMaszko I've added tests based on #1261 that I believe should be part of this PR. They basically check if `focus`, `changeText`, `submitEditing` events are prevented on non-editable `TextInput` while `layout` event is allowed. We would need to add some kind of inclusion / exclusions list of events blocked by `editable={false}`.

pierrezimmermannbam:
@mdjastrzebski this is very interesting, I hadn't done tests on Android, it's quite surprising indeed! I'm wondering if we want to simulate all of this behavior, including the difference between platforms. I feel like we're hitting the limitations of testing in a node environment here, specially with TextInput component being mocked in React Native preset and it feels like we'd need to reimplement the TextInput for everything to work as expected so I'm not really sure it's worth it. Maybe we should just recommend using matchers from jest-native to test disabled pressables / text inputs? I guess the issue is that fireEvent already partially handles non editable  text inputs or disabled pressables which gives the impression it handles all cases which is not true. Simplifying fireEvent could be an option but it would be a huge breaking change and it may not be understood so this may not be the best option. I think it would be better to implement all of this in userEvent instead and then keep fireEvent as just an API that calls props

mdjastrzebski:
@pierrezimmermannbam I am coming to the conclusion that if `editable` + touch propagation differs between the platforms, then the users should not rely on it working either iOS or Android way, as that would work only on as single platform. Therefore, we should be able to simulate it either way. It seems like going with Android approach would just be simpler, as we could shutdown all touch-related events when `editable == false`.

In the real world of RN apps, putting `Pressable` (or any other touch consuming container) around `TextInput` is rather rare, so we should not worry too much about this.

As for this PR, it seems to be improving our simulation by basing the `editable` check on nearest touch responder (i.e. host TextInput), rather than composite or host element. Let's add necessary tests, and perform necessary research using the new experiments app, but otherwise it seems like the right direction.

mdjastrzebski:
I've prepared the PR for review:
* rebased on current main
* moved text-input related test to separate file, as they started to grow
* added support for allowing events not affected by `editable` prop to be called, e.g. `layout`
* refactored the `TextInput` event check to take into account only nearest touch responder instead of checking the element itself
* cleaned up the code

The subject is quite complex, so I am looking forward for a challenging review @pierrezimmermannbam, @MattAgn, @AugustinLF, @thymikee, @TMaszko :-)

src/__tests__/fireEvent-textInput.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { Text, TextInput, TextInputProps } from 'react-native';
+import { render, fireEvent } from '..';
+
+function WrappedTextInput(props: TextInputProps) {
+  return <TextInput {...props} />;
+}
+
+function DoubleWrappedTextInput(props: TextInputProps) {
+  return <WrappedTextInput {...props} />;
+}
```
MattAgn, line 11: @mdjastrzebski i don't fully understand the use of WrappedTextInput / DoubleWrappedTextInput, could you explain it to me?

src/__tests__/fireEvent-textInput.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+const layoutEvent = { nativeEvent: { layout: { width: 100, height: 100 } } };
+
+test('should not fire only touch-related events on non-editable TextInput', () => {
```
TMaszko, line 15:
```suggestion
test('should fire only non-touch-related events on non-editable TextInput', () => {
```

mdjastrzebski: yeah, double negation is hard 🤣

src/__tests__/fireEvent-textInput.test.tsx
mdjastrzebski: The current `editable` check would work only on host or composite `TextInput` exported from RN:

```typescript
const isTextInput = (element?: ReactTestInstance) => {
  if (!element) {
    return false;
  }

  return (
    filterNodeByType(element, TextInput) ||
    filterNodeByType(element, getHostComponentNames().textInput)
  );
};

if (isTextInput(element)) return element?.props.editable !== false;
``` 

When user would create their own `TextInput` wrapper or use such component from some UI library, then the `editable` prop would not be taken into account, as that wrapped `TextInput` would not pass the `isTextInput` check.
That would cause all of the events on such wrapper `TextInput` to be called regardless of `editable` prop.

mdjastrzebski: It's related to host/composite components but also to nearest touch responder. @AugustinLF do you have any suggestions that would make test descriptions better?

AugustinLF:
>  It seems like going with Android approach would just be simpler, as we could shutdown all touch-related events when editable == false.
That sounds like a fair tradeoff to me, we should however make sure that we document it.

Overall I agree that it seems like this PR makes it better. It's not perfect, but it's going to be a bit better.

src/fireEvent.ts
```diff
@@ -1,31 +1,18 @@
...
-  if (!element) {
-    return false;
-  }
-
-  // We have to test if the element type is either the `TextInput` component
-  // (for composite component) or the string "TextInput" (for host component)
-  // All queries return host components but since fireEvent bubbles up
-  // it would trigger the parent prop without the composite component check.
-  return (
-    filterNodeByType(element, TextInput) ||
```
TMaszko, line 20: Please correct me if I'm wrong but we can remove this check since all queries return host components(AFAIR that has been added recently)

mdjastrzebski: Yes, the composite `TextInput` check is being removed, and we only leave the host `"TextInput"` check.

src/__tests__/fireEvent-textInput.test.tsx
```diff
@@ -41,6 +41,37 @@ test('should not fire only touch-related events on non-editable TextInput', () =
   expect(onLayout).toHaveBeenCalledWith(layoutEvent);
 });
 
+test('should not fire only touch-related events on non-editable TextInput with nested Text', () => {
```
pierrezimmermannbam: I don't really understand this test case, what difference does it make that there is a nested text? You're still targeting the TextInput through a testID

pierrezimmermannbam:
> > It seems like going with Android approach would just be simpler, as we could shutdown all touch-related events when editable == false.
> 
> That sounds like a fair tradeoff to me, we should however make sure that we document it.
> 
> Overall I agree that it seems like this PR makes it better. It's not perfect, but it's going to be a bit better.

I agree, I think this is a good tradeoff for now!

src/__tests__/fireEvent-textInput.test.tsx
mdjastrzebski: Fixed, now querying nested `Text`

mdjastrzebski (approved):
LGTM but I've got quite involved in the PR, so looking for more objective reviewers.

AugustinLF (approved):
Left a small clarification regarding the test name/comment, but nothing that prevents from merging. And since I don't have a good suggestion (I'm not sure what exactly this test is testing), it's good to be merged for me :)

src/__tests__/fireEvent-textInput.test.tsx
mdjastrzebski: Added relevant comments for tests.

PR #1386, opened 2023-04-13
Feat/user event press

pierrezimmermannbam:
### Summary

With @MattAgn  we've been working on the userEvent feature and this is a first proposition for an implementation of userEvent.press. It's not perfect yet but I'm pretty excited about it and it can be a support for discussion.

Unlike fireEvent, this uses only the host components. This has several benefits:
- Allows better and easier simulation of real touch events. We trigger events (onResponderGrant, onResponderRelease) rather than calling props. For instance we just have to change the press duration to trigger a longPress instead of a press.
- It's easier to work only with host components versus host & composite. For instance with the TextInputs we had several issues where it worked on host and not on composite or the opposite.
- We're working on a lower level so we don't have to code React Native again for good simulations so this makes the implementation way simpler and I feel more maintainable as well. 

Overall I feel this works really well and provides tons of advantages. It has also some drawbacks though:
- Since real touch events involve time, the implementation relies on the usage of fake timers but I think this should be the recommended approach. An alternative would be to make the API asynchronous and just wait but we don't want longPresses to take 500ms. But this means that it will be harder to use this feature for some users. Also I think we'd need to check if fake timers are enabled and throw if that's not the case, which I haven't done yet.
- The implementation can be a bit complex since it requires to dig into the react native code to see what happens at host component level. It should be easier to maintain though and if it provides users with a better abstraction I feel it's a good compromise to make.

Regarding the option pressDuration, I was thinking initially of making a userEvent.longPress API but since the delayLongPress can be modified by the user, it would have required to look at a composite components props, so I figured it would make more sense to think in terms of press duration rather than press / longPress. 

For now this doesn't handle the touch events (for instance onTouchStart prop would not be called), this handles only press events because pressability and gestures are handled by different classes in React Native. It can very well be added later though.

If you think it's promising then in a second time we'll need to start discussing the strategy for release and user adoption.

### Test plan

I added a bunch of test cases showcasing what props are called and how the pressDuration option works.

Curious to have your opinion on this @mdjastrzebski  @MattAgn  @thymikee  @AugustinLF


mdjastrzebski:
@pierrezimmermannbam, @MattAgn I'm really happy that you are working on this.

From reading the tests it seems that the current version of `userEvent.press` works with `Pressables`. Would it also work with other components that have press events, like `Touchable*` family, `Text`, etc? I think it's important for this API to be universal, so the user can safely invoke `userEvent.press` on any element where they could use `fireEvent.press` but with more realistic event simulation.

src/userEvent/touchEvents.ts
```diff
@@ -0,0 +1,53 @@
...
+  press(hostParentElement, options);
+};
+
+const triggerPressEvent = (
+  element: ReactTestInstance,
+  options: PressOptions = { pressDuration: 0 }
+) => {
+  act(() => {
+    element.props.onResponderGrant(defaultPressEvent);
+    jest.advanceTimersByTime(options.pressDuration);
```
mdjastrzebski, line 41: This seems to assume that user uses fake timers. Would we be able to support both fake timers and real timers?

src/userEvent/touchEvents.ts
```diff
@@ -0,0 +1,53 @@
...
+const defaultPressEvent = {
+  persist: jest.fn(),
+  nativeEvent: {
+    timestamp: new Date().getTime(),
+  },
+  currentTarget: { measure: jest.fn() },
+};
+
+export type PressOptions = {
+  pressDuration: number;
```
mdjastrzebski, line 15: Q: What are the use cases for `pressDuration` prop?

src/__tests__/userEvent/userEvent.press.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+  test('does not calls onPress when onLongPress is called', () => {
+    const mockOnLongPress = jest.fn();
+    const mockOnPress = jest.fn();
+
+    render(
+      <Pressable onLongPress={mockOnLongPress} onPress={mockOnPress}>
+        <Text>press me</Text>
+      </Pressable>
+    );
+    userEvent.press(screen.getByText('press me'), { pressDuration: 500 });
```
mdjastrzebski, line 183: I'm wondering whether having separate `userEvent.longPress` would be more ergonomic API than passing `pressDuration` above certain threshold, which might not be obvious.

As a quite inexact metaphor, the original User Event library has [both `click` and `dblClick`](https://testing-library.com/docs/user-event/convenience#clicks) (and `tripleClick` as well ;-)).

```suggestion
    userEvent.longPress(screen.getByText('press me'));
```

src/__tests__/userEvent/userEvent.press.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+      <Pressable onPress={mockOnPress}>
+        <Text>press me</Text>
+      </Pressable>
+    );
+    userEvent.press(screen.getByText('press me'));
+
+    expect(mockOnPress).toHaveBeenCalled();
+  });
+
+  test('calls onPressIn', () => {
```
mdjastrzebski, line 98: I would suggest writing a single longer test that would check the sequence of mocks being called when invoking `press` event, instead of separately checking `onPressIn`, `onPressOut` and `onPress`.

src/helpers/isHostElementPointerEventEnabled.ts
```diff
@@ -0,0 +1,20 @@
+import { ReactTestInstance } from 'react-test-renderer';
+import { getHostParent } from './component-tree';
+
+export const isHostElementPointerEventEnabled = (
```
mdjastrzebski, line 4: It would be helpful if this method contained code comment about how `pointerEvents` exactly work, how different values affect the enabled status of an element and its descenants.

src/__tests__/userEvent/userEvent.press.test.tsx
```diff
@@ -0,0 +1,188 @@
+import React from 'react';
+import { Pressable, Text, View } from 'react-native';
+import { render, screen, userEvent } from '../../pure';
+
+jest.useFakeTimers();
```
mdjastrzebski, line 5: If feasible it would be good to have test that also check real times behavior.

src/userEvent/touchEvents.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const press = (
```
mdjastrzebski, line 18: It seems that most of the original User Event functions return `Promise<void>`, not sure why though, looks like it has `delay` options that puts a pause between subsequent events to simulator user behaviour. I wonder if that would make sense for our implementation as well.

src/userEvent/touchEvents.ts
pierrezimmermannbam: In order to make it work with real timers, we'd need to make it asynchronous and wait for some time. It's a more accessible API but I don't think we should recommend it. If it takes 500ms to trigger a longPress, that's a huge problem that I think the API should not permit. Triggering events that depend on time should require using fake timers imo and I believe we should promote that

src/userEvent/touchEvents.ts
pierrezimmermannbam: The main use case is to simulate a long press. If press duration is greater or equal than the long press duration offset then it will call onLongPress and not onPress. You can look at the test cases for pressDuration, it should make things clearer

src/__tests__/userEvent/userEvent.press.test.tsx
pierrezimmermannbam: Yes I was thinking initially of making a press and longPress API but the thing is long press only exists on composite side of components. To implement it, we'd need to look at the value of the delayLongPress prop on the composite component and I wanted to only use host components to keep the implementation simple and closer to what happens at host level

src/__tests__/userEvent/userEvent.press.test.tsx
pierrezimmermannbam: I'm not in favor of supporting real timers for this for the reasons I explained in https://github.com/callstack/react-native-testing-library/pull/1386#discussion_r1172362950

src/userEvent/touchEvents.ts
pierrezimmermannbam: Not sure but I think it's async because it also works with real timers. At least that's the only way I think it could work. Here we only do a press so I don't see where we could add some delay, or maybe at the end? Having some delay while typing text would make sense probably. But then again the more we add delay simulation, the more this would be slow without fake timers so I really feel like we should advocate the usage of fake timers

pierrezimmermannbam:
> 
> From reading the tests it seems that the current version of `userEvent.press` works with `Pressables`. Would it also work with other components that have press events, like `Touchable*` family, `Text`, etc? I think it's important for this API to be universal, so the user can safely invoke `userEvent.press` on any element where they could use `fireEvent.press` but with more realistic event simulation.

This is based on the Pressability implementation in React Native so it should work with every components that is pressable, I'll make sure and add test cases for those

src/__tests__/userEvent/userEvent.press.test.tsx
mdjastrzebski: I agree that we should encourage users to use this library with fake timers, but I think we should still support using real times as some users might be reluctant to switch to fake timers, even if it would make their code run faster. Especially not all tests need to trigger `longPress` events.

src/userEvent/touchEvents.ts
mdjastrzebski: I agree that using fake timers is probably more useful/recommended approach, but I would be wary of not supporting users who would want to use real times with User Event.

The default touch duration for regular press could be kept at a low value (e.g. 50 ms). It seems that RTL User Event is having a `delay` option for specifying similar idea delay between keystrokes, etc and they use timeout of 0 ms by default. (https://testing-library.com/docs/user-event/options#delay)

Long press would probably used much more rarely than regular `press`.

src/userEvent/touchEvents.ts
mdjastrzebski: Hmmm, exposing `pressDuration` is looking like quite a low-level implementation detail. To use it properly it would require user to know the default press duration of `Pressable`. I think a simpler API would be to expose two events:

```
press: () => void,
longPress: ({ duration = 500 }) => void
```

As for regular presses, which are far more popular than long press, the duration does not matter that much, and for `longPresses` most of the users would probably stick with the default `pressDuration` value. If they tweaked it, then they would also have option to apply the theaked value in test.

src/__tests__/userEvent/userEvent.press.test.tsx
mdjastrzebski: Perhaps we could structure in the way where `press` is having a very brief duration (e.g. 0 ms, or 50 ms, etc), while `longPress` would accept `duration` option with the default value of 500 ms (the default for `Pressable`).

That would result in quite intuitive tests:

```
```

src/userEvent/touchEvents.ts
mdjastrzebski: RTL's User Event seems to use `wait` utility to handle both real and fake timers wait:
https://github.com/testing-library/user-event/blob/main/src/utils/misc/wait.ts

```
export function wait(config: Instance['config']) {
  const delay = config.delay
  if (typeof delay !== 'number') {
    return
  }
  return Promise.all([
    new Promise<void>(resolve => globalThis.setTimeout(() => resolve(), delay)),
    config.advanceTimers(delay),
  ])
}
```
Looks like `Promise` + `setTimeout` is used for real timers.

mdjastrzebski:
re Pressability: I am really curious if this will work with other components out of the box. That would be the lucky option, as we would handle all types of components with single flow. In the unlucky case that they need a separate implementation, I think we should still support them, e.g. by checking the element type `=== Pressable` and then going with pressable or basic flow (perhaps just invoking `fireEvent.press`, etc)

src/__tests__/userEvent/userEvent.press.test.tsx
pierrezimmermannbam: Indeed I think you're right. It should be doable to support both. In that case we should think of an appropriate way to recommend using fake timers. I'm thinking of a warning linking to a documentation, wdyt?

src/userEvent/touchEvents.ts
pierrezimmermannbam: Ok it's what I had in mind as well, I'll try to make that work. The only major downside is that the API will be async even for fake timers users but I guess it's not a major issue either

src/__tests__/userEvent/userEvent.press.test.tsx
pierrezimmermannbam: Yes that could work. The only thing that bothers me is that when using the API, longPress may be more interpreted as "it calls the onLongPressProp" rather than user is pressing for a long time. And if the delayLongPress is greater than 500ms then it won't be called which may be perceived as a bug. The major advantage of this is that most users may not know what the default delay is for react to consider it to be a long press. In the end it's a question of how much we want to abstract this. I'm wondering if it wouldn't be better to start with only the press API and then based on user feedback see if there is a need for a longPress API

src/userEvent/touchEvents.ts
mdjastrzebski: RTL's User Event also seems to be async and throws `awaits` in [every](https://testing-library.com/docs/user-event/intro/) [example](https://testing-library.com/docs/user-event/utility). So if we are modelling APIs after them, then let it be so.

src/userEvent/touchEvents.ts
pierrezimmermannbam: I don't agree with press duration being an implementation detail, it's directly related to the user action. In my opinion longPress will be interpreted like it triggers my onLongPress prop so I think it's more linked to implementation details. So I think it's a good addition to be able to change the duration, having an additional longPress api with a different default may be a good idea I'm not too sure on this

src/__tests__/userEvent/userEvent.press.test.tsx
mdjastrzebski: I think `longPress` is a much rare case than regular `press`, so I would optimise the API for regular `press`. Then for long press most of the users would probably stick to the default delay, so in the next order I would optimise the API for them. Only final, there will be people who use long press with custom duration, so IMO they are the edge case of an edge case. 

Therefore, making their API a bit more complex yet still possible seems like a good tradeoff, as most of the users will have simple and intuitive API.

BTW afaik we are neither calling `onPress` nor `onLongPress` directly. If users want direct `onLongPress` then they can use `fireEvent`

pierrezimmermannbam:
The initial implementation works with every touchable/pressable but it doesn't work this Text or TextInput. The reason for that is that the implementation of Text is mocked in the React Native preset so it's just a component with name Text that has the same props than the composite one. It's not the same for pressables because the View is mocked but not pressable directly. It would work if only the native part of the Text component was mocked and I don't know why it's done this way, I'm gonna ask on the React Native repo if they know and maybe open a PR but that would be a huge breaking change and I don't know yet what impacts it could have so I'd need to try it on my projects to see how tests behave. In the meantime we can have a special treatment for Text and TextInput as implemented in this pr but it's a shame

src/__tests__/userEvent/userEvent.press.test.tsx
mdjastrzebski: I think we might add warning if they use some long duration action that affects the performance of the script, like the 500ms or more longer touch. In case of brief actions like regular press we should not bother them

pierrezimmermannbam:
For the real timers :
- I have added support for real timers as well
- api is now async
- users get a warning when using press or longPress with real timers
- I have duplicated tests to have tests with both fake and real timers

There are some issues right now with the tests:
- we are testing long presses with real timers so it takes a lot of time (~5s). We can't shorten that time I think
- we have to duplicate the tests to run with both fake and real timers

I think the solution would be to have a jest config with two projects, one with fake timers enabled and another with real timers. This way every test would be run with both setups without having to duplicate them. Also we could have a command that would run only tests with fake or real timers so that while developing we're still able to have fast tests and run tests with real timers less frequently. If you think 

For longPress, I have reworked the API as @mdjastrzebski suggested:
- press does not have a duration option anymore (I let the default at 0 for now)
- I added longPress api that has a default duration of 500ms and a duration option

For Text and TextInput I have done a workaround for now, I still need to ask on the React Native repo if they want to change or not the way those are mocked but it will take time and we'll still need support for earlier version of RN either way so the workaround will stay for a while

@mdjastrzebski  @AugustinLF  @MattAgn  wdyt?

src/userEvent/utils/warnAboutRealTimers.ts
```diff
@@ -0,0 +1,6 @@
+export const warnAboutRealTimers = () => {
+  // eslint-disable-next-line no-console
+  console.warn(`It is not recommended to use userEvent without using fake timers
```
mdjastrzebski, line 3: I think it's good idea to warn in case using real timers would actually cause delays, e.g. when wait would be 50 ms or longer.

As far as I checked the User Event's code, their default wait is basically 0 (as in `setTimeout(..., 0)`, so there is only pause to process promises and events): https://github.com/testing-library/user-event/blob/main/src/setup/setup.ts#L24

src/userEvent/touchEvents.ts
```diff
@@ -0,0 +1,115 @@
...
+import { getHostParent } from '../helpers/component-tree';
+import { filterNodeByType } from '../helpers/filterNodeByType';
+import { isHostElementPointerEventEnabled } from '../helpers/isHostElementPointerEventEnabled';
+import { getHostComponentNames } from '../helpers/host-component-names';
+import { jestFakeTimersAreEnabled } from '../helpers/timers';
+import { DEFAULT_MIN_PRESS_DURATION } from './constants';
+import { warnAboutRealTimers } from './utils/warnAboutRealTimers';
+
+const defaultPressEvent = {
+  persist: jest.fn(),
```
mdjastrzebski, line 12: I wonder should we actually provide anything outside of `nativeEvent` key? Can that be useful in both real code and tests. @pierrezimmermannbam wdyt?

src/userEvent/touchEvents.ts
```diff
@@ -0,0 +1,115 @@
...
+  // doesn't implement the pressability class
+  // Thus we need to call the props directly on the host component
+  const isEnabledHostText =
+    filterNodeByType(element, getHostComponentNames().text) &&
+    !element.props.disabled;
+  const isEnabledTextInput =
+    filterNodeByType(element, getHostComponentNames().textInput) &&
+    element.props.editable !== false;
+
+  if (isEnabledHostText || isEnabledTextInput) {
```
mdjastrzebski, line 37: Seems like we should apply the delay here as well. Wdyt?

src/userEvent/touchEvents.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (areFakeTimersEnabled) {
+      jest.runOnlyPendingTimers();
+    } else {
+      await wait(DEFAULT_MIN_PRESS_DURATION);
+    }
+  });
+};
+
+const isEnabledTouchResponder = (element: ReactTestInstance) => {
```
mdjastrzebski, line 105: Q: In the fireEvent we also seem to check `onMoveShouldSetResponder`, does it make sense to include it in this check?

src/helpers/isHostElementPointerEventEnabled.ts
```diff
@@ -0,0 +1,26 @@
+import { ReactTestInstance } from 'react-test-renderer';
+import { getHostParent } from './component-tree';
+
+// pointerEvents controls whether the View can be the target of touch events.
+// 'auto': The View can be the target of touch events.
+// 'none': The View is never the target of touch events.
+// 'box-none': The View is never the target of touch events but its subviews can be
+// 'box-only': The view can be the target of touch events but its subviews cannot be
+// see the official react native doc https://reactnative.dev/docs/view#pointerevents
+export const isHostElementPointerEventEnabled = (
```
mdjastrzebski, line 10: The current version of `isPointerEventEnabled` is essentially the same, so you should be able to use it.

mdjastrzebski (commented):
Awesome job @pierrezimmermannbam I really like the direction where this is going. Supporting both advanced `Pressables`/etc, as well as simpler `Text`/etc is very useful.

The current version of the code, seems to use older User Event API (`userEvent.press()`), the latest, v14, User Event API is slightly different:
1. `const user = userEvent.setup();`
2. `user.press()`

While there are some pros of using the older version of the API, e.g. it's simpler, I think the way forward is to use the latest version of the API, in order to follow RTL's best practices and provide high degree of compatibility. It also has some useful features as configuring the User Event default params, etc.

I've been recently working on implementing User Event `type()` API, and I've got the `userEvent.setup()`, etc already implemented in #1396. Perhaps, I should extract that + some reusable utils like `wait` to a separate PR that could get reviewed and merged quickly so that we would be working on the same base API.

I am really happy how fast is the progress for User Event moving. Having `press()` and `type()` will cover the majority of the basic use cases.

mdjastrzebski (commented):
I've extracted the common User Event code in #1405. @pierrezimmermannbam pls take a look so that we have a common base API to add proper functions.

mdjastrzebski:
I've merged #1405 to main, @pierrezimmermannbam pls rebase 🚀

src/helpers/isHostElementPointerEventEnabled.ts
pierrezimmermannbam: I kept a single version of it

src/userEvent/touchEvents.ts
pierrezimmermannbam: I have added the minimum so that the implementation worked, persist and currentTarget are both used by React Native in Pressability so that's why they're here. For now there is no option to customize the event. I very rarely needed that option so I'm not sure what the use cases for that would be but it could be available in a v2 if users have a need for it

src/userEvent/touchEvents.ts
pierrezimmermannbam: Yes you're right, I added a delay before the press and also now wait before the press out

src/userEvent/touchEvents.ts
pierrezimmermannbam: This is more related to gestures handling, onMoveShouldSetResponder is used nowhere in the Pressability class, those are two separate things. For now this implementation of press doesn't trigger any gesture events but they should be added in an ulterior version as well

mdjastrzebski:
@pierrezimmermannbam I've got some thoughts/questions/tradeoffs about event timing to use. Pls check [this comment](https://github.com/callstack/react-native-testing-library/pull/1396#discussion_r1196233414) in the `type` PR, as the approach we choose should probably be consistent in both PR.s

pierrezimmermannbam:
@mdjastrzebski I have rebased. I have also added documentation and made some fixes, it should be good for the first version

mdjastrzebski:
@pierrezimmermannbam thanks for updating the PR 🚀.  I plan to review it next week, due to tight schedule right now.

website/docs/UserEvent.md
```diff
@@ -28,3 +28,39 @@ Creates User Event instances which can be used to trigger events.
 ### Options
 - `delay` - controls the default delay between subsequent events, e.g. keystrokes, etc.
 - `advanceTimers` - time advancement utility function that should be used for fake timers. The default setup handles both real and Jest fake timers.
+
+
+## `press()`
+
+```ts
+type(
```
mdjastrzebski, line 36:
```suggestion
press(
```

website/docs/UserEvent.md
```diff
@@ -28,3 +28,39 @@ Creates User Event instances which can be used to trigger events.
...
+  element: ReactTestInstance,
+): Promise<void>
+```
+
+Example
+```ts
+const user = userEvent.setup();
+
+await user.press(touchable);
```
mdjastrzebski, line 45: [nit] `element` name would seem more adequante here
```suggestion
await user.press(element);
```

website/docs/UserEvent.md
```diff
@@ -28,3 +28,39 @@ Creates User Event instances which can be used to trigger events.
...
+```
+
+This helper simulates a press on any pressable element, e.g. Pressable, Text or TextInput. Unlike fireEvent.press which is a simpler API that will only call the onPress prop, this simumlates the entire press event in a more realistic way by reproducing what really happens when a user presses a pressable component. This means for instance for onPressIn and onPressOut props will also be invoked.
```
mdjastrzebski, line 48:
```suggestion
This helper simulates a press on any pressable element, e.g. `Pressable`, `TouchableOpacity`, `Text`, `TextInput`, etc. Unlike `fireEvent.press()` which is a simpler API that will only call the `onPress` prop, this simulates the entire press event in a more realistic way by reproducing what really happens when a user presses an interface view. This will trigger additional events like `pressIn` and `pressOut`. 
```

website/docs/UserEvent.md
```diff
@@ -28,3 +28,39 @@ Creates User Event instances which can be used to trigger events.
...
+
+await user.press(touchable);
+```
+
+This helper simulates a press on any pressable element, e.g. Pressable, Text or TextInput. Unlike fireEvent.press which is a simpler API that will only call the onPress prop, this simumlates the entire press event in a more realistic way by reproducing what really happens when a user presses a pressable component. This means for instance for onPressIn and onPressOut props will also be invoked.  
+
+## `longPress()`
+
+```ts
+type(
```
mdjastrzebski, line 53:
```suggestion
longPress(
```

website/docs/UserEvent.md
```diff
@@ -28,3 +28,39 @@ Creates User Event instances which can be used to trigger events.
...
+  element: ReactTestInstance,
+  options: { pressDuration: number } = { pressDuration: 500 }
```
mdjastrzebski, line 55: [nit] Perhaps simple `duration` would be enough, as "press" should be evident from the context.

```suggestion
  options: { duration: number } = { duration: 500 }
```

website/docs/UserEvent.md
```diff
@@ -28,3 +28,39 @@ Creates User Event instances which can be used to trigger events.
...
+): Promise<void>
+```
+
+Example
+```ts
+const user = userEvent.setup();
+
+await user.longPress(touchable);
```
mdjastrzebski, line 63: ditto: 
```suggestion
await user.longPress(element);
```

mdjastrzebski:
@pierrezimmermannbam after reviewing the public API I think the most ergonomic API would be:

```
function press(element, options = { duration: 0 });
function longPress(element, options = { duration: 500 });
```

So `longPress` would be exactly like `press` but with default duration of 500 ms, as this matches the default RN long press delay. In this pattern users could use `longPress()` API to alter the default press duration as well as to clarify the intent. While they could use `press()` API with arbitrary duration to trigger either `press` or `longPress` events.

@pierrezimmermannbam wdyt?

src/helpers/isPointerEventEnabled.ts
mdjastrzebski, line 1: [nit] `pointer-events.ts`?

src/helpers/isPointerEventEnabled.ts
```diff
@@ -0,0 +1,26 @@
+import { ReactTestInstance } from 'react-test-renderer';
+import { getHostParent } from './component-tree';
+
+// pointerEvents controls whether the View can be the target of touch events.
```
mdjastrzebski, line 4: Pls replace with TSDocs comment, so that VSCode will show intellisense on that method:
```/** ... */```

File: src/helpers/isPointerEventEnabled.ts [hunk 20]
```diff
@@ -0,0 +1,26 @@
...
+// 'auto': The View can be the target of touch events.
```
mdjastrzebski, line 5:
```suggestion
// 'auto': The View and it's children can be the target of touch events.
```

src/user-event/event-builder/common.ts
```diff
@@ -45,4 +45,14 @@ export const CommonEventBuilder = {
...
     };
   },
+
+  press: () => {
+    return {
+      persist: jest.fn(),
+      nativeEvent: {
+        timestamp: new Date().getTime(),
+      },
+      currentTarget: { measure: jest.fn() },
```
mdjastrzebski, line 55: Why

src/user-event/event-builder/common.ts
```diff
@@ -45,4 +45,14 @@ export const CommonEventBuilder = {
       },
     };
   },
+
+  press: () => {
+    return {
+      persist: jest.fn(),
+      nativeEvent: {
+        timestamp: new Date().getTime(),
```
mdjastrzebski, line 53: Perhaps we could provide some `setEventTimestamp` function to inject timestamp value or function generating timestamp values so that we could have easier time working with snapshots? Wdyt?

src/test-utils/events.ts
```diff
@@ -18,3 +18,7 @@ export function createEventLogger() {
 
   return { events, logEvent };
 }
+
+export function getEventsName(events: EventEntry[]) {
```
mdjastrzebski, line 22: Is it worth it? Maybe we could expose `eventNames` next to `events` in the `createEventLogger` return. Wdyt?

src/user-event/press/__tests__/longPress.real-timers.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const user = userEvent.setup();
+
+    render(
+      <Pressable onLongPress={mockOnLongPress}>
+        <Text>press me</Text>
+      </Pressable>
+    );
+    await user.longPress(screen.getByText('press me'));
+
+    expect(mockOnLongPress).toHaveBeenCalled();
```
mdjastrzebski, line 27: Let's add check for `mockWarnAboutRealTimers` in this test instead of having a separate one just for this warning.

src/user-event/press/__tests__/longPress.real-timers.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  test('calls onLongPress when duration is greater than specified onLongPressDelay', async () => {
```
mdjastrzebski, line 30:
```suggestion
  test('calls onLongPress when duration is greater than specified longPressDelay', async () => {
```

src/user-event/press/__tests__/longPress.real-timers.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+        <Text>press me</Text>
+      </Pressable>
+    );
+    await user.longPress(screen.getByText('press me'));
+
+    expect(mockOnLongPress).toHaveBeenCalled();
+    expect(mockOnPress).not.toHaveBeenCalled();
+  });
+
+  test('longPress is accessible directly in userEvent', async () => {
```
mdjastrzebski, line 73: This test seems redundant, as all other tests test the same thing. Wdyt?

src/user-event/press/__tests__/longPress.real-timers.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Pressable, Text } from 'react-native';
+import { render, screen } from '../../../pure';
+import { userEvent } from '../..';
+import * as WarnAboutRealTimers from '../utils/warnAboutRealTimers';
+
+const mockWarnAboutRealTimers = jest
+  .spyOn(WarnAboutRealTimers, 'warnAboutRealTimers')
+  .mockImplementation();
+
+describe('userEvent.longPress with real timers', () => {
```
mdjastrzebski, line 11: Pls add tests using `test-utils/events` so that we can verify the whole sequence of events, including their payloads. This could replace the existing `jest.fn()` for both called/not called scenarios, as event name will either be present or absent in the events array.

src/user-event/press/__tests__/press.real-timers.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+        <Text>press me</Text>
+      </Pressable>
+    );
+
+    await userEvent.press(screen.getByText('press me'));
+
+    expect(mockOnPress).toHaveBeenCalled();
+  });
+
+  test('warns about using real timers with userEvent', async () => {
```
mdjastrzebski, line 297: [discussion, tradeoffs] I think we should allow the regular `press` to be called without warning for real timers, this will decrease slightly how realistic this simulation is but it will ease adoption by the users since it will not require the to switch to fake timers, which migth be a hurdle for some of the users.

Perhaps we could enable more realistic mode for advanced users using some `userEvent` option or by detecting fake timers?

@pierrezimmermannbam wdyt?

src/user-event/press/__tests__/press.test.tsx
```diff
@@ -1,67 +1,291 @@
...
+        pointerEvents="box-none"
+        onPressIn={logEvent('pressIn')}
+        onPressOut={logEvent('pressOut')}
+      />
+    );
+    await userEvent.press(screen.getByPlaceholderText('email'));
+    expect(events).toEqual([]);
+  });
+
+  test('press is accessible directly in userEvent', async () => {
```
mdjastrzebski, line 278: This test seems redundant.

src/user-event/press/__tests__/press.real-timers.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+      <Pressable
+        onPress={logEvent('press')}
+        onPressIn={logEvent('pressIn')}
+        onPressOut={logEvent('pressOut')}
+        testID="pressable"
+      />
+    );
+    await user.press(screen.getByTestId('pressable'));
+
+    expect(getEventsName(events)).toEqual(['pressIn', 'press', 'pressOut']);
```
mdjastrzebski, line 38: IMO some of the core tests should also test the event payload. Not all of the tests but just the core cases.

src/user-event/press/utils/warnAboutRealTimers.ts
```diff
@@ -0,0 +1,6 @@
+export const warnAboutRealTimers = () => {
+  // eslint-disable-next-line no-console
+  console.warn(`It is not recommended to use userEvent without using fake timers
+Some events involve duration so your tests may take a long time to run.
+For instance calling userEvent.longPress with real timers will take 500ms`);
```
mdjastrzebski, line 5:
```suggestion
For instance calling userEvent.longPress with real timers will take 500 ms`);
```

website/docs/UserEvent.md
pierrezimmermannbam: Yes i agree, I changed the option name

src/test-utils/events.ts
pierrezimmermannbam: I find it worth it yes, after writing the same map several times I felt we needed an additional helper to do this. I like the idea of returning `eventNames` directly from `createEventLogger ` though, I'll make the change

src/user-event/event-builder/common.ts
mdjastrzebski: The idea behind this function is non-exported test util for our own convenience, as we are potentially interested in comparing event sequence with snapshot, etc. The users will probably not need it, unless they are doing some timing calculation, but in such case  they will just mock `Date.now()` etc. 

There are couple of ways we could achieve similar result:
1. Event timestamp setter function:
```
let getEventTimestamp = () => Date.now();
function setEventTimestamp(fn: () => Date) {
  getEventTimestamp = fn
}
```

2. Spying on event timestamp getter
```
function getEventTimestamp() {
  return Date.now();
}

jest.spyOn(UserEvent, "getEventTimestamp").mockImplementation(() => 0);
```

This option is to mock `Date.now()` in our tests, but that feels fragile compared to other approaches.

src/user-event/press/__tests__/longPress.real-timers.test.tsx
pierrezimmermannbam: I prefer having a separate test because the test name makes it very clear what we're testing. Also there is no connection between longPress being called and the warning being emitted so either could fail independently so I think they're better tested in isolation

src/user-event/press/__tests__/longPress.real-timers.test.tsx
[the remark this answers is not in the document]
pierrezimmermannbam: It's actually hidden in the same test, we start by doing a regular longPress with a threshold of 1000 ms and check than `onLongPress` has not been called. It's obviously unclear though so I'll split it in two tests

src/user-event/press/__tests__/longPress.real-timers.test.tsx
pierrezimmermannbam: The other tests use the return from the setup function, this checks that longPress is directly available in the userEvent object imported from the library

mdjastrzebski: ah, ok, I confused these two

src/user-event/event-builder/common.ts
pierrezimmermannbam: I don't think this approach could work. The issue is that we wait for the press duration between calling `onResponderGrant` and `onResponderRelease` and RN will call onLongPress or onPress based on the time difference between those two events. If we simply mock the date, then the timestamp will stay the same and longPress will not work anymore. What I would suggest is to write a test with fake timers enabled where we use `jest.setSystemTime` to check for the events payload and skip that test for real timers. Either that or write the same test and do an inline snapshot of events with timestamp set to 0 but I don't think there would be much value in that

src/test-utils/events.ts
pierrezimmermannbam: Actually I'm not sure it's a good idea because I can't return directly the eventNames because it wouldn't be updated whenever we push a new event, unless we'd have two arrays but that would be having two sources of truth so it does not look like a great idea. Another solution would be to make it so that `createEventLogger` returns a function getEventsName. It would probably be a bit better because we don't need to pass events as a parameter to the function, wdyt?

src/user-event/press/__tests__/press.test.tsx
pierrezimmermannbam: It is in a way yes, however it explicitly states that it tests that so I think it's still useful. It's possible that other tests are refactored to use the return of the `setup` function and this would not be tested anymore. I believe that every thing we want to test should be explicitly said in a test name, wdyt? If the intention behind every test is not explicit then it's very difficult to know which cases are covered and which aren't and I think having a bit of duplication is a small price to pay for that

src/user-event/press/press.ts
```diff
@@ -1,16 +1,118 @@
...
+    await wait(config);
+    if (onPressIn) {
+      onPressIn(EventBuilder.Common.press());
+    }
+    if (onPress) {
+      onPress(EventBuilder.Common.press());
+    }
+    await wait(config, options.duration);
+    if (onPressOut) {
+      if (DEFAULT_MIN_PRESS_DURATION - options.duration > 0) {
```
mdjastrzebski, line 61: Is that 130 ms happening experimentally in an actual RN app for `Text`/`TextInput`? We should consider adding it artificially if this is the case.

src/user-event/press/__tests__/press.real-timers.test.tsx
mdjastrzebski: If this is resulting from`Pressability` implementation then we should accept this 130 ms delay as a fact. Such delay for one test is insignificant but for numerous tests will quickly add up to meaningful numbers. 

Perhaps what we could do is to issue the warning only in `Pressability` branch.

pierrezimmermannbam: Yes it's a possibility but I fear it would make things more confusing for users and I don't see anyone that's not willing to use the API because they're not using fake timers use it only for Text and TextInputs, especially since using it on those will be much less common

src/user-event/press/__tests__/longPress.real-timers.test.tsx
pierrezimmermannbam: I have added a test that also checks payload events with fake timers. Like I explained in https://github.com/callstack/react-native-testing-library/pull/1386#discussion_r1211649723, I don't think it can be tested with real timers

src/user-event/press/__tests__/press.real-timers.test.tsx
pierrezimmermannbam: I now test the events payload in one of the tests with fake timers

src/user-event/press/press.ts
pierrezimmermannbam: Yes it happens also for `Text` and `TextInput`because they use the same implementation of `Pressability`. Not sure I understand your comment though, we're already adding it artificially here

pierrezimmermannbam:
@mdjastrzebski I think I addressed all your comments, could you please review when you find the time?

src/test-utils/events.ts
mdjastrzebski: Current version is ok.

src/user-event/press/press.ts
```diff
@@ -1,16 +1,118 @@
...
+    if (onPress) {
+      onPress(EventBuilder.Common.touch());
+    }
+    await wait(config, options.duration);
+    if (onPressOut) {
+      if (DEFAULT_MIN_PRESS_DURATION - options.duration > 0) {
+        await wait(config, DEFAULT_MIN_PRESS_DURATION - options.duration);
+      }
+      onPressOut(EventBuilder.Common.touch());
+    }
```
mdjastrzebski, line 65: [high] Looks like we need an early return here, otherwise it will continue propagation to the host parent.

src/user-event/press/press.ts
```diff
@@ -1,16 +1,118 @@
...
+    });
+
+    await wait(config, options.duration);
+
+    element.props.onResponderRelease({
+      ...EventBuilder.Common.touch(),
+      dispatchConfig: { registrationName: 'onResponderRelease' },
+    });
+
+    if (DEFAULT_MIN_PRESS_DURATION - options.duration > 0) {
```
mdjastrzebski, line 106: [mid] q: why do we need to do additional waiting to reach at least min press duration limit here and in the Text/TextInput block above?

I've found [this line](https://github.com/facebook/react-native/blob/06668fcbacd750771f1d53cce829dc55e86f3f3c/packages/react-native/Libraries/Pressability/Pressability.js#L955) in RN source. This seems to be an un-documented internal implementation detail of RN. I wonder if:
1) If it's already being applied so that we would not need to add manual wait
2) If not is it worth relying on such implementation details.

@pierrezimmermannbam wdyt?

src/user-event/press/press.ts
```diff
@@ -1,16 +1,118 @@
...
+      await wait(config, DEFAULT_MIN_PRESS_DURATION - options.duration);
+    }
+  });
+};
+
+const isEnabledTouchResponder = (element: ReactTestInstance) => {
+  return (
+    isPointerEventEnabled(element) &&
+    element.props.onStartShouldSetResponder &&
+    element.props.onStartShouldSetResponder()
```
mdjastrzebski, line 116: optional chaining should work here
```suggestion
    element.props.onStartShouldSetResponder?.()
```

src/user-event/press/utils/warnAboutRealTimers.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { warnAboutRealTimers } from './warnAboutRealTimers';
+
+const mockConsoleWarn = jest.spyOn(console, 'warn').mockImplementation();
+
+test('logs a warning about usign real timers with user event', () => {
+  warnAboutRealTimers();
+
+  expect(mockConsoleWarn.mock.calls[0][0]).toMatchInlineSnapshot(`
+    "It is not recommended to use userEvent without using fake timers
```
mdjastrzebski, line 9: [nit] Perhaps it's worth removing double negation: "It is recommended to use `userEvent` with fake timers."

src/user-event/press/utils/warnAboutRealTimers.ts
```diff
@@ -0,0 +1,6 @@
+export const warnAboutRealTimers = () => {
+  // eslint-disable-next-line no-console
+  console.warn(`It is not recommended to use userEvent without using fake timers
+Some events involve duration so your tests may take a long time to run.
+For instance calling userEvent.longPress with real timers will take 500 ms`);
```
mdjastrzebski, line 5:
```suggestion
For instance calling userEvent.longPress with real timers will take 500 ms.`);
```

src/user-event/press/__tests__/press.real-timers.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+    jest.useRealTimers();
+  });
+
+  test('calls onPressIn, onPress and onPressOut prop of touchable', async () => {
+    const { events, logEvent } = createEventLogger();
+    const user = userEvent.setup();
+
+    render(
+      <Pressable
+        onPress={logEvent('press')}
```
mdjastrzebski, line 30: [mid] Adding `onLongPress={logEvent('longPress')}` in each test, would allow us to make sure it's not actually invoked when not expected.

src/user-event/press/utils/warnAboutRealTimers.test.tsx
```diff
@@ -0,0 +1,13 @@
+import { warnAboutRealTimers } from './warnAboutRealTimers';
+
+const mockConsoleWarn = jest.spyOn(console, 'warn').mockImplementation();
+
+test('logs a warning about usign real timers with user event', () => {
```
mdjastrzebski, line 5: Testing this method seems redundant. In order to achieve code coverage it would be better to check if by calling `userEvent.press` with real times. That way would would at least check that it's being invoked from `press()` function. Here we are just confirming the logged message.

mdjastrzebski (changes requested):
Awesome work @pierrezimmermannbam! Also thank you for patience with that, as due to large amount of work and holiday time I was not very responsive lately.

I wonder if we should simulate the min 130 ms press delay by ourselves. First of all it looks like it might already be applied for us by Pressability code in RN (could you check that?). Second of all, it might create additional hurdle for adoption of `userEvent.press` as users would be nagged to enable fake timers with warnings even for default zero duration presses. Personally, I would rather have easier adoption for the less advanced users rather than increased reality due to 130 ms delay.

@thymikee @AugustinLF @MattAgn perhaps you can give your outsider perspective on how you see that trade off.

src/user-event/press/press.ts
pierrezimmermannbam: Oh you're right, good catch!

src/user-event/press/press.ts
pierrezimmermannbam: What happens in RN code when the press is finished is:

- the delay for the pressOut is computed: it's equal to `minimum press duration` -  `press duration`(see [source code](https://github.com/facebook/react-native/blob/4c944540f732c6055d447ecaf37d5c8f3eec1bc4/packages/react-native/Libraries/Pressability/Pressability.js#L810)
- if the delay is > 0 then pressOut is programmed through a setTimeout with the delay computed at the above step (see [source code](https://github.com/facebook/react-native/blob/4c944540f732c6055d447ecaf37d5c8f3eec1bc4/packages/react-native/Libraries/Pressability/Pressability.js#L814)) 

In the implementation for `Pressable`, we wait because if the press duration is shorter than the minimum press duration, then `onPressOut`has not been called yet. This is problematic for two reasons: 
- users would expect `onPressOut` to have been called
- it results in an act warning

This is why we wait in this case and in the second scenario, we could just call `onPressOut` without waiting, it would just be less realistic. The upside of not waiting is that tests that don't use fake timers would be faster

src/user-event/press/__tests__/press.real-timers.test.tsx
pierrezimmermannbam: I added `onLongPress={logEvent('longPress')}` in tests that use `logEvent`

src/user-event/press/utils/warnAboutRealTimers.test.tsx
pierrezimmermannbam: Yes I agree it's far from ideal, the reason I tested this separately is because I'm mocking here console warn to prevent warnings to appear in console when running tests and I didn't want to mock console.warn for every test in the userEvent.press tests. I did it a bit differently by putting the other tests in separate describe to have different mock just for them within the test file

pierrezimmermannbam:
Regarding the 130ms delay, it's not quite the same when pressing Text/TextInput and other Pressable like:

For Pressables, we don't directly call the `onPressOut` prop, we only call `onResponderRelease` and based on that RN uses a `setTimeout` to program the call to `onPressOut`. Here we don't really have the choice to wait I think. If we didn't, `onPressOut` would not be called synchronously which would be confusing imo and also the user would see an act warning if he doesn't have `waitFor` later on in his test.

For Text/TextInput it's a bit different. We call directly `onPressOut` so here we could skip the delay and call it directly. It would mean faster tests for real timers users but would make the simulation a bit less close to reality so here I'm not sure what the best tradeoff is. 

However I think that the main use cases for `press` and `longPress` will be to target `Pressable`. For those there will be a delay because it's the way RN works so I would be suprised to see users adopt `userEvent.press` while using real timers anyway, especially given that we log a warning. I agree that it's important to make it easy to adopt but this feature cannot work properly with real timers, I mean it can it will just be slow but that's maybe even worse so we should not encourage users to use it with real timers imo

src/user-event/press/press.ts
mdjastrzebski: Could you confirm that I get that right: the `Pressable` is actually delaying the action, so that we wait at least 130 ms, event when running under tests. If we do not wait to that 130 ms, then the event handler will be called outside of async `act` causing async `act` warning? So we are actually waiting to prevent async `act` warning from happening.

pierrezimmermannbam: Yes that's correct. We're waiting both to prevent act warning but also to have `onPressOut` called synchronously so user doesn't have to wait for it

mdjastrzebski:
It is not added to package-level exports yet, waiting for `userEvent.type()` PR.

src/userEvent/utils/warnAboutRealTimers.ts
henbrand: Hi! @pierrezimmermannbam @mdjastrzebski - sorry to dredge this up, i've just got a question on this as i'm not sure I understand.

My understanding is that fake timers are a good idea when timers are being used (setTimeout, setInterval, clearTimeout, clearInterval). 

By "some events" do you mean long press only?  if so could we then just show this warning if long press is used?

If not, what timer is happening? and do you recommend we use fake timers on every single test where there we need to press on something?

thanks!

pierrezimmermannbam: Hello @henbrand ! Actually there are timers involved in every press events. This is because in react native, the `onPressOut` callback can only be called at least 130ms after the `onPressIn`. We wait for the `onPressOut` to be called before resolving `userEvent.press` because it's closer to what the user experiences and makes writting tests easier so if you're not using fake timers then every press would last at least 130ms. 

And yes this is only a problem for press events, but this is due to how react native jest preset is made, pressables are not mocked at the same level than texts or textinputs so the mechanism that should be the same is not in practice. 

Regarding fake timers more generally I consider it a good practice to have them enabled by default in your whole test suite because it brings a lot of benefits (predictability, faster tests, control over timers) and I don't see any drawback to having them always enabled. In my experience they're not used by default because it may require a bit of work on the test config to make them work and there have been issues with RNTL before but now these issues have been fixed and it should not be that hard to activate them on a whole test suite so i highly recommend doing so

Hope this helps, let me know if you have any further questions!

PR #1391, opened 2023-04-23
chore: experiments app

mdjastrzebski:
### Summary

Add simple Expo app that could be used to conduct experiments on React Native behavior in a more structured and shareable way. Such experiments are frequently needed in order to build understanding necessary for improving simulation code inside RNTL.

### Test plan

All checks should pass.


pierrezimmermannbam (approved):
Great idea I really like this!

MattAgn:
Love the idea! 🚀

PR #1396, opened 2023-04-28
feat: User Event `type()`

mdjastrzebski:
### Summary

Minimal User Event `type()` implementation including initial User Event `setup()` API mimicking RTL's one.

Scope:
- Basic User Event API:
  - [x] `userEvent.setup` 
  - [x] `wait` utility
- `type()` API:
  - [x] Perform experiments on [single-line](https://github.com/callstack/react-native-testing-library/wiki/TextInput-Events#single-line---basic-flow) and [multi-line](https://github.com/callstack/react-native-testing-library/wiki/TextInput-Events#mutli-line---basic-flow) TextInputs
  - [x] Basic single-line implementation and tests
  - [x] Basic multi-line implementation and test
  - [x] Works correctly with both managed and un-managed TextInput
  - [x] Works correctly when managed `TextInputs` performs input transformation
  - [x] Support for Backspace button
  - [x] Support for fake timers (modern & legacy)
  - Edge cases:
    - [x] Restrict API to host `TextInput` elements
    - [x] Works correctly with not all event handlers attached
  - Docs:
    - [x] API Docs
    - [x] TSDoc comments (types and functions)

Exclusions:
- Simulating `target` and `eventCount` event props (just pass `0` there)

### Test plan

Added tests covering 100% specs


src/user-event/type/type.ts
```diff
@@ -1,20 +1,129 @@
...
 export async function type(
   this: UserEventInstance,
   element: ReactTestInstance,
-  text: string
-) {
-  // TODO provide real implementation
+  text: string,
+  options?: TypeOptions
+): Promise<void> {
+  if (element.type !== getConfig().hostComponentNames?.textInput) {
```
pierrezimmermannbam, line 26: To access host component names the function getHostComponentNames should be used instead

src/user-event/type/type.ts
```diff
@@ -1,20 +1,129 @@
...
+    await wait(this.config);
+    dispatchHostEvent(element, 'pressIn', EventBuilder.Common.touch());
+  }
+
   await wait(this.config);
   dispatchHostEvent(element, 'focus', EventBuilder.Common.focus());
 
-  await wait(this.config);
-  dispatchHostEvent(element, 'changeText', text);
+  if (options?.skipPress !== true) {
```
pierrezimmermannbam, line 43: nit: I prefer if (!options?.skipPress), it's more succinct and readable imo

src/user-event/type/type.ts
```diff
@@ -1,20 +1,129 @@
...
+    const previousText = element.props.value ?? currentText;
+    currentText = applyKey(previousText, key);
+
+    await wait(this.config);
+    emitTypingEvents(element, key, currentText, previousText);
+  }
+
+  const finalText = element.props.value ?? currentText;
+
+  if (options?.submitEditing === true) {
```
pierrezimmermannbam, line 58: if (options?.submitEditting)

src/user-event/utils/text-range.ts
```diff
@@ -0,0 +1,11 @@
+export interface TextRange {
+  start: number;
+  end: number;
+}
+
+export function getTextRange(text: string): TextRange {
+  return {
+    start: text.length,
```
pierrezimmermannbam, line 8: Why do start and end have the same value?

src/user-event/type/type.ts
```diff
@@ -1,20 +1,129 @@
...
+  }
+
+  const keys = parseKeys(text);
+
+  if (options?.skipPress !== true) {
+    await wait(this.config);
+    dispatchHostEvent(element, 'pressIn', EventBuilder.Common.touch());
+  }
+
   await wait(this.config);
```
pierrezimmermannbam, line 40: why wait here? The wait accounts for delay between user actions like typing right? So it should only occur between keys being typed no?

src/user-event/utils/__tests__/events.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  it('does NOT dispatch event to parent host component', () => {
+    const onPressParent = jest.fn();
+    const screen = render(
+      <Text onPress={onPressParent}>
+        <Text testID="text" />
+      </Text>
+    );
+
+    dispatchOwnHostEvent(screen.getByTestId('text'), 'press', TOUCH_EVENT);
```
pierrezimmermannbam: I don't find this naming very clear, I understood what it meant but it needed some reflexion. Not sure I could come up with a better one though this one is tricky. How about dispatchHostEventWithoutBubbling?

src/user-event/utils/events.ts
```diff
@@ -15,6 +16,29 @@ export function dispatchHostEvent(
...
+    return;
+  }
+
+  act(() => {
+    handler(event);
+  });
+}
+
+/**
+ * Dispatch event functio only to given element. Does not look event handler
```
pierrezimmermannbam, line 31: typo: function

src/helpers/deprecation.ts
```diff
@@ -37,7 +37,7 @@ function deprecateQuery<QueryFn extends (...args: any) => any>(
 
 const warned: { [functionName: string]: boolean } = {};
 
-// istambul ignore next: Occasionally used
+/* istanbul ignore next: occasionally used */
```
pierrezimmermannbam, line 40: what is istanbul?

mdjastrzebski: This comment disable code coverage checking of the next line. Instanbul is the code cov used by Jest.

src/user-event/utils/dispatch-event.ts
```diff
@@ -26,12 +50,28 @@ export function dispatchHostEvent(
...
+}
+
 function getEnabledEventHandler(
   element: ReactTestInstance,
   eventName: string
 ): EventHandler | null {
   const touchResponder = isTouchResponder(element) ? element : undefined;
-
   const handler = getEventHandler(element, eventName);
   if (handler && isEventEnabled(element, eventName, touchResponder)) {
```
pierrezimmermannbam, line 76: I don't think we shoud use the function isEventEnabled for userEvent. It is used for any kind of events and supports every eventName, whereas here we could have a pretty straightforward check on the editable prop. We know we want to type something and not do something else and that allows better handling for specific events, which fireEvent lacks because it's generic and I think we should use that for userEvent

src/user-event/utils/text-range.ts
mdjastrzebski: This is how the event is passed on regular typing, it means that the cursor is at that position, but no text is actually seleted.

src/user-event/type/type.ts
mdjastrzebski: In real life there is a slight delay between press in and press out events and focus is connected to press out

src/user-event/utils/__tests__/events.test.tsx
mdjastrzebski: Yeah naming here is difficult.

As far as I understand the React event processing, capture and bubbling phases, the looking up just for the first ancestor is not actually bubbling, as bubbling would be invoking given event for all ancestors listening for it. So using bubbling might be confusing in this case.

src/user-event/type/type.ts
mdjastrzebski: Done.

src/user-event/type/type.ts
mdjastrzebski: Done

src/user-event/utils/events.ts
mdjastrzebski: Done

src/user-event/utils/dispatch-event.ts
mdjastrzebski: I have to check experimentally if parent `pointerEvents` could block `TextInput` focus, if not then we could just inspect the `editable` prop.

src/user-event/type/type.ts
pierrezimmermannbam: press out is called after press ends and a minimum of 130ms have ellapsed after the start of the press, so it would make more sense I think to wait 130ms rather than waiting for the delay which accounts for user interactions

src/user-event/utils/dispatch-event.ts
pierrezimmermannbam: Oh I hadn't thought about that, I'll probably need to check for pointer events as well for Text and TextInput for user.press

src/user-event/type/type.ts
mdjastrzebski: I am wondering how realistic we want to be for events timing. There is a spectrum of possibilities:
* on the most realistic end we would try to check the timing of various events use it in respective functions
* on the other end we could follow RTL's UE approach and use delay of 0 ms by default (= next event queue iteration).

For some time-related event like `longPress` obviously we need to introduce the 500ms or so waiting, but for regular `press` and typing events we could go either of the above ways.

Pros and cons of realistic approach:
+ realistic timing
- user will need to use fake timers for all tests or have long running tests
- variable timing of different events, not consistent with RTL's UE `delay` option API
- need to conduct experiments to get actual timing

Pros and cons of lightweight approach:
+ simple implementation
+ consistent with RTL's UE `delay` option API
+ can run most of the tests without fake timers
+ simulation good enough for most cases (next event loop iteration)
- artificial (0ms or fixed) timing delay between events

If user is to take benefit of the realistic timing, e.g. for touch interactions calculation, we should probably restore non-zero `timestamp` for press events, as this might be useful in such cases.

So far the lightweight / RTL's UE approach seems to offer better tradeoffs, but I might have the whole picture of the situation.

@pierrezimmermannbam @MattAgn @AugustinLF  What are your thoughts on the event timing subject, is there any aspect I've omitted or misunderstood.

pierrezimmermannbam: For every timing that would represent the time for a user to do an action, e.g. between typing two characters or to press a button, I would agree that sticking to the RTL's approach looks better. I think that how much time is spent between two actions will very rarely matter and since it can vary a lot from a user to another the code should not to rely on that. Also the users currently still have the possibility to change that delay if they want to.

Then for some other situations, it can be interesting to have a more accurate timing. For press events on Pressables for instance, since they're not mocked by RN preset we don't really have a choice, we need to wait 500ms to trigger a long press and the pressOut only happens after a minimum 130ms after the press starts so we also have to wait for that. For press on Text and TextInput, we don't really have to wait but if the users specifies a press duration (which defaults to 0) then we should wait for that time in any case. Technically we don't have to wait before the press out because Text and Textinputs are currently mocked but since it's a RN rule that pressOut will be called after a minimum of 130ms I would still wait, even though it will make tests without fake times slower 

The event builder for press events does not use a zero timestamp and maybe we should reconsider using a zero timestamp for any of the events because the only motivation to do so if i remember correctly was to make test writing easier

mdjastrzebski: Thanks for sharing your perspective. Do you have more info on that min 130 ms wait time? Is that for RN docs, source code or experiments? I would like to get better understanding around this as it seems important in our context.

pierrezimmermannbam: This is from source code of [Pressability](https://github.dev/facebook/react-native/blob/6d24ee13a47a58ff1b19c5a51f39f52c9a4a8a28/packages/react-native/Libraries/Pressability/Pressability.js#L788-L789). I came across that when building press. I realized that onPressOut was not immediately called after calling onResponderRelease. That's why in my [implementation](https://github.com/callstack/react-native-testing-library/blob/dc7452284f352f2b366011610467d7f79d3365fa/src/user-event/press/press.ts#L104) I'm waiting for the press out when the press duration is bellow 130ms

src/user-event/utils/dispatch-event.ts
```diff
@@ -15,6 +17,30 @@ export function dispatchHostEvent(
...
   event: unknown
+) {
+  const handler = findEnabledEventHandler(element, eventName);
+  if (!handler) {
+    return;
+  }
+
+  act(() => {
+    handler(event);
+    flushMicroTasks();
```
pierrezimmermannbam, line 28: We don't call `flushMicroTasks` in `fireEvent` if I'm correct, should it be called there as well? Is it needed at all if we're using act? If so could you provide a test case that would fail if we didn't call this?

src/user-event/utils/dispatch-event.ts
mdjastrzebski: So on iOS: pointerEvents="none" on either View wrapping TextInput or TextInput itself prevents TextInput from gaining focus
On Android:  pointerEvents="none" on either View wrapping TextInput or prevents TextInput from gaining focus, but on TextInput itself is not preventing focus 🤦‍♂️

src/user-event/utils/dispatch-event.ts
mdjastrzebski: I've spotted error, as flushing microtasks requires `await`. Thanks for prompting me. 

It is not feasible to flush micro tasks in `fireEvent` as it would require signature change to `async`. BTW sync `act` does not have ability to flush micro tasks.  Beside that `fireEvent` it emits only a single event, and user has an option to flush the tasks manually if needed.

mdjastrzebski: I've done a background research on RTL's UE approach. So they use [`dispatchEvent`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent) which actually delivers event synchronously in contrast to "native" browser events which are using event loop. 

So if sync events are good enough for UE, then we should not bother too much with going async here.

pierrezimmermannbam:
This looks really good! I have one question left though: a lot of code that's in `dispatch-event` seems to be quite similar to the code of `fireEvent` (e.g. `getEventHandler`). Is it intentional to have code duplication or could those be reused in fireEvent? Also I'm still not convinced that the check on wether the event is enabled or not should be done at `dispatchOwnHostEvent ` level. What bothers me is that this is very generic so the check has to work for all kind of events but that can lead to the same issues that we had with fireEvent previously (lot of if conditions, hard to maintain, lot of bugs). Another thing is that if we call userEvent.type on a non editable TextInput, we'll call `dispatchOwnHostEvent` with the same events and wait between each events even though dispatching the events will do nothing, and I think it would be more natural to skip for instance the whole type events emitting, wdyt?

mdjastrzebski:
Good point, in the specific case of TextInput we can do single check on TextInput if it's enabled.

I wonder if we could do something similar for more general case, e.g. When doing `press` on `Text` child of 'Pressable'. In this case not only enabled check should be lifted out of dispatch but also locating the actual element to press (going from child `Text` to parent `Pressable`).

mdjastrzebski:
Rebased after merging `userEvent.press`

mdjastrzebski:
@pierrezimmermannbam could you take a look at this PR.

I've address the remaining comment about making TextInput enable check separately from dispatching the events. I've also greatly simplified the `dispatchEvent` method to make it really simple yet useful.

Finally, I've tweaked some of the `press` & `type` code to have matching code style (e.g. by using `dispatchEvent`).

website/docs/UserEvent.md
```diff
@@ -59,8 +63,62 @@ longPress(
...
+The sequence of events depends on `multiline` prop, as well as passed options.
+
+Events will not be emitted if `editable` prop is set to `false`.
+
+Entering the element:
+- `pressIn` (optional)
+- `focus`
+- `pressOut` (optional)
+
+The `pressIn` and `pressOut` events are send by default, but can be skipped by passing `skipPress: true` option.
```
pierrezimmermannbam, line 108:
```suggestion
The `pressIn` and `pressOut` events are sent by default, but can be skipped by passing `skipPress: true` option.
```

website/docs/UserEvent.md
```diff
@@ -59,8 +63,62 @@ longPress(
...
+- `selectionChange`
+
+The `textInput` event is sent only for mutliline text inputs.
+
+Leaving the element:
+- `submitEditing` (optional)
+- `endEditing`
+- `blur`
+
+The `submitEditing` event is skipped by default. It can be send by setting `submitEditing: true` option.
```
pierrezimmermannbam, line 124:
```suggestion
The `submitEditing` event is skipped by default. It can be sent by setting `submitEditing: true` option.
```

pierrezimmermannbam (approved):
This looks very good! I just caught some typos in the docs but other than that this is more than what I had expected for a v1 of userEvent, i look forward to trying it!

mdjastrzebski:
🎉 This PR has been released in v12.2.0 🚀

PR #1401, opened 2023-05-02
refactor: `fireEvent` cleanup

mdjastrzebski:
### Summary

Scope:
- [x] cleanup `fireEvent` code before further changes
  - [x] removal of unused `callsite` argument
  - [x] simplify `fireEvent` object construction by removing `invokeEvent`
- [x] fix case when passing `onPress` event name would defeat pointer events checks for `press` event despite invoking the same handler.
- [x] improve code coverage by removing dead code paths
- [x] improve code coverage by disabling checks for deprecation functions 

### Test plan

Improved existing pointer event test by using both event forms `press` and `onPress.

### DIFF
--- a/src/fireEvent.ts
+++ b/src/fireEvent.ts
@@ -93,7 +93,7 @@ const getEventHandler = (
   element: ReactTestInstance,
   eventName: string
 ): EventHandler | undefined => {
-  const eventHandlerName = toEventHandlerName(eventName);
+  const eventHandlerName = getEventHandlerName(eventName);
   if (typeof element.props[eventHandlerName] === 'function') {
     return element.props[eventHandlerName];
   }
@@ -105,19 +105,20 @@ const getEventHandler = (
   return undefined;
 };
 
-const invokeEvent = (
+const getEventHandlerName = (eventName: string) =>
+  `on${eventName.charAt(0).toUpperCase()}${eventName.slice(1)}`;
+
+const fireEvent = (
   element: ReactTestInstance,
   eventName: string,
   ...data: Array<any>
 ) => {
   const handler = findEventHandler(element, eventName);
-
   if (!handler) {
     return;
   }
 
   let returnValue;
-
   act(() => {
     returnValue = handler(...data);
   });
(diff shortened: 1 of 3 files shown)


src/helpers/component-tree.ts
```diff
@@ -4,9 +4,7 @@ import { ReactTestInstance } from 'react-test-renderer';
  * Checks if the given element is a host element.
  * @param element The element to check.
  */
-export function isHostElement(
-  element?: ReactTestInstance | null
-): element is ReactTestInstance {
+export function isHostElement(element?: ReactTestInstance | null) {
```
MattAgn: you don't need the the typeguard anymore?

MattAgn (approved):

MattAgn:
That's some nice cleaning! 🔥

src/fireEvent.ts
```diff
@@ -20,18 +16,20 @@ const isTextInput = (element?: ReactTestInstance) => {
...
-const isTouchResponder = (element?: ReactTestInstance) => {
-  if (!isHostElement(element)) return false;
+function isTouchResponder(element: ReactTestInstance) {
+  if (!isHostElement(element)) {
+    return false;
+  }
 
-  return !!element?.props.onStartShouldSetResponder || isTextInput(element);
-};
+  return !!element.props.onStartShouldSetResponder || isTextInput(element);
```
MattAgn, line 26: [nitpick] what do you think of using Boolean() instead of `!!` ? I find it better represents what we're doing most of the time and it's easier to read and understand

src/helpers/component-tree.ts
mdjastrzebski: I've added the type guard temporary in this PR but it caused more trouble than benefit, as in negative case TS assumed that element is `null`, which is not the case.

src/fireEvent.ts
mdjastrzebski: Tbh I personally prefer `!!` because it's concise, but if you find `Boolean()` more readable we can go this way.

src/helpers/component-tree.ts
MattAgn: oh ok i understand

src/fireEvent.ts
MattAgn: both are fine for me ;) (it's just a small personal preference, also because i still find `!!` a bit hacky ^^)

MattAgn (approved):

src/fireEvent.ts
mdjastrzebski: With JS you get use to some hacking techniques over time that they feel normal ;-)

MattAgn: haha so true

MattAgn (approved):

AugustinLF (approved):

mdjastrzebski:

PR #1403, opened 2023-05-02
refactor: component tree dead code

mdjastrzebski:
### Summary

Scope:
- [x] Migrate away for composite based checks in `component-tree.ts`
- [x] Add `Switch` to `detectHostComponentNames` and React Native API assumptions tests
- [x] Remove unused `component-tree.ts` functions: `getHostSelf`, `getCompositeParentOfType`, `isHostElementForType`.

### Test plan

N/A

### DIFF
--- a/src/__tests__/react-native-api.test.tsx
+++ b/src/__tests__/react-native-api.test.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { View, Text, TextInput } from 'react-native';
+import { View, Text, TextInput, Switch } from 'react-native';
 import { render } from '..';
 
 /**
@@ -19,7 +19,6 @@ test('React Native API assumption: <View> renders single host element', () => {
 
 test('React Native API assumption: <Text> renders single host element', () => {
   const view = render(<Text testID="test">Hello</Text>);
-  expect(view.getByText('Hello')).toBe(view.getByTestId('test'));
 
   expect(view.toJSON()).toMatchInlineSnapshot(`
     <Text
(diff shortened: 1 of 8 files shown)


src/__tests__/react-native-api.test.tsx
```diff
@@ -91,3 +81,27 @@ test('React Native API assumption: <TextInput> renders single host element', ()
     />
   `);
 });
+
+test('React Native API assumption: <Switch> renders single host element', () => {
+  const view = render(
+    <Switch testID="test" value={true} onChange={jest.fn()} />
+  );
+  expect(view.getByTestId('test')).toBe(view.getByTestId('test'));
```
MattAgn, line 89: why did you remove this expectation from the other tests for view and text but not for switch? (i don't really understand the value of this assertion)

MattAgn:
Nice cleaning! Are you on a refactoring streak? 😄

MattAgn (approved):

mdjastrzebski:
@MattAgn I've managed to find some time and energy to work on User Event `type()` and I've noticed some code in need for cleanup 🧹🧹🧹

mdjastrzebski:

BillyBlaze:
@mdjastrzebski thanks for the work.

I'm looking into our dependency update and why the tests are failing from `12.1.1` to `12.1.2`. It seems that the new Switch component is generating an useState hook and that is messing up tests that spies on the useState. Is this by design? and if yes, is this not a breaking change since everyone that spies on the useState suddenly gets more calls and receives a value `{ value: null }` instead of what their useState value should be. It took me some time to figure out that the Switch component was emitting this value and that this was generated in `host-component-names`.

mdjastrzebski:
@BillyBlaze thank you for reaching out with your issue. The root cause why RNTL internal code change caused test failures for you is that your test relied on spying on global React primitives, like `useState`. This way any other component change, in RNTL, React Native or even 3rd party library, could potentially cause test failures. 

We do not consider this to be a breaking change, as our public API remained the same and we do not recommend such testing patterns as they result in fragile testing code that is dependent on tested components internal details rather than public API.

Per our guiding principle:

> The more your tests resemble the way your software is used, the more confidence they can give you.

The recommended way is to use the props and event handlers (public API) of your components instead of internal implementation details like internal state, etc.

I hope that clarifies our stance in this regards and allows you to write more resilient tests.

PR #1404, opened 2023-05-04
fix:  support `onXxx` event name for TextInput event checks

mdjastrzebski:
### Summary

This fixed an edge case when user would use `onLayout` event handler name instead of `layout` event name with `fireEvent` and that would skip the `editable` prop check.

Additionally, refactored event array checks to use `Set` type.

### Test plan

Added test code to detect the fixed scenario.


thymikee (approved):

mdjastrzebski:

PR #1405, opened 2023-05-06
feat: User Event core code

mdjastrzebski:
### Summary

This PR contain core User Event code to be used by both `type()` and `press()` functions. The API is based on latest User Event v14 API. 

This PR contains stub implementations for `type()` and `press()`, as well as some re-usable utils:
- `wait()` based on User Event's implementation
- `createEventLogger` testing util for inspecting the sequence of events.

Some of this stuff is quite opinionated, so if you have ideas of alternative approaches, e.g. to testing, pls voice them so that we have a commonly agreed base for User Event code.

### Test plan

Some new tests added, before extracting from #1396 had 100% code coverage.


src/user-event/utils/wait.ts
```diff
@@ -0,0 +1,15 @@
+import { UserEventConfig } from '../setup';
+
+export function wait(config: UserEventConfig) {
+  const delay = config.delay;
+  if (typeof delay !== 'number') {
+    return;
+  }
+
+  return Promise.all([
```
pierrezimmermannbam, line 9: This is very confusing to me, I don't understand really why the advanceTimers function is used only for fake timers. Here we wait for two promises, one being useful in case fake timers are used and the other in case real timers are used, is this because we don't want to use `jestFakeTimersAreEnabled `?

Also I don't really see the point of having  `advanceTimers` as an option, what could it be used for? We don't really support any other test runner aside from jest currently right? I don't really know what users are using as a test runner but I have the impression that in RN with all the required presets Jest is currently the only solution that's supported by most major libraries

website/docs/UserEvent.md
```diff
@@ -0,0 +1,30 @@
...
+```ts
+userEvent.setup(options?: {
+  delay: number;
+  advanceTimers: (delay: number) => Promise<void> | void;
+})
+```
+
+Example
+```ts
+const user = userEvent.setup();
```
pierrezimmermannbam, line 23: I'm not a huge fan of having to call the setup function to be able to trigger events. Why not have a config like the one we already have ? User event config could even be added to the existing global config maybe. I think having to call the setup function in each test makes the API not as nice to use and I'd prefer to have a global config and be able then to customize options at events level, very much like what is currently available for queries

src/user-event/event-builder/common.ts
```diff
@@ -0,0 +1,48 @@
+export const CommonEventBuilder = {
```
pierrezimmermannbam, line 1: Very nice! One thing I'm not sure to get however is what would be considered common and what would not be?

src/user-event/press/__tests__/press.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { userEvent } from '../..';
+
+beforeEach(() => {
+  jest.resetAllMocks();
+});
+
+describe('user.press()', () => {
+  it('dispatches required events on Text', async () => {
+    // Required for touch events which contain timestamp
+    jest.spyOn(Date, 'now').mockReturnValue(100100100100);
```
pierrezimmermannbam, line 14: Do we really want the events builded to use Date.now? Having to mock it in tests, especially when not using real timers does not seem ideal. Is it because they're also used in userEvent implementation? If that's the case I think default events used for the implementation and mocks should be two distinct things, although they could share the same interface

website/docs/UserEvent.md
mdjastrzebski: This is the current [User Event](https://testing-library.com/docs/user-event/intro/) API, v14. In the previous version, v13 you didn't have to create that `user` object.

Although I would prefer to have simpler API (`userEvent.press()`), imo there is high value in matching the current RLT's API, as we try to do for all possible RTL API, which means using `userEvent.setup()`.

src/user-event/event-builder/common.ts
mdjastrzebski: The others I see for now would be [TextInput specific ones](https://github.com/callstack/react-native-testing-library/pull/1396/files#diff-2a2277746640ea803ab6aee547f4e8e3c5050f82a860946497a768c8ee1c3293), which there are quite a few.

src/user-event/press/__tests__/press.test.tsx
mdjastrzebski: We could go with zero here, will simplify the code indeed. 

The reason I've used `Date.now()` was that of all the values we have in that event, this is the single one where would code provide some meaningful value ;-)

src/user-event/utils/wait.ts
mdjastrzebski: This implementation is taken straight from actual User Event code: https://github.com/testing-library/user-event/blob/main/src/utils/misc/wait.ts

The way it works is that the first part: `new Promise<void>(resolve => globalThis.setTimeout(() => resolve(), delay)),` works for both real and fake timers.

The second part `config.advanceTimers(delay)` is configurable through `userEvent.setup()` call. By default UserEvent has `() => Promise.resolve()` there, which allows this function to work with real timers. Having it as an option allows to support different fake timer types, e.g. [Sinon.JS](https://sinonjs.org/releases/latest/fake-timers/) and not just Jest ones.

My innovation in that regard is that I've created a [universal `advanceTimers` default](https://github.com/callstack/react-native-testing-library/pull/1405/files#diff-64635189362ce288644481b4ecbcfd937bd623c96831718a6d77e04ef460e591) function that works with both real timers and fake Jest timers, which I assume are used by majority of our users using fake timers at all.

Having that as an option allows other fake timers users, still to use them with our User Event. And there have been [tickets](https://github.com/callstack/react-native-testing-library/pull/1169) from users using Sinon.JS as more powerful timers implementation.

src/user-event/press/__tests__/press.test.tsx
mdjastrzebski: Changed to use `timestamp: 0`.

website/docs/UserEvent.md
pierrezimmermannbam: Oh it seems to be used also to carry state between events, now I understand better why they chose this API. It doesn't look very useful in our case for now but if they opted for this on react side it may be the best long term option here as well. Also I agree it is good to have the same API so I'm onboard with this

src/user-event/event-builder/common.ts
pierrezimmermannbam: Ok I see, seems good

src/user-event/utils/wait.ts
pierrezimmermannbam: Ok this seems reasonable if we want to support other fake timers than the ones from Jest. I'm curious though is there a way in user-event to configure the advanceTimers option globally or does it need to be done every time in the setup function?

pierrezimmermannbam (approved):

src/user-event/utils/wait.ts
mdjastrzebski: For global, RTL recommends putting User Event setup into custom render function:

```
// setup function
function setup(jsx) {
  return {
    user: userEvent.setup(),
    ...render(jsx),
  }
}

test('render with a setup function', async () => {
  const {user} = setup(<MyComponent />)
  // ...
})
```

Source: https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent

mdjastrzebski:
@pierrezimmermannbam I've discovered that RTL offers direct access to User Event functions (e.g. `userEvent.press()`). This is optional API:
> Note that, while directly invoking APIs such as userEvent.click() (which will trigger setup internally) is [still supported in v14](https://testing-library.com/docs/user-event/setup#direct-apis), this option exists to ease the migration from v13 to v14, and for simple tests. We recommend using the methods on the instances returned by userEvent.setup().
[Source](https://testing-library.com/docs/user-event/intro#writing-tests-with-userevent)

I've added that to this PR.

mdjastrzebski:
Tbh this seems relatively minor. If TL/UE decides to remove this API that would be a breaking change, which would force a major release. 

Triggering a major release just for removing this API seems unlikely, more probably they would introduce some other breaking changes as well. That would probably require us to also increase the major version in order to remain compatible, which we could use to also remove the same API, in addition to bringing in other changes.

pierrezimmermannbam:
Ok yeah that makes sense, all good for me then!

mdjastrzebski:
It is not added to package-level exports yet, waiting for `userEvent.type()` PR.

PR #1415, opened 2023-05-31
chore: switch Codecov PR comment to default behavior

thymikee:
### Summary

It's super annoying to get dozens of meaningless codecov notifications in my inbox every day, because of the current behavior. Switching to default to "update, if exists. Otherwise post new"

Ref: https://docs.codecov.com/docs/pull-request-comments#behavior

### Test plan

Codecov updates instead of recreating the comment

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,6 +4,6 @@ coverage:
   range: 70...100
 
 comment:
-  behavior: new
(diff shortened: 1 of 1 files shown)


thymikee:
No extra emails and the comment is now edited. Perfect!

image

PR #1416, opened 2023-05-31
chore: experiments app tweaks

mdjastrzebski:
### Summary

Scope:
* add logging of micro and macro tasks enqueued during event processing.
* update Expo deps for experiments app

Experiments results are described [here](https://github.com/callstack/react-native-testing-library/wiki/TextInput-Events#micro--macro-tasks-after-events). TLDR: we should flush micro tasks after each emitted event. Macro task behavior vary, but for simplicity we should also flush these as well.

### Test plan

None

### DIFF
--- a/experiments-app/package.json
+++ b/experiments-app/package.json
@@ -18,7 +18,7 @@
     "expo": "~48.0.11",
     "expo-status-bar": "~1.4.4",
     "react": "18.2.0",
-    "react-native": "0.71.6",
+    "react-native": "0.71.8",
     "react-native-safe-area-context": "4.5.0",
     "react-native-screens": "~3.20.0",
     "yarn": "^1.22.19"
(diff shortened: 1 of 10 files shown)


experiments-app/package.json
```diff
@@ -18,14 +18,15 @@
...
-    "react-native": "0.71.6",
+    "react-native": "0.71.8",
     "react-native-safe-area-context": "4.5.0",
     "react-native-screens": "~3.20.0",
     "yarn": "^1.22.19"
   },
   "devDependencies": {
     "@babel/core": "^7.20.0",
     "@types/react": "~18.0.14",
+    "@types/react-native": "0.71",
```
pierrezimmermannbam, line 29: Is this needed? Types are included in `react-native` for versions >= 71 no?

PR #1421, opened 2023-06-22
fix: correct type inference for props when initialProps is used

### LINKED ISSUE #1420: renderHook's rerender not has undefined prop type unless explicitly typed
## Describe the bug

The following code returns a typescript error

```ts
import { renderHook } from '@testing-library/react-native'
import { useMyHook } from '../hooks'

describe('useMyHook', () => {
  it('should pass', () => {
    const { result, rerender } = renderHook((num: number) => useMyHook(num), { initialProps: undefined })

    expect(result.current).toBeUndefined()

    rerender('granted') //Error here, props type is undefined

    expect(result.current).toBe('granted')
  })
})
```

For it to work fine I have to to do
```ts
    const { result, rerender } = renderHook<number, number>((num: number) => useMyHook(num), { initialProps: undefined })
```

this doesn't happen when using 
```ts
import { renderHook } from '@testing-library/react-hooks'
```

## Expected behavior

Not having a typescript error and instead pull the prop types correctly.

## Steps to Reproduce

See description.

## Screenshots

Screenshot 2023-06-20 at 12 58 02

Screenshot 2023-06-20 at 12 58 17

## Versions

```
  npmPackages:
    @testing-library/react-native: 12.1.2 => 12.1.2 
    react: 18.1.0 => 18.1.0 
    react-native: 0.70.9 => 0.70.9 
    react-test-renderer: 18.1.0 => 18.1.0 
```

pierrezimmermannbam:
### Summary

Fixes #1420

Because we were using conditionals in RenderHookOptions and the way Typescript works with conditionals, Props type was inferred too strictly. For instance when doing:

```tsx
 renderHook(
    (num: number | undefined) => useMyHook(num),
    {
      initialProps: 5,
    }
  );
```
Props type would be infered as 5 and not number | undefined which would be the expected type. 

This PR fixes this issue by removing the use of the conditional. This doesn't change much the behavior and it aligns with the types from [testing library react hooks](https://github.com/testing-library/react-hooks-testing-library/blob/main/src/types/index.ts#LL73C3-L73C3) and [react testing library](https://github.dev/testing-library/react-testing-library/blob/6b4180e71286cef86a359435697965e59d408d91/types/index.d.ts#L133)

### Test plan

I added a test case to check that the type is inferred correctly

### DIFF
--- a/src/renderHook.tsx
+++ b/src/renderHook.tsx
@@ -8,12 +8,10 @@ export type RenderHookResult<Result, Props> = {
   unmount: () => void;
 };
 
-export type RenderHookOptions<Props> =
-  | {
-      initialProps?: Props;
-      wrapper?: ComponentType<any>;
-    }
-  | undefined;
+export type RenderHookOptions<Props> = {
+  initialProps?: Props;
+  wrapper?: ComponentType<any>;
+};
 
 export function renderHook<Result, Props>(
   renderCallback: (props: Props) => Result,


src/__tests__/renderHook.test.tsx
```diff
@@ -60,3 +60,37 @@ test('allows wrapper components', async () => {
...
+  );
+
+  expect(result.current).toBe(5);
+
+  rerender(6);
+
+  expect(result.current).toBe(6);
+});
+
+test('props type is infered correctly when initial props is explicitly undefined', () => {
```
mdjastrzebski, line 83:
```suggestion
test('props type is inferred correctly when initial props is explicitly undefined', () => {
```

src/renderHook.tsx
```diff
@@ -8,16 +8,12 @@ export type RenderHookResult<Result, Props> = {
...
-  | boolean
-  ? {
-      initialProps: Props;
+export type RenderHookOptions<Props> =
+  | {
+      initialProps?: Props;
       wrapper?: ComponentType<any>;
     }
-  : { wrapper?: ComponentType<any>; initialProps?: never } | undefined;
+  | undefined;
```
mdjastrzebski, line 16: [mid] We should be able to remove `| undefined` as `options?: RenderHookOptions<Props>` should handle it.

src/renderHook.tsx
```diff
@@ -8,16 +8,12 @@ export type RenderHookResult<Result, Props> = {
   unmount: () => void;
```
mdjastrzebski, line 8: [nit] Comparing to RTL typing, their `rerender` `Props` argument is optional: https://github.com/testing-library/react-testing-library/blob/6b4180e71286cef86a359435697965e59d408d91/types/index.d.ts#L105

Not sure if we should follow that as well.

src/renderHook.tsx
pierrezimmermannbam: Indeed, nice catch!

src/renderHook.tsx
pierrezimmermannbam: I don't really understand how it work on their side but what happens when we call rerender without props is that it will render the component with undefined as props and I think the expected outcome would be to rerender with the previous props. We could implement it but I'm not sure I see a use case for it

mdjastrzebski (approved):
Looks good, let's merge!

mdjastrzebski:

Issue #1423, opened 2023-06-30
First test running render/renderHook is slow


KrastanD:
## Describe the bug

  The first test in a file that uses render/renderHook is slow. This is exacerbated when ran without jest cache. 

  Locally on my M1 Pro Macbook:
  Without jest cache, the first test takes ~27 times as long as the second test. 1512ms vs 55ms
  With jest cache, the first test takes ~3 times as long which seems reasonable. 142ms vs 56ms

  On Github Actions:
   Without jest cache, the first test takes ~70 times as long as the second test. 
  
  I know it's not related to the actual tests because if I change the order of them it is always the one that is first that takes significantly longer.

## Expected behavior

I would expect that it takes the tests to run under a second whether they are cached or not, first or not.

## Steps to Reproduce

React Query has an example repo using CRA and react testing library with some sample tests [here](https://github.com/TkDodo/testing-react-query/tree/main)
I recreated the repo and tests using expo and react native testing library [here](https://github.com/KrastanD/testing-react-native-query)

When running the RTL repo tests without cache I get the following: 
![image](https://github.com/callstack/react-native-testing-library/assets/32044675/a19b5b94-1d7d-482a-af74-a36d3fb6878f)
All tests take under 100 ms. 

When running the RNTL repo tests without cache I get the following:
![image](https://github.com/callstack/react-native-testing-library/assets/32044675/4498d36f-ea31-4280-a1f0-4b7439a045d8)
Both first tests take 1.6 seconds.

I also ran the RNTL repo tests on Github Actions [here](https://github.com/KrastanD/testing-react-native-query/actions/runs/5424525570/jobs/9864005024) with the following results:
![image](https://github.com/callstack/react-native-testing-library/assets/32044675/89b28cd6-5e30-4eae-99c8-4313dcea92e2)
Interestingly its only the first renderHook that takes a long time of 4.1 seconds. 

I created a branch in the RNTL repo called [renderHook-isolated](https://github.com/KrastanD/testing-react-native-query/tree/renderHook-isolated). There I created two new test files that don't use react-query or msw - just basic react-native and react native testing library. 
Here are the results:
![image](https://github.com/callstack/react-native-testing-library/assets/32044675/5ebe67a5-10cb-449f-adb7-b2376687b426)
Even the very basic react hook test takes 3/4 of a second and it seems react query just makes the matter worse. 

[...]

PR #1425, opened 2023-07-06
fix: add prop to render to skip configureHostComponentNamesIfNeeded

KrastanD:
### Summary

https://github.com/callstack/react-native-testing-library/issues/1423

### Test plan

This change can be tested by running the renderHook.test.tsx tests after clearing the jest cache. Before my change, the first test takes 445 ms and after my change it takes 5 ms. 

Before: 
``` ~/Code/react-native-testing-library   main  yarn jest ./src/__tests__/renderHook.test.tsx
yarn run v1.22.19
$ ~/Code/react-native-testing-library/node_modules/.bin/jest ./src/__tests__/renderHook.test.tsx
 PASS  src/__tests__/renderHook.test.tsx
  ✓ gives comitted result (445 ms)
  ✓ allows rerendering (1 ms)
  ✓ allows wrapper components (2 ms)

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        1.273 s
Ran all test suites matching /.\/src\/__tests__\/renderHook.test.tsx/i.
✨  Done in 2.19s.
```

After: 
```~/Code/react-native-testing-library   fix/renderHook-skip-component-names-config  yarn jest ./src/__tests__/renderHook.test.tsx 
yarn run v1.22.19
$ ~/Code/react-native-testing-library/node_modules/.bin/jest ./src/__tests__/renderHook.test.tsx
 PASS  src/__tests__/renderHook.test.tsx
  ✓ gives comitted result (5 ms)
  ✓ allows rerendering (1 ms)
  ✓ allows wrapper components

Test Suites: 1 passed, 1 total
Tests:       3 passed, 3 total
Snapshots:   0 total
Time:        0.819 s
Ran all test suites matching /.\/src\/__tests__\/renderHook.test.tsx/i.
✨  Done in 1.74s.
```

### DIFF
--- a/src/renderHook.tsx
+++ b/src/renderHook.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import type { ComponentType } from 'react';
-import render from './render';
+import { renderInternal } from './render';
 
 export type RenderHookResult<Result, Props> = {
   rerender: (props: Props) => void;
@@ -36,10 +36,13 @@ export function renderHook<Result, Props>(
     return null;
   }
 
-  const { rerender: baseRerender, unmount } = render(
+  const { rerender: baseRerender, unmount } = renderInternal(
     // @ts-expect-error since option can be undefined, initialProps can be undefined when it should'nt
     <TestComponent renderCallbackProps={initialProps} />,
-    { wrapper, skipHostComponentNamesConfiguration: true }
+    {
+      wrapper,
+      detectHostComponentNames: false,
+    }
   );
 
   function rerender(rerenderCallbackProps: Props) {
--- a/src/renderHook.tsx
+++ b/src/renderHook.tsx
@@ -39,7 +39,7 @@ export function renderHook<Result, Props>(
   const { rerender: baseRerender, unmount } = render(
     // @ts-expect-error since option can be undefined, initialProps can be undefined when it should'nt
     <TestComponent renderCallbackProps={initialProps} />,
-    { wrapper }
+    { wrapper, skipHostComponentNamesConfiguration: true }
   );
 
   function rerender(rerenderCallbackProps: Props) {
--- a/src/__tests__/render.test.tsx
+++ b/src/__tests__/render.test.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable no-console */
 import * as React from 'react';
 import { View, Text, TextInput, Pressable } from 'react-native';
+import { getConfig, resetToDefaults } from '../config';
 import { render, screen, fireEvent, RenderAPI } from '..';
 
 const PLACEHOLDER_FRESHNESS = 'Add custom freshness';
@@ -242,6 +243,14 @@ test('RenderAPI type', () => {
 test('returned output can be spread using rest operator', () => {
   // Next line should not throw
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  const { rerender, ...rest } = render(<View testID="inner" />);
+  const { rerender, ...rest } = render(<View testID="test" />);
   expect(rest).toBeTruthy();
+});
+
+test('render calls detects host component names', () => {
+  resetToDefaults();
+  expect(getConfig().hostComponentNames).toBeUndefined();
+
+  render(<View testID="test" />);
+  expect(getConfig().hostComponentNames).not.toBeUndefined();
 });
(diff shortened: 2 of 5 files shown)


pierrezimmermannbam:
Thanks for opening this pr @KrastanD!  Regarding the implementation, I don't think the API of `render` should change because it won't work if the host names detection is not run. Also it never needs to be executed in `renderHook` so it does not look like we need this to be configurable since in one case we always need it and in the other one we never do. Could you update your pr to keep the public API and just make it so that `renderHook` doesn't run name detection?

This would be a good step forward but it would only make `renderHook` faster though so it doesn't fully solve the problem. I wonder if we should allow host component names to be configured by the user. It would give more control to users. On the other hand I'm not sure it should be part of the public API. I'd probably use it but it also makes the API more complex so I'm not sure it's worth it. Wdyt @mdjastrzebski @MattAgn  @AugustinLF ?

KrastanD:
Hopefully this update is more of what you had in mind @pierrezimmermannbam. I pulled out only the necessary parts of the render function into the renderHook. 

I didn't want to change the api in any way so there is a bit of code duplication between `updateWithAct` in the render function and `baseRerender` in renderHook. With your blessing, I can pull out `updateWithAct` into its own file like `renderWithAct` is and reuse it in renderHook.

src/renderHook.tsx
```diff
@@ -42,12 +43,20 @@ export function renderHook<Result, Props>(
     return null;
   }
 
-  const { rerender: baseRerender, unmount } = render(
+  const wrap = (element: React.ReactElement) =>
```
mdjastrzebski, line 46: What is the purpose of this change? It seems to do the same thing as previous code with just passed wrapper to `renderWithAct`

mdjastrzebski:
@KrastanD I support resolving the issue of `renderHook` unnecessarily calling `detectHostComponents`. I've encountered another use-case when it would be beneficial: in case user mocks `react-native` package using `jest.mock()` and then tests hook, the `renderHook` fails as `Text`/`TestInput` is not longer available (user just mocked whole `react-native` package), but the hook itself does not need it at all. 

Current implementation feels unnecessarily complex though. I would opt for "internal" (non-documented) `render` flag that would control `detectHostComponents` method being called. By default it would be called (when user just invokes `render`), but when `renderHook` would call `render` it would use it to ignore that step.

Proposed naming:
* `render` option name: `detectHostComponentNames`

@pierrezimmermannbam, @KrastanD wdyt?

KrastanD:
@mdjastrzebski My initial version of this PR which you can see [here](https://github.com/callstack/react-native-testing-library/pull/1425/commits/ccd01ffc616d580fb35eded1c29bff70eaa59e8a) solved the problem in a similar way to what you are suggesting except instead of naming it `detectHostComponentNames` i named it `skipHostComponentNamesConfiguration` because by default it would be set to false. In @pierrezimmermannbam's comment above, he says the API of `render` shouldn't change which is why I rewrote it to use the internals of `render`, like `renderWithAct`, but without the hostComponentNames detection.

src/renderHook.tsx
KrastanD: `renderWithAct` does what `render` does but without detecting hostComponentNames

pierrezimmermannbam:
@mdjastrzebski  @KrastanD I personally like this implementation, I don't think it's complex, in fact I think it's an improvement. Sure there is a bit of code duplication to handle wrapper option and update with act but it doesn't add an additional prop to the render function just for renderHook's sake. This is a better separation of concern imo, if tomorrow we're to add a new feature that's only for components we'll need to add one more flag. And with this implementation it's very easy to change renderHook or render without changing the behavior of the other one, which is what we want I believe. I think this implementation will be easier to maintain so I would keep it. Also even of the flag is not documented it would be visible to users with autocomplete, it's not a big problem but still a downside

mdjastrzebski (approved):

mdjastrzebski:
@KrastanD, @pierrezimmermannbam I've refactored the code so that it uses internal option `detectHostComponentNames` which is available internally on `renderInternal` function but not exposed publicly. I've also added tests that ensure proper behavior in that regard for both `render` and `renderHook` public methods.

mdjastrzebski:

PR #1426, opened 2023-07-10
refactor: type all queries to return host components

pierrezimmermannbam:
### Summary

- Updates the return type of all queries to `HostReactTestInstance`
- Moves logic of filtering non host elements in `findAllInternal` rather than on each query level

Now that all queries return host elements, this can be made explicit by a more precise typing. It also ensures that queries will consistently return host elements from now on. I also moved the logic of filtering non host elements in `findAllInternal` rather than on each query level to remove some code duplication and most importantly remove checks that didn't use `isHostElement` and therefore were not explicit.

### Test plan

All existing tests and type checks pass

### DIFF
--- a/src/helpers/component-tree.ts
+++ b/src/helpers/component-tree.ts
@@ -1,10 +1,14 @@
 import { ReactTestInstance } from 'react-test-renderer';
 
+export type HostReactTestInstance = ReactTestInstance & { type: string };
+
 /**
  * Checks if the given element is a host element.
  * @param element The element to check.
  */
-export function isHostElement(element?: ReactTestInstance | null) {
+export function isHostElement(
+  element?: ReactTestInstance | null
+): element is HostReactTestInstance {
   return typeof element?.type === 'string';
 }
 
--- a/src/helpers/component-tree.ts
+++ b/src/helpers/component-tree.ts
@@ -1,5 +1,8 @@
 import { ReactTestInstance } from 'react-test-renderer';
 
+/**
+ * ReactTestInstance referring to host element.
+ */
 export type HostReactTestInstance = ReactTestInstance & { type: string };
 
 /**
@@ -18,7 +21,7 @@ export function isHostElement(
  */
 export function getHostParent(
   element: ReactTestInstance | null
-): ReactTestInstance | null {
+): HostReactTestInstance | null {
   if (element == null) {
     return null;
   }
@@ -41,12 +44,12 @@ export function getHostParent(
  */
 export function getHostChildren(
   element: ReactTestInstance | null
-): ReactTestInstance[] {
+): HostReactTestInstance[] {
   if (element == null) {
     return [];
   }
 
-  const hostChildren: ReactTestInstance[] = [];
+  const hostChildren: HostReactTestInstance[] = [];
 
   element.children.forEach((child) => {
     if (typeof child !== 'object') {
@@ -72,10 +75,8 @@ export function getHostChildren(
  */
 export function getHostSelves(
   element: ReactTestInstance | null
-): ReactTestInstance[] {
-  return typeof element?.type === 'string'
-    ? [element]
-    : getHostChildren(element);
+): HostReactTestInstance[] {
+  return isHostElement(element) ? [element] : getHostChildren(element);
 }
 
 /**
@@ -84,7 +85,7 @@ export function getHostSelves(
  */
 export function getHostSiblings(
   element: ReactTestInstance | null
-): ReactTestInstance[] {
+): HostReactTestInstance[] {
   const hostParent = getHostParent(element);
   const hostSelves = getHostSelves(element);
   return getHostChildren(hostParent).filter(
--- a/src/helpers/component-tree.ts
+++ b/src/helpers/component-tree.ts
@@ -3,7 +3,7 @@ import { ReactTestInstance } from 'react-test-renderer';
 /**
  * ReactTestInstance referring to host element.
  */
-export type HostReactTestInstance = ReactTestInstance & { type: string };
+export type HostTestInstance = ReactTestInstance & { type: string };
 
 /**
  * Checks if the given element is a host element.
@@ -11,7 +11,7 @@ export type HostReactTestInstance = ReactTestInstance & { type: string };
  */
 export function isHostElement(
   element?: ReactTestInstance | null
-): element is HostReactTestInstance {
+): element is HostTestInstance {
   return typeof element?.type === 'string';
 }
 
@@ -21,7 +21,7 @@ export function isHostElement(
  */
 export function getHostParent(
   element: ReactTestInstance | null
-): HostReactTestInstance | null {
+): HostTestInstance | null {
   if (element == null) {
     return null;
   }
@@ -44,12 +44,12 @@ export function getHostParent(
  */
 export function getHostChildren(
   element: ReactTestInstance | null
-): HostReactTestInstance[] {
+): HostTestInstance[] {
   if (element == null) {
     return [];
   }
 
-  const hostChildren: HostReactTestInstance[] = [];
+  const hostChildren: HostTestInstance[] = [];
 
   element.children.forEach((child) => {
     if (typeof child !== 'object') {
@@ -75,7 +75,7 @@ export function getHostChildren(
  */
 export function getHostSelves(
   element: ReactTestInstance | null
-): HostReactTestInstance[] {
+): HostTestInstance[] {
   return isHostElement(element) ? [element] : getHostChildren(element);
 }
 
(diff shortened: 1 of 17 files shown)


src/helpers/component-tree.ts
```diff
@@ -1,10 +1,14 @@
 import { ReactTestInstance } from 'react-test-renderer';
 
+export type HostReactTestInstance = ReactTestInstance & { type: string };
```
mdjastrzebski, line 3: Q: Is that actually working? I remember trying to do something similar in the past, but there were some TS issues that made me abandon such approach. Unfortunately I do not remember these issues, they might also be related to older TS or React/RN types versions.

src/helpers/findAll.ts
```diff
@@ -17,7 +19,7 @@ export function findAll(
   root: ReactTestInstance,
   predicate: (element: ReactTestInstance) => boolean,
   options?: FindAllOptions
-) {
+): Array<HostReactTestInstance> {
```
mdjastrzebski, line 22: [nit] `HostReactTestInstance[]`

src/helpers/findAll.ts
```diff
@@ -41,11 +43,11 @@ function findAllInternal(
   root: ReactTestInstance,
   predicate: (element: ReactTestInstance) => boolean,
   options?: FindAllOptions
-): Array<ReactTestInstance> {
-  const results: ReactTestInstance[] = [];
+): Array<HostReactTestInstance> {
```
mdjastrzebski, line 46: [nit]
```suggestion
): HostReactTestInstance[] {
```

src/helpers/findAll.ts
```diff
@@ -54,9 +56,10 @@ function findAllInternal(
   });
 
   if (
+    isHostElement(root) &&
+    predicate(root) &&
```
mdjastrzebski, line 60: [nit] Could we put predicates from cheapest to (potentially) most expensive?

1. deepestOnly
2. host element
3. passed `predicate`

src/helpers/component-tree.ts
mdjastrzebski, line 3: We probably need to export this type as well in `pure.js`

mdjastrzebski:
@pierrezimmermannbam I've got back from my summer break. Great to see your` PRs!

Regarding this particular PR, the idea of using host check in `findAll` definitely makes sense. There are still two queries that can return composite elements (`UNSAFE_*`) but they do not use the same `findAll` implementation.

I am a bit concerned with the `HostTestInstance` type, as I recall that in the past I've tried to do something similar but had encountered some TS errors, as React typed `type` prop with HTML string literal union type (`'a' | 'div' | ...`), etc with somehow clashed with some other types. Unfortunately I do not remember the details. Perhaps TS and/or React types have improved. I will try to do some experiments on your PR to see if I can replicate the issues.

src/helpers/component-tree.ts
pierrezimmermannbam: What do you mean by is it working? Elements of type HostReactTestInstance have a type property of type string so I'd say it's working but maybe you're thinking of something else? 

It should work fine with `jest-native` yes, matchers expect a ReactTestInstance and a HostReactTestInstance is still of type ReactTestInstance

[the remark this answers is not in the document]
pierrezimmermannbam: I think it's important to keep the information that the type `HostReactTestInstance` comes from `ReactTestInstance` and that functions that expect a `ReactTestInstance` will work with a `HostReactTestInstance`. A solution would be to alias `ReactTestInstance`to `TestInstance` to keep this semantic, but i don't know if it makes sense if we're not adapting `react-test-renderer`'s type at all

src/helpers/findAll.ts
pierrezimmermannbam: I personally prefer the Array syntax because the the first thing you see is that it's an array, whereas I didn't see the [] at the end multiple times in the past. Also we say "an array of number" so it feels more natural to me. That being said it's better to have consistency across the codebase so I'll make the change

src/helpers/component-tree.ts
pierrezimmermannbam: Oh right, good catch!

src/helpers/findAll.ts
mdjastrzebski: As I wrote, this is nit, so let's keep it as is.

src/helpers/component-tree.ts
mdjastrzebski: iirc the issue was that `{ type: string }` could not be properly cast onto `ReactTestInstance` which expected a narrower type of `{ type: 'a' | 'span' | ... }`. Unfortunately I have not yet been able to test drive this PR.

mdjastrzebski:
@pierrezimmermannbam I've done some experiments on this PR and here are some interesting findings:

1. Your PR seems to work fine in the current scope.
2. `ReactTestInstance['type']` is indeed a string literal union of HTML tag names + class and function components:
image

So React incorrectly assumes that host components would be only HTML tags. Ignoring the React Native situation.

3. That means that `HostReactTestInstance` type is actually intersection of `string` and above string literal union of HTML tags:
image

With TS errors like:
> Argument of type '"Image"' is not assignable to parameter of type 'ElementType & string'.

Now part for the questions:
1. I wonder how serious issue would that `type` being slightly incorrect be actually be in the real world.
2. We should consider is changing the external types to the new type could be a breaking change for any of our users.

@pierrezimmermannbam looking forward for some thoughts on that.

src/helpers/component-tree.ts
mdjastrzebski: Indeed the `type` only accepts string literal union values of HTML tags.
image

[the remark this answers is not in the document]
pierrezimmermannbam: Yes users probably don't know about it, I was more thinking about people that work on this repository, it's important to know that a `HostReactTestInstance` is assignable to a `ReactTestInstance` parameter imo. There are also several representations for elements in `react-test-renderer`, the tree, the instance and the json and it's important to be able to make the difference

pierrezimmermannbam:
I don't think it's a big problem that `type` is incorrectly typed. It was already typed incorrectly before this PR (even more so) and we haven't gotten complaints afaik. It could be a problem for users if they try to directly access the type property and check if e.g. type is 'Text' in which case they'd probably get an error saying that that's impossible. I don't really see a use case for it though and I don't think a lot of users would do that. We could change the type of the public API for the users but I don't think it's worth it, especially since we don't really know what can the host component names can be except for those we use in queries.

This PR would technically be a breaking change since it narrows the return type of queries. It should have very little impact though but still I think it's probably better to release it with the next major version. Also this PR mostly brings benefits for the code maintenance but holds little to no value for users so there's no real inconvenience in delaying its release

mdjastrzebski:
I think that returning covariant (more specialized) type from query functions should not be considered a breaking change, as any variable assignments that user has would still be valid, as `HostInstance` is subtype of `ReactTestInstance`. 

As for TS errors, I've did some checks and using checks like `elements === 'Text'` does not rise TS warnings nor errors. It is only problematic in cases when we would want to pass `View`, `Text`, etc to function accepting the TS type derived from `ReactTestInstance` but that would be also problematic in the first case. 

The one decision I am still contemplating is whether we should make the host type information public or keep it internal only. I think that we should rather keep this internal for now, as it would bring some benefits to us in terms of code simplification/expressiveness, while it could be considered confusing for the users and would require proper documentation, etc.

mdjastrzebski:
@pierrezimmermannbam yes, I am proposing to use the `HostReactTestInstance` inside the library, as it simplifies the code, but not to expose this type outside, and declare returning `ReactTestInstance` as we did so far.
The reasoning behind is is that it will complicate our documentation and learning curve for new users, while most of the them should just be using proper Jest matchers instead of going into implementation details like this.

pierrezimmermannbam:
@mdjastrzebski I agree with your motives, keeping the same public API would indeed make things simpler for users. What bothers me with your changes is that we no longer are certain that queries return host components. I pushed an alternative version that adds an internal type for queries so that we're sure every query return host components while keeping the same public API. It does add a bit of complexity but it seems like a fair price to pay, wdyt?

mdjastrzebski:
@pierrezimmermannbam It seems we are not using queries internally in a way that having return type of `HostReactTestInstance` would benefit the code. Do you know any such cases? If not then perhaps we could just leave the return type on queries as regular `ReactTestInstance`?

pierrezimmermannbam:
@mdjastrzebski  the return type is indeed not used in the code so specifying it doesn't simplify the code but the true motive here is specifying the type. I think that we want all queries to return host components and that typing enforces that. It's like typing the public API, only it's hidden

mdjastrzebski:
Looks good. I've reverted the last commit on InternalQueryAll types, and renamed `HostReactTestInstance to a tad simpler `HostTestInstance`. I've also done some minor cleanup in existing code.

mdjastrzebski (approved):

mdjastrzebski:
This PR has been released in v12.2.1 🚡

PR #1428, opened 2023-07-12
fix: fixing TypeScript accessibility issue for custom render function

FelipeESchmidt:
### Summary

This Pull Request addresses a problem encountered when trying to use a custom render function in TypeScript. The issue throws an error:
"Exported variable 'customRender' has or is using name 'DebugFunction' from external module '/node_modules/@testing-library/react-native/build/render' but cannot be named."

To resolve this problem, we're exporting the DebugFunction interface from the module it's declared in. This change will allow the TypeScript compiler to recognize the DebugFunction type when it's used in external modules, thus eliminating the naming error that is currently being thrown.

### Test plan

The error:

image

With this change the error goes away:

image
image

The error can be solved in the custom render by using `RenderResult` as the return type, but it'd be nice to work directly. 😀

image

### DIFF
--- a/src/render.tsx
+++ b/src/render.tsx
@@ -136,7 +136,7 @@ function updateWithAct(
   };
 }
 
-interface DebugFunction {
+export interface DebugFunction {
   (options?: DebugOptions | string): void;
   shallow: (message?: string) => void;
 }
--- a/src/pure.ts
+++ b/src/pure.ts
@@ -19,6 +19,7 @@ export type {
   RenderOptions,
   RenderResult,
   RenderResult as RenderAPI,
+  DebugFunction,
 } from './render';
 export type { RenderHookOptions, RenderHookResult } from './renderHook';
 export type { Config } from './config';


mdjastrzebski:
@FelipeESchmidt the error you are getting seems weird, afaik TS should be able to use the type's shape without using it's name. Are you experiencing it with an app or are you creating your own librarary/package?

src/render.tsx
```diff
@@ -136,7 +136,7 @@ function updateWithAct(
   };
 }
 
-interface DebugFunction {
```
mdjastrzebski, line 139: You probably need to add it to package public exports in `pure.ts`

FelipeESchmidt:
> @FelipeESchmidt the error you are getting seems weird, afaik TS should be able to use the type's shape without using it's name. Are you experiencing it with an app or are you creating your own librarary/package?

I'm currently using it with an app and the tests gotta use some wrappers, then I've created a custom render to add them around my component that is being tested.

src/render.tsx
FelipeESchmidt: I've added the type DebugFunction to the `pure.ts` export types in my last commit. That's what you meant?

pierrezimmermannbam:
Hello @FelipeESchmidt, thanks or submitting this pr! I can't reproduce the issue, I suspect it's due to a difference in typescript config, could you please copy your `tsconfig.json` file?

FelipeESchmidt:
Sure, follow below:
```
{
  "compilerOptions": {
    "target": "ESNext",
    "moduleResolution": "node",
    "module": "CommonJS",
    "esModuleInterop": true,
    "outDir": "dist",
    "noEmit": false,
    "declaration": true,
    "strict": true,
    "skipLibCheck": true,
    "lib": ["DOM", "ESNext"],
    "jsx": "react",
    "types": ["jest", "node"],
    "allowJs": true
  },
  "include": ["src", "assets"],
  "exclude": ["dist", "scripts", "node_modules"]
}
```
Was you able to create a custom render without the problem? If yes, could you share it?

pierrezimmermannbam:
You see this error because you have `declaration: true` in your `tsconfig.json`. What happens is that TS needs to be able to name types in declaration files and since `DebugFunction` is not exported it cannot name it. Is there a reason why you'd want to generate declaration files? If not you can change your tsconfig to have `declaration: false` and `noEmit: true`. 

I still believe that this type should be exported though since it's part of the public API and there may be valid use cases for wanting to generate declaration files

pierrezimmermannbam (approved):

mdjastrzebski (approved):

FelipeESchmidt:
@pierrezimmermannbam Thank you very much for the explanation! I really don't know if the project's configuration needs to be the way it currently is, since it's the application where I work and I wasn't there during the project's creation. But again, thank you so much!

mdjastrzebski:

PR #1436, opened 2023-07-24
fix: UserEvent longPress on Text component

mdjastrzebski:
### Summary

Fixed small issues with `UserEvent.press()`:
* `longPress` should call `onLongPress` on `Text` element
* avoid calling `onPress` on `TextInput` as it's not supported
* consider `Text` to pressable if it has any press-related event handler

### Test plan

Added `longPress` tests for `Text` and `TextInput`


pierrezimmermannbam (approved):

mdjastrzebski:
🎉 This PR has been released in v12.2.0 🚀

PR #1438, opened 2023-08-04
chore: update basic example with User Event

mdjastrzebski:
### Summary

Update example/basic with UserEvent code.

### Test plan

Example test should pass.


pierrezimmermannbam (approved):
Very nice to already see `userEvent` in action!

PR #1440, opened 2023-08-07
feature: userEvent `clear()`

mdjastrzebski:
### Summary

Method for clearing the content of `TextInput` element. This method is useful as `type()` helper, following OG User Event implementation, is appending text and not replacing it. Hence, users need a way to clear it in some cases. Currently they have to resort to `fireEvent.changeText(element, '')` workaround.

### Test plan

Added tests based on subset `type()` tests.

### DIFF
--- a/src/user-event/__tests__/clear.test.tsx
+++ b/src/user-event/__tests__/clear.test.tsx
@@ -5,7 +5,6 @@ import { render, userEvent } from '../..';
 
 beforeEach(() => {
   jest.useRealTimers();
-  jest.clearAllMocks();
 });
 
 function renderTextInputWithToolkit(props: TextInputProps = {}) {
--- a/src/user-event/__tests__/clear.test.tsx
+++ b/src/user-event/__tests__/clear.test.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import { View, TextInput, TextInputProps } from 'react-native';
 import { createEventLogger } from '../../test-utils/events';
-import { render } from '../../..';
-import { userEvent } from '../..';
+import { render, userEvent } from '../..';
 
 beforeEach(() => {
   jest.useRealTimers();
--- a/src/user-event/__tests__/clear.test.tsx
+++ b/src/user-event/__tests__/clear.test.tsx
@@ -114,11 +114,7 @@ describe('clear()', () => {
     });
 
     const user = userEvent.setup();
-    await expect(
-      user.clear(textInput)
-    ).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"clear() only supports editable elements."`
-    );
+    user.clear(textInput);
 
     expect(textInput.props.value).toBe('Hello!');
   });
@@ -130,11 +126,7 @@ describe('clear()', () => {
     });
 
     const user = userEvent.setup();
-    await expect(
-      user.clear(textInput)
-    ).rejects.toThrowErrorMatchingInlineSnapshot(
-      `"clear() cannot focus on given element due to "pointerEvents" prop."`
-    );
+    user.clear(textInput);
 
     expect(textInput.props.value).toBe('Hello!');
   });
--- a/src/user-event/__tests__/clear.test.tsx
+++ b/src/user-event/__tests__/clear.test.tsx
@@ -29,21 +29,23 @@ function renderTextInputWithToolkit(props: TextInputProps = {}) {
     />
   );
 
+  const textInput = screen.getByTestId('input');
+
   return {
-    ...screen,
     events,
+    textInput,
   };
 }
 
 describe('clear()', () => {
   it('supports basic case', async () => {
     jest.spyOn(Date, 'now').mockImplementation(() => 100100100100);
-    const { events, ...queries } = renderTextInputWithToolkit({
+    const { textInput, events } = renderTextInputWithToolkit({
       value: 'Hello!',
     });
 
     const user = userEvent.setup();
-    await user.clear(queries.getByTestId('input'));
+    await user.clear(textInput);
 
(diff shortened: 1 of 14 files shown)


src/user-event/__tests__/clear.test.tsx
```diff
@@ -0,0 +1,206 @@
+import * as React from 'react';
+import { View, TextInput, TextInputProps } from 'react-native';
+import { createEventLogger } from '../../test-utils/events';
+import { render } from '../../..';
+import { userEvent } from '../..';
+
+beforeEach(() => {
+  jest.useRealTimers();
+  jest.clearAllMocks();
```
MattAgn, line 8: we already have `clearMocks: true` in the jest config so this line is not needed i think

src/user-event/clear.ts
```diff
@@ -0,0 +1,60 @@
...
+): Promise<void> {
+  if (!isHostTextInput(element)) {
+    throw new ErrorWithStack(
+      `clear() works only with host "TextInput" elements. Passed element has type "${element.type}".`,
+      clear
+    );
+  }
+
+  // Skip events if the element is disabled
+  if (element.props.editable === false || !isPointerEventEnabled(element)) {
```
MattAgn, line 22: could we want to log or throw to inform the user the action is not possible? because I think if he did it, he did not realize his component is not editable and it would be nice to notify him of it

src/user-event/clear.ts
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  // 1. Enter element
+  dispatchEvent(element, 'focus', EventBuilder.Common.focus());
+
+  // 2. Select all
+  const previousText = element.props.value ?? element.props.defaultValue ?? '';
+  const selectionRange = {
+    start: 0,
+    end: previousText.length,
```
MattAgn, line 33: since it starts at 0, shoud not the end be `previousText.length -1` ?

src/user-event/type/type.ts
```diff
@@ -79,14 +78,16 @@ export async function type(
   dispatchEvent(element, 'blur', EventBuilder.Common.blur());
 }
 
-async function emitTypingEvents(
+export async function emitTypingEvents(
```
MattAgn, line 81: [nitpick] i often see config options passed as the last argument (usually because they're optional), which makes me wonder if we always want these options? could we want a default config ? or a global config ?

src/user-event/clear.ts
mdjastrzebski: I think this is a good idea to throw a helpful error in such case. 

There seems to be some discrepancy in OG User Event on that front:
* `clear()` and `click()`/`pointer()` seem to throw if element is not enabled
* `type()` seems to silently ignore the fact.

I think we should have consistent semantics in such case, i.e. always throw a helpful error. 

One thing to note, this would be again the Fire Event API which silently ignores such events.

@MattAgn @pierrezimmermannbam @AugustinLF @thymikee any thoughts on that more general subject?

src/user-event/type/type.ts
```diff
@@ -79,14 +78,16 @@ export async function type(
...
+  config: UserEventConfig,
```
MattAgn, line 82: [nitpick] i often see config options passed as the last argument (usually because they're optional), which makes me wonder if we always want these options? could we want a default config ?

MattAgn (approved):

src/user-event/clear.ts
MattAgn: i agree, we should always throw errors and document it well so people are not too surprised it's different with FireEvent. And maybe we could also think if we want to change the behavior of fireEvent in the future

mdjastrzebski: Added throwing error here in case element is not editable nor pressable (pointer events prop).

For other types of elements, I'll open an issue for discussion.

mdjastrzebski: BTW `fireEvent` was already throwing errors in the past when no enable event handler was found but that has been removed. Here are refs to OG issue: https://github.com/callstack/react-native-testing-library/issues/469, and PR https://github.com/callstack/react-native-testing-library/pull/470

src/user-event/clear.ts
mdjastrzebski: No this is actually fine, `0` is before first character, `1` is after first character, so `n` would be after `n` characters.

src/user-event/clear.ts
```diff
@@ -0,0 +1,60 @@
...
+  };
+  dispatchEvent(
+    element,
+    'selectionChange',
+    EventBuilder.TextInput.selectionChange(selectionRange)
+  );
+
+  // 3. Press backspace
+  const finalText = '';
+  await emitTypingEvents(
```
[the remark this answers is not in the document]
mdjastrzebski: `emitTypingEvents` is primitive function that just emits events and does not calculate changes. It expects to receive all of following: key pressed, text after change, text before change.

src/user-event/type/type.ts
mdjastrzebski: This is an internal function (export is not added to `user-event/index.ts`), we need the current config in order to invoke `wait`, etc

src/user-event/clear.ts
MattAgn: oh yes understood

src/user-event/clear.ts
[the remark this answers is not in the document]
MattAgn: oh alright, i did not go check the implementation behind it

mdjastrzebski:
I've removed throwing error on disabled/not focusable element.

mdjastrzebski:
This PR has been released in v12.2.1 🚡

PR #1442, opened 2023-08-07
ScrollView experiments

siepra:
### Summary

I added an experiment for `ScrollView`.

I simply copied the example from official docs https://reactnative.dev/docs/scrollview and attached logger for every event prefixed with 'on' word.

I'm marking it as draft for now as I plan to continue with `FlatList` and `SectionList` within the same PR.

### DIFF
--- a/experiments-app/src/screens/ScrollViewEvents.tsx
+++ b/experiments-app/src/screens/ScrollViewEvents.tsx
@@ -16,10 +16,15 @@ const handleOnScrollEndDrag = buildEventLogger('onScrollEndDrag');
 const handleOnScrollToTop = buildEventLogger('onScrollToTop');
 
 export function ScrollViewEvents() {
+  const handleOnContentSizeChange = (w: number, h: number) => {
+    console.log(`Event: contentSizeChange`, w, h);
+  };
+
   return (
     <SafeAreaView style={styles.container}>
       <ScrollView
         style={styles.scrollView}
+        onContentSizeChange={handleOnContentSizeChange}
         onMomentumScrollBegin={handleOnMomentumScrollBegin}
         onMomentumScrollEnd={handleOnMomentumScrollEnd}
         onScroll={handleOnScroll}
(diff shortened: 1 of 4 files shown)


experiments-app/src/screens/ScrollViewEvents.tsx
```diff
@@ -0,0 +1,56 @@
...
+        style={styles.scrollView}
+        onMomentumScrollBegin={handleOnMomentumScrollBegin}
+        onMomentumScrollEnd={handleOnMomentumScrollEnd}
+        onScroll={handleOnScroll}
+        onScrollBeginDrag={handleOnScrollBeginDrag}
+        onScrollEndDrag={handleOnScrollEndDrag}
+        onScrollToTop={handleOnScrollToTop}
+      >
+        <Text style={styles.text}>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
mdjastrzebski, line 31: Is that text long enough to make the screen larger than viewport? For this experiment to work well the content should take 2+ viewports.

experiments-app/src/screens/ScrollViewEvents.tsx
```diff
@@ -0,0 +1,56 @@
...
+const handleOnMomentumScrollBegin = buildEventLogger('onMomentumScrollBegin');
+const handleOnMomentumScrollEnd = buildEventLogger('onMomentumScrollEnd');
+const handleOnScroll = buildEventLogger('onScroll');
+const handleOnScrollBeginDrag = buildEventLogger('onScrollBeginDrag');
+const handleOnScrollEndDrag = buildEventLogger('onScrollEndDrag');
+const handleOnScrollToTop = buildEventLogger('onScrollToTop');
+
+export function ScrollViewEvents() {
+  return (
+    <SafeAreaView style={styles.container}>
```
mdjastrzebski, line 20: Could you also add similar experiment for flat list. We would want to support both of these.

experiments-app/src/screens/ScrollViewEvents.tsx
siepra: It is twice the viewport on my Pixel 6 device. I can make this text longer though.

siepra:
Oh, sorry for not following the commit naming convention. I focused on the branch name instead (my bad). I promise to obey from now on

mdjastrzebski (approved):
Looks good! I've done some minor tweaks to matching code style of the other experiments.

Issue #1441, opened 2023-08-07
feature: UserEvent `scroll()`


mdjastrzebski:
## Describe the Feature

Create a User Event interaction(s) for scrolling. This seems to be pretty frequent interaction on touch based devices, at the same time there are numerous events involved: `onScroll`, `onMomentumScrollBegin/End`, onScrollBegin/EndDrag`, etc. This feature should work with `ScrollView` as well as `FlatList`/`SectionList` (hopefully these will be same event sequences).

## Possible Implementations

As with other User Event interactions:
1. Create ScrollView/FlatList experiment in the experiments app
2. Conduct scroll interactions (dragging, vertical swiping) on both iOS & Android, record logged events in our Wiki
3. Define unified event sequence roughly matching both iOS & Android, as well as events documentation
4. Write implementation and relevant unit tests

## Related Issues
#1405 - core User Event code
#1386 - `press()` implementation
#1396 - `type()` implementation

siepra:
Hi, do you mind if I tried?

mdjastrzebski:
@siepra go ahead.

In order to achieve the high-quality simulation pls submit experiments app PR first, then fill in the Wiki, only then start to work on the actual code (as described in the issue desc).

mdjastrzebski:
You should log it in experiment, but probably does not happen on actual scrolling

mdjastrzebski:
@siepra step 1 merged. Time for recording logged events on wiki.

siepra:
Great, thank you!
I don't have access to edit wiki pages in this repo, am I right? I'm attaching a file containing my recorded logs along with notes on observations (I'm not sure whether the formatting is fully correct) so you can post it by yourself. Or if there's another way, please let me know!
[wiki-draft.md](https://github.com/callstack/react-native-testing-library/files/12290015/wiki-draft.md)

mdjastrzebski:
@siepra I've copied your research into the [wiki](https://github.com/callstack/react-native-testing-library/wiki/ScrollView-Events).

Now we need to decide on the API for user interactions: 
* what interactions to support
* what should be it's parameters
* what events should they emit

Some initial thoughts about API:
* `user.scrollTo({ x, y })`
  * with x and y being optional in order to allow both vertical or horizontal scroll.
  * We should assume the initial scroll position is (x, y) = (0, 0).
[...]

PR #1445, opened 2023-08-08
feat: user event scroll

siepra:
### Summary

API for user event scrolling

Solves https://github.com/callstack/react-native-testing-library/issues/1441, solves #1450


src/user-event/scroll/__tests__/scroll.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+        onScroll={logEvent('scroll')}
+        onScrollBeginDrag={logEvent('scrollBeginDrag')}
+        onScrollEndDrag={logEvent('scrollEndDrag')}
+        onMomentumScrollBegin={logEvent('momentumScrollBegin')}
+        onMomentumScrollEnd={logEvent('momentumScrollEnd')}
+        onScrollToTop={logEvent('scrollToTop')}
+        testID="scrollable"
+      >
+        <Text>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
mdjastrzebski, line 28: [nit] We won't be able to get the layout measurements in RNTL as it's not running Yoga/native code, so there is no point in making this so long. Just put a single <Text> or <View> inside.

src/user-event/scroll/__tests__/scroll.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+              "pageY": 0,
+              "target": 0,
+              "timestamp": 0,
+              "touches": [],
+            },
+            "persist": [MockFunction],
+          },
+        },
+        {
+          "name": "scroll",
```
mdjastrzebski, line 92: We should replete the scroll event some number of times, perhaps based on the amount of scroll value:
e.g. for scroll y=1000, we could do y=100, y=200, ... y=1000, to simulate that you get intermediate scroll events

siepra:
Right now the tests are failing due to `<ScrollView />` scroll state being shared between all the test cases

mdjastrzebski:
Ready for review. I've:
* transformed API to match the #1450 
* descoped `scrollToTop`
* added missing tests

Pls help review this after my changes: @siepra, @pierrezimmermannbam, @MattAgn, @AugustinLF

src/user-event/scroll/__tests__/scrollTo.test.tsx
```diff
@@ -0,0 +1,265 @@
...
+    expect(events).toMatchSnapshot('scrollTo({ x: [15, 30, 45, 60] })');
+  });
+
+  it('supports horizontal momentum scroll', async () => {
+    const { scrollView, events } = renderScrollViewWithToolkit({
+      horizontal: true,
+    });
+    const user = userEvent.setup();
+
+    await user.scrollTo(scrollView, { momentumX: [0, 10, 20] });
```
pierrezimmermannbam, line 146: you can't have a scroll with only momentum no? I think this should be allowed only when x is also defined

pierrezimmermannbam:
Sorry I just got back from my vacations, I just responded to the api proposal, I think it will be faster to discuss there and align on API before finishing reviewing this pr

src/user-event/scroll/__tests__/scrollTo.test.tsx
mdjastrzebski: Let's discuss that within the [API proposal](https://github.com/callstack/react-native-testing-library/issues/1450#issuecomment-1687756141)

mdjastrzebski:
I've added initial docs for the `scrollTo` event.
I've also decided to descope `scrollToTop` variant from this PR, as we should focus on creating solid base API first and then we can add more rarely-used variants.

siepra (commented):
LGTM. I cannot approve as I'm one of the authors. I wish I was more involved in polishing the final API next time!

website/docs/UserEvent.md
```diff
@@ -178,3 +178,55 @@ The `textInput` event is sent only for mutliline text inputs.
...
+await user.scrollTo(scrollView, { y: 100, momentumY: 200 });
+```
+
+This helper simulates user scrolling a host `ScrollView` element. 
+
+This function supports only host `ScrollView` elements, passing other element types will result in error. Note that `FlatList` is accepted as it renders to a host `ScrolLView` element, however in the current iteration we focus on base `ScrollView` only.
+
+Scroll interaction should match host `ScrollView` direction. For vertical scroll view (`horizontal={false}`) you should pass only `y` (or `momentumY`) options, for horizontal scroll view (`horizontal={true}`) you should pass only `x` (or `momentumX`) options.
+
+Each scroll interaction consists of mandatory drag scroll part which simulates user dragging the scroll view with his finger (`y` or `x` option). This may optinally be followed by momentum scroll movement which simulates the inertial movement of scroll view content after user lifts his finger up (`momentumY` or `momentumX` options).
```
pierrezimmermannbam, line 207: Maybe it could be worth mentioning that we don't have access to content size and so props that rely on it like ` onViewableItemsChanged` or `onEndReached` will never be called?

pierrezimmermannbam: That being said the sequence of events is detailed at the end so it may not be necessary

src/user-event/scroll/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+  return interpolateLinearSteps(
+    targetX,
+    initialOffset.x,
+    DEFAULT_STEPS_COUNT
+  ).map((x) => ({ x, y: initialOffset.y }));
+}
+
+/**
+ * Generates scroll intermediate values.
+ * @param from
```
pierrezimmermannbam, line 45: The jsdoc is not correct here, names of params have changed probably?

src/user-event/scroll/scrollTo.ts
```diff
@@ -0,0 +1,139 @@
...
+import { dispatchEvent } from '../utils';
+import { ContentOffset } from '../event-builder/scroll';
+import {
+  createHorizontalScrollSteps,
+  createVerticalScrollSteps,
+} from './utils';
+import { getElementScrollOffset, setElementScrollOffset } from './state';
+
+export interface VerticalScrollToOptions {
+  y: number | number[];
```
pierrezimmermannbam, line 15: nit: I like adding x?: never in these type of cases so that I can do if(options.x) rather than if('x' in options). It's a more usual syntax and if the prop name changes then there will be an error whereas with if('x' in options) you'll never get an error

pierrezimmermannbam (changes requested):
Great work on this from both of you! Sorry I couldn't review this earlier. I left some minor comments. Also the support for custom steps should be removed based on the discussion in the API proposal

src/user-event/scroll/__tests__/scrollTo.test.tsx
mdjastrzebski: Changed. Now the drag part (`y` or `x`) is mandatory, while momentum part is optional.

website/docs/UserEvent.md
mdjastrzebski: @pierrezimmermannbam good idea to mention content size thing as the current state of things. In the future we could actually work on improving the FlatList features, e.g. by allowing user to pass mock content size props or do some reasonable defaults, etc. I think it might be a topic worth exploring.

src/user-event/scroll/scrollTo.ts
mdjastrzebski: Changed!

src/user-event/scroll/utils.ts
mdjastrzebski: Fixed

mdjastrzebski (approved):
After long review and discussion this PR is finally ready! ✅ That was a long one. Thank you to @siepra for initial version of the PR 🥇, thanks to @pierrezimmermannbam and @MattAgn and others ❤️ for participating in discussion on API shape and implementation. It's time to finally merge it 🌊🏄‍♂️

mdjastrzebski:
This PR has been released [v12.3.1](https://github.com/callstack/react-native-testing-library/releases/tag/v12.3.1) 🚀

PR #1447, opened 2023-08-09
chore: React Test Library experiments app

mdjastrzebski:
### Summary

As one of our state goals is compatibility with web React Testing Library, I've added a basic React app (Next.js) and added RTL + some sample testing code, so that we have a handy test environment for checking RTL behavior.

### Test plan

None, this is not part of RNTL code.


MattAgn:
Great idea!

MattAgn (approved):

PR #1454, opened 2023-08-14
feature: Jest matchers core

mdjastrzebski:
### Summary

Initial PR laying the groundwork for migration of Jest Native matchers to RNTL repo. Other matchers will be covered in separate PR so that the review process can be done properly for each matcher.

These matchers are not enable by default but could be added by doing import into `matchers/extend-expect`.

### Test plan

All tests should pass.


src/helpers/component-tree.ts
```diff
@@ -92,3 +92,22 @@ export function getHostSiblings(
...
   );
 }
+
+/**
+ * Returns the unsafe root element of the tree (probably composite).
+ *
+ * @param element The element start traversing from.
+ * @returns The root element of the tree (host or composite).
+ */
+export function getRootElement(element: ReactTestInstance | null) {
```
thymikee, line 102: if it's unsafe, maybe let's state this explicitly with `getHostOrCompositeRootElement` or `unsafe_getRootElement`?

mdjastrzebski: it's internal fn, none of functions in this file are exported publically, so it's safe for us ;-)
Anyway I'll go for `getUnsafeRootElement()`

thymikee (approved):

src/helpers/component-tree.ts
thymikee: I believe it should raise awareness in contributors as well :)

mdjastrzebski:

PR #1461, opened 2023-08-19
feat: toHaveTextContent() matcher [WIP]

mdjastrzebski:
### Summary

Implement `toHaveTextContent` matcher.

### Test plan

Added relevant unit tests.


pierrezimmermannbam (approved):
Looks good to me! I just have a question, I understood that we would start by exposing jest native matchers prefixed with `legacy` to ease migration, is it still the plan? If so shouldn't we start with that?

mdjastrzebski:
@pierrezimmermannbam re `legacy` matchers. Yes that is still the plan. I plan to do it in the `@testing-library/jest-native` matchers as optional import: `import '@testining-library/jest-native/legacy-extend-expect` instead of typical `import '@testining-library/jest-native/extend-expect`.

That way it would be still available for a very long time but we would not have to manage it in RNTL repo.

mdjastrzebski:

PR #1462, opened 2023-08-19
feat: `toBeEmptyElement` matcher

### LINKED ISSUE #1459: feature: implement `toBeEmptyElement` Jest matcher
## Describe the Feature

[`toBeEmptyDOMElement`](https://github.com/testing-library/jest-dom#tobeemptydomelement) is a Jest DOM matcher that asserts the visibility of given element by inspecting whether it contains any child elements.

Adapting it to our case it would assert that given host element does not contain any host children elements. It might contain some composite elements as long as these do not have host descendants. We already have function for getting host children of given element ([`getHostChildren`](https://github.com/callstack/react-native-testing-library/blob/4bfd7fbc6ba76ad376e24f3d41be1d638b6351a2/src/helpers/component-tree.ts#L45)) so that should be relatively straightforward.

## Possible Implementations
Proposed API:

```ts
export function toBeEmptyElement(
  this: jest.MatcherContext,
  element: ReactTestInstance
)
```

The matcher should:
1) Validate that it is invoked on a host element
2) Check if given element contains any host child elements (using `getHostChildren` fn)

Each matcher should have a fairly comprehensive test suite.

## Links

* `toBeEmptyDOMElement` matcher code from Jest DOM: https://github.com/testing-library/jest-dom/blob/main/src/to-be-empty-dom-element.js
* `toBeEmptyElement` implementation for Jest Native: https://github.com/testing-library/jest-native/blob/main/src/to-be-empty-element.ts

## Related Issues
#1454

kyawthura-gg:
### Summary

Fixes #1459 

### Test plan

`yarn test`

### DIFF
--- a/src/matchers/__tests__/to-be-empty-element.test.tsx
+++ b/src/matchers/__tests__/to-be-empty-element.test.tsx
@@ -1,0 +1,30 @@
+import React from 'react';
+import { View } from 'react-native';
+import { render, screen } from '../..';
+import '../extend-expect';
+
+test('.toBeEmptyElement', () => {
+  render(
+    <View testID="not-empty">
+      <View testID="empty" />
+    </View>
+  );
+
+  const empty = screen.getByTestId('empty');
+  const notEmpty = screen.queryByTestId('not-empty');
+  const nonExistentElement = screen.queryByTestId('not-exists');
+  const fakeElement = { thisIsNot: 'an html element' };
+
+  expect(empty).toBeEmptyElement();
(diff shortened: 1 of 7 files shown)


kyawthura-gg:
@mdjastrzebski, while running same command `yarn typecheck` does not show any error locally. Not sure what is the issue, may be cache?

mdjastrzebski (approved):
@kyawthura-gg Great work with this PR. Before merging I've tweaked the error formatting and test suit.

mdjastrzebski:

Issue #1458, opened 2023-08-17
feature: implement `toBeVisible` Jest matcher


mdjastrzebski:
## Describe the Feature

`toBeVisible` is a [Jest DOM matcher](https://github.com/testing-library/jest-dom#tobevisible) that asserts the visibility of given element by inspecting various CSS styles and other semantic attributes that might make element invisible.

Adapting it to our case it would assert the value in `TextInput` host elements, to match given value.

## Possible Implementations
Proposed API:

```ts
export function toBeVisible(
  this: jest.MatcherContext,
  element: ReactTestInstance
)
```

The matcher should:
1) Validate that it is invoked on host element
2) Check visibility status of given element and all of it's ancestors

Each matcher should have a fairly comprehensive test suite.

Open Questions:
* should this matcher take into account accessibility props? Consider use cases and pros and cons,

## Links

* `toBeVisible` matcher code from Jest DOM: https://github.com/testing-library/jest-dom/blob/main/src/to-be-visible.js
* `toBeVisible` implementation for Jest Native: https://github.com/testing-library/jest-native/blob/main/src/to-be-visible.ts

## Related Issues
#1454

thiagobrez:
Hey, would love to give a try at this one!

mdjastrzebski:
@thiagobrez go ahead! Simply submit PR with your work in progress.

thiagobrez:
PR is ready to review here: https://github.com/callstack/react-native-testing-library/pull/1465

mdjastrzebski:
🎉 Release in version: [v12.4.0](https://github.com/callstack/react-native-testing-library/releases/tag/v12.4.0)

PR #1465, opened 2023-08-21
feat: `toBeVisible` matcher

thiagobrez:
### Summary

https://github.com/callstack/react-native-testing-library/issues/1458 - Implement `toBeVisible()` matcher.

Follows [Jest DOM's visibility rules](https://github.com/testing-library/jest-native/blob/main/src/to-be-visible.ts)
Based on [Jest Native's toBeVisible()](https://github.com/testing-library/jest-native/blob/main/src/to-be-visible.ts)

Style checks:
- `display !== 'none'`
- `opacity !== 0`

Accessibility checks:
- `accessibilityElementsHidden` (iOS)
- `importantForAccessibility !== 'no-hide-descendants'` (Android)
- [NEW] `aria-hidden`. Looking through [RN's accessibility docs](https://reactnative.dev/docs/accessibility#aria-hidden), this seemed to make sense to add. WDYT?

Modal checks:
- `visible !== false`

### Test plan

Added relevant unit tests.


src/matchers/to-be-visible.tsx
```diff
@@ -0,0 +1,64 @@
...
+function isVisibleForAccessibility(element: ReactTestInstance) {
+  return (
+    !element.props.accessibilityElementsHidden &&
+    element.props.importantForAccessibility !== 'no-hide-descendants' &&
+    !element.props['aria-hidden']
+  );
+}
+
+function isModalVisible(element: ReactTestInstance) {
+  return element.type.toString() !== 'Modal' || element.props.visible !== false;
```
thiagobrez, line 22: Had to check `element.type` here converted to String instead of using the actual `Modal` class (as done in Jest Native). Any clues why @mdjastrzebski? Seems to work with no downside though

mdjastrzebski: I think we should skip the Modal code, as it relies heavily on RN internal implementation details.

mdjastrzebski:
Rebased on main, solved TS def issue.

src/matchers/__tests__/to-be-visible.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+        <View testID="test" />
+      </View>
+    </View>
+  );
+  expect(
+    getByTestId('test', { includeHiddenElements: true })
+  ).not.toBeVisible();
+});
+
+test('toBeVisible() on empty Modal', () => {
```
mdjastrzebski, line 58: Optimally we should support Modal in test but do not rely on it internallny.

src/matchers/__tests__/to-be-visible.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const { getByTestId } = render(<Modal testID="test" />);
+  expect(getByTestId('test')).toBeVisible();
+});
+
+test('toBeVisible() on view within modal', () => {
+  const { getByTestId } = render(
+    <Modal>
```
mdjastrzebski, line 65: add explict visible flag to make test more readable

src/matchers/__tests__/to-be-visible.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+        <View testID="view-within-modal" />
+      </View>
+    </Modal>
+  );
+
+  expect(getByTestId('test')).not.toBeVisible();
+  // Children elements of not visible modals are not rendered.
+  expect(() =>
+    expect(getByTestId('view-within-modal')).not.toBeVisible()
+  ).toThrow();
```
mdjastrzebski, line 87: Use: `toThrowErrorMatchingInlineSnapshot` in such cases

src/matchers/__tests__/to-be-visible.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+    expect(getByTestId('test')).not.toBeVisible()
+  ).toThrowErrorMatchingSnapshot();
+
+  update(<View testID="test" style={{ opacity: 0 }} />);
+  expect(() =>
+    expect(getByTestId('test')).toBeVisible()
+  ).toThrowErrorMatchingSnapshot();
+});
+
+test('toBeVisible() on Pressable with function style prop', () => {
```
mdjastrzebski, line 201: What is the purpose of this test?

src/matchers/__tests__/to-be-visible.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+test('toBeVisible() throws an error when expectation is not matched', () => {
+  const { getByTestId, update } = render(<View testID="test" />);
+  expect(() =>
+    expect(getByTestId('test')).not.toBeVisible()
+  ).toThrowErrorMatchingSnapshot();
+
+  update(<View testID="test" style={{ opacity: 0 }} />);
+  expect(() =>
+    expect(getByTestId('test')).toBeVisible()
+  ).toThrowErrorMatchingSnapshot();
```
mdjastrzebski, line 198: inline snapshot pls.

src/matchers/to-be-visible.tsx
```diff
@@ -0,0 +1,64 @@
+import type { ReactTestInstance } from 'react-test-renderer';
+import { matcherHint } from 'jest-matcher-utils';
+import { StyleSheet } from 'react-native';
+import { getHostParent } from '../helpers/component-tree';
+import { checkHostElement, formatElement } from './utils';
+
+function isVisibleForStyles(element: ReactTestInstance) {
+  const style = element.props.style || {};
```
mdjastrzebski, line 8: `||` => `??`

src/matchers/to-be-visible.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { display, opacity } = StyleSheet.flatten(style);
+  return display !== 'none' && opacity !== 0;
+}
+
+function isVisibleForAccessibility(element: ReactTestInstance) {
+  return (
+    !element.props.accessibilityElementsHidden &&
+    element.props.importantForAccessibility !== 'no-hide-descendants' &&
+    !element.props['aria-hidden']
```
mdjastrzebski, line 17: Afaik `aria-hidden` is not passed to host view but translated into `accessibilityElementsHidden` for ios and `importantForAccessibility` so we should be able to skip this in check and still pass the tests.

thiagobrez: Makes perfect sense - thanks!

thiagobrez: @mdjastrzebski You're right, but the test still fails when rendering a `<View />`, because RN only translates `aria-hidden` to the accessibility props when rendering the native component itself, which does not get caught by the test renderer

#### View.js
```jsx
      <ViewNativeComponent
        ...
        accessibilityElementsHidden={ariaHidden ?? accessibilityElementsHidden}
        importantForAccessibility={
          ariaHidden === true
            ? 'no-hide-descendants'
            : importantForAccessibility
        }
        ...
      />
```

My second choice would be to use `<Pressable />`, since we're already using it in this test file, but it also fails because of a bug in RN, I believe. This component accepts `aria-hidden` as a prop but only passes it down to the underlying `View`.

In contrast, using any `Touchable` makes the test pass, because their implementation also translates the `aria-hidden` prop into the accessibility props, and they do not render a native component directly, which makes the test renderer aware of these props.

What do you think? Should we use one of the `Touchable`s or you prefer not to rely this much on RN's implementation details?

mdjastrzebski: @thiagobrez Can you post `screen.toJSON()` result when rendering `View` with `aria-hidden` prop?

mdjastrzebski: I am really surprised now 👀

mdjastrzebski: Pls restore `aria-hidden` check then

src/matchers/__tests__/to-be-visible.test.tsx
thiagobrez: To test that the matcher still works when passing `style` as a function, which is exclusive to the `Pressable` component. But I can also see the point for removing it :)

src/matchers/to-be-visible.tsx
thiagobrez: @mdjastrzebski You're saying to remove the Modal check? To me it feels like an edge-case that should be handled, equivalent to the `<details />` check in Jest DOM

src/matchers/__tests__/to-be-visible.test.tsx
mdjastrzebski: AFAIK that style as a function is working on Pressable composite element level, but get's transformed to style object on the underlying host View element.

thiagobrez: Yep! Ok, I'll remove it

src/matchers/to-be-visible.tsx
jaworek: `jest-native` has an util for checking type of an element. I imported it in my PR #1469. 

https://github.com/testing-library/jest-native/blob/main/src/utils.ts#L68

mdjastrzebski: Let's not do it the way @jaworek suggested. This method is flawed, as it relies on RN implementation details which might change. We have some tools for checking host element type, but should limit it to core components like `Text`, `TextInput`, `View`, etc. In case of Modal seems like pretty primitive mock is actually used: https://github.com/facebook/react-native/blob/main/packages/react-native/jest/mockModal.js

mdjastrzebski: This is host element tree for modals:

1. `visible={true}`
```
test('Modal render', () => {
  render(
    <Modal>
      <View testID="content" />
    </Modal>
  );

  expect(screen.toJSON()).toMatchInlineSnapshot(`
    <Modal
      hardwareAccelerated={false}
      visible={true}
    >
      <View
        testID="content"
      />
    </Modal>
  `);
});
```

1. `visible={false}`
```
test('Modal render', () => {
  render(
    <Modal visible={false}>
      <View testID="content" />
    </Modal>
  );

  expect(screen.toJSON()).toMatchInlineSnapshot(`
    <Modal
      hardwareAccelerated={false}
      visible={false}
    />
  `);
});

```

thiagobrez: @mdjastrzebski Ok, makes sense. But I did not understand what action are you suggesting to take. Remove the `isModalVisible()` check and tests?

mdjastrzebski: So regarding **content** of Modal we are good, as this won't be rendered at all. However, the `Modal` host element being rendered when modal is not visible seems incorrect. I'll drop a quick PR there to make the modal not render any host elements if it's not visible.

As a workaround we should be to extend [`hostComponentNames`](https://github.com/callstack/react-native-testing-library/blob/cf08c069c2a0b573297c30e1a920523281d48eeb/src/helpers/host-component-names.tsx#L31) feature to detect the Modal host component name, and then use that value to make check in the Jest matcher, similar to the one from legacy Jest Native matcher.

thiagobrez:
@mdjastrzebski Comments addressed!

mdjastrzebski:
Rebased on main

src/matchers/to-be-visible.tsx
mdjastrzebski: Sent PR to RN repo https://github.com/facebook/react-native/pull/39157

mdjastrzebski (approved):
Looks good! Let's 🚢

mdjastrzebski:

Issue #1468, opened 2023-08-22
API proposal: Jest Native Matchers


mdjastrzebski:

MattAgn:
Fully agree with all of the above! I'll try to take one or two tasks  when i come back from vacation in a week or so.
Quick naming question: why go for `toHaveAccessibleName` instead of `toHaveAccessibilityLabel` or `toHaveAccessibleLabel` to stay coherent with the prop name?

pierrezimmermannbam:
That seems perfect! Great work @mdjastrzebski !

mdjastrzebski:
Update: Jest Native package includes legacy extend as of 5.4.3 🚀🎉

This is not yet publicly documented, that will be done after we complete RNTL matcher migration.

In order to use legacy extend `expect(...)`:
1. Use `import '@testing-library/jest-native/legacy-extend-expect` instead of  `import '@testing-library/jest-native/extend-expect`
2. Use `expect(...).legacy_toXxx` matchers instead of `expect(...).toXxx`.

In the default setup nothing changes, so you need the changed import to enable `legacy_toXxx` matchers.

AntoineThibi:
Hello ! 
I created a PR for `toHaveProp` matcher : https://github.com/callstack/react-native-testing-library/pull/1477

hduprat:
Hi everyone,

Here is a PR for the `toBeBusy()` matcher: #1493

AntoineThibi:
Hello !

Here is one for the `toBeSelected()` matcher : https://github.com/callstack/react-native-testing-library/pull/1488

siepra:
Hi, is anyone currently working on `toContainElement` or am I free to go?

mdjastrzebski:
@siepra go ahead, it's yours now.

siepra:
Here is also a PR for the `toBeExpanded()` matcher: https://github.com/callstack/react-native-testing-library/pull/1497

PR #1477, opened 2023-08-29
feat: `toHaveProp` matcher

AntoineThibi:
### Summary

Implement `toHaveProp` matcher

### Test plan

`yarn test`

### DIFF
--- a/src/matchers/__tests__/to-have-prop.test.tsx
+++ b/src/matchers/__tests__/to-have-prop.test.tsx
@@ -3,7 +3,7 @@ import { View, Text, TextInput } from 'react-native';
 import { render, screen } from '../..';
 import '../extend-expect';
 
-test('.toHaveProp basic case', () => {
+test('toHaveProp() basic case', () => {
   render(
     <View testID="view" style={null}>
       <Text ellipsizeMode="head">Hello</Text>
@@ -30,7 +30,7 @@ test('.toHaveProp basic case', () => {
   expect(input).not.toHaveProp('editable', false);
 });
 
-test('.toHaveProp error messages', () => {
+test('toHaveProp() error messages', () => {
   render(<View testID="view" collapsable={false} />);
 
   const view = screen.getByTestId('view');
--- a/src/matchers/__tests__/to-have-prop.test.tsx
+++ b/src/matchers/__tests__/to-have-prop.test.tsx
@@ -1,79 +1,87 @@
 import React from 'react';
-import { Button, Text, View } from 'react-native';
+import { View, Text, TextInput } from 'react-native';
 import { render, screen } from '../..';
 import '../extend-expect';
 
-test('.toHaveProp', () => {
+test('.toHaveProp basic case', () => {
   render(
-    <View style={null} testID="view">
-      <Text allowFontScaling={false} testID="text" ellipsizeMode="head">
-        text
-      </Text>
-      <Button disabled testID="button" title="ok" />
+    <View testID="view" style={null}>
+      <Text ellipsizeMode="head">Hello</Text>
+      <TextInput testID="input" textAlign="right" />
     </View>
   );
 
-  const text = screen.getByTestId('text');
-  const button = screen.getByTestId('button');
+  const view = screen.getByTestId('view');
+  expect(view).toHaveProp('style');
+  expect(view).toHaveProp('style', null);
+  expect(view).not.toHaveProp('ellipsizeMode');
+
+  const text = screen.getByText('Hello');
+  expect(text).toHaveProp('ellipsizeMode');
+  expect(text).toHaveProp('ellipsizeMode', 'head');
+  expect(text).not.toHaveProp('style');
+  expect(text).not.toHaveProp('ellipsizeMode', 'tail');
+
+  const input = screen.getByTestId('input');
+  expect(input).toHaveProp('textAlign');
+  expect(input).toHaveProp('textAlign', 'right');
+  expect(input).not.toHaveProp('textAlign', 'left');
+  expect(input).not.toHaveProp('editable');
+  expect(input).not.toHaveProp('editable', false);
+});
+
+test('.toHaveProp error messages', () => {
+  render(<View testID="view" collapsable={false} />);
+
   const view = screen.getByTestId('view');
 
-  expect(button).toHaveProp('accessibilityState', {
-    disabled: true,
-  });
-  expect(text).toHaveProp('ellipsizeMode', 'head');
-  expect(text).toHaveProp('allowFontScaling', false);
-
-  expect(button).not.toHaveProp('accessibilityStates');
-  expect(button).not.toHaveProp('ellipsizeMode', undefined);
-  expect(button).not.toHaveProp('allowFontScaling', false);
-  expect(text).not.toHaveProp('style');
-
-  // title is no longer findable as it is a React child
-  expect(() => expect(button).toHaveProp('title', 'ok'))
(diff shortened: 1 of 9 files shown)


src/matchers/__tests__/to-have-prop.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  expect(queryByTestId('text')).toHaveProp('ellipsizeMode', 'head');
+  expect(queryByTestId('text')).toHaveProp('allowFontScaling', false);
+
+  expect(queryByTestId('button')).not.toHaveProp('accessibilityStates');
+  expect(queryByTestId('button')).not.toHaveProp('ellipsizeMode', undefined);
+  expect(queryByTestId('button')).not.toHaveProp('allowFontScaling', false);
+  expect(queryByTestId('text')).not.toHaveProp('style');
+
+  // title is no longer findable as it is a React child
+  expect(() =>
```
pierrezimmermannbam, line 34: I guess it was like that in jest-native matcher but wouldn't it be better to use not rather than expect the contrary to throw?

pierrezimmermannbam: Oh I realize now using this could allow to test the message, can you replace to throw by toThrowErrorMatchingInlineSnapshot()?

src/matchers/to-have-prop.ts
```diff
@@ -0,0 +1,52 @@
...
+    : `element.getAttribute(${name}) === ${stringify(value)}`;
+}
+
+export function toHaveProp(
+  this: jest.MatcherContext,
+  element: ReactTestInstance,
+  name: string,
+  expectedValue: unknown
+) {
+  checkReactElement(element, toHaveProp, this);
```
pierrezimmermannbam, line 21: You should use the already existing method `checkHostElement` and delete this one

src/matchers/utils.tsx
```diff
@@ -121,3 +121,69 @@ export function formatMessage(
 function formatValue(value: unknown) {
   return typeof value === 'string' ? value : stringify(value);
 }
+
+function display(value: unknown) {
```
pierrezimmermannbam, line 125: I think all of the code added here can be removed if you use checkHostElement instead

pierrezimmermannbam (changes requested):
Hello @AntoineThibi, thanks for the contribution! I've left some comments to address before this can be merged

src/matchers/__tests__/to-have-prop.test.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import { Button, Text, View } from 'react-native';
+import { render } from '../..';
+import '../extend-expect';
```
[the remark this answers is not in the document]
mdjastrzebski: Imo that could be moved to jest-setup.js assuming that it will work ;-)

src/matchers/to-have-prop.ts
mdjastrzebski: Yes, `checkHostElement` is the check to use

src/matchers/utils.tsx
```diff
@@ -121,3 +121,69 @@ export function formatMessage(
...
+  return typeof value === 'string' ? value : stringify(value);
+}
+
+export function getMessage(
```
mdjastrzebski, line 129: ditto: pls use `formatMessage` already present in this file instead.

src/matchers/utils.tsx
```diff
@@ -121,3 +121,69 @@ export function formatMessage(
...
+  receivedValue: string | null
+) {
+  return [
+    `${matcher}\n`,
+    `${expectedLabel}:\n${EXPECTED_COLOR(redent(display(expectedValue), 2))}`,
+    `${receivedLabel}:\n${RECEIVED_COLOR(redent(display(receivedValue), 2))}`,
+  ].join('\n');
+}
+
+class ReactElementTypeError extends Error {
```
mdjastrzebski, line 143: ditto: pls use `checkHostComponent` already present in this file instead.

src/matchers/utils.tsx
```diff
@@ -121,3 +121,69 @@ export function formatMessage(
...
+        ''
+      ),
+      '',
+      `${RECEIVED_COLOR('received')} value must be a React Element.`,
+      withType,
+    ].join('\n');
+  }
+}
+
+export function checkReactElement(
```
mdjastrzebski, line 176: ditto: pls use `checkHostComponent` already present in this file instead.

src/matchers/to-have-prop.ts
```diff
@@ -0,0 +1,52 @@
...
+import { checkReactElement, getMessage } from './utils';
+
+function printAttribute(name: string, value: unknown) {
+  return value === undefined ? name : `${name}=${stringify(value)}`;
+}
+
+function getPropComment(name: string, value: unknown) {
+  return value === undefined
+    ? `element.hasProp(${name})`
+    : `element.getAttribute(${name}) === ${stringify(value)}`;
```
mdjastrzebski, line 12:
```suggestion
  return value === undefined
    ? `Element should have prop ${name}`
    : `Element should have prop ${name} with value ${stringify(value)}`;
```

src/matchers/to-have-prop.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export function toHaveProp(
+  this: jest.MatcherContext,
+  element: ReactTestInstance,
+  name: string,
+  expectedValue: unknown
+) {
+  checkReactElement(element, toHaveProp, this);
+
+  const prop = element.props[name];
```
mdjastrzebski, line 23:
```suggestion
  const propValue = element.props[name];
```

src/matchers/__tests__/to-have-prop.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+  render(
+    <View style={null} testID="view">
+      <Text allowFontScaling={false} testID="text" ellipsizeMode="head">
+        text
+      </Text>
+      <Button disabled testID="button" title="ok" />
+    </View>
+  );
+
+  const text = screen.queryByTestId('text');
```
mdjastrzebski, line 16: [nit] queryBy => getBy

src/matchers/__tests__/to-have-prop.test.tsx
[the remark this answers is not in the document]
AntoineThibi: It doesn't work with the `extend-expect.test` which checks that those new methods are not present in the expect by default.

AntoineThibi:
@pierrezimmermannbam @mdjastrzebski I've made the modifications !

src/matchers/__tests__/to-have-prop.test.tsx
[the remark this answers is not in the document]
pierrezimmermannbam: Oh indeed I hadn't thought about that. I can't think of a simple solution to solve this, creating a jest config only for this file seems overkill for what we're trying to achieve. In that case either we remove that test or we keep the imports, but I think it's better to just keep the imports, this test is rather valuable and having to add imports is not a huge downside

pierrezimmermannbam (approved):
That's ok for me, thanks @AntoineThibi ! Just need to wait for @mdjastrzebski 's approval before we merge

mdjastrzebski (approved):
@AntoineThibi great work! 💯 I've done some final polishing before merging.

If you have spare time please pick some more Jest matchers from #1468.

mdjastrzebski:

PR #1478, opened 2023-08-29
feat: `aria-*` state props

mdjastrzebski:
### Summary

Add support for `aria-disabled`, `aria-selected`, `aria-checked`, `aria-busy` and `aria-expanded` state props.

### Test plan

Added & updated relevant tests.


pierrezimmermannbam (approved):
What a streak of prs! Great idea to add support for those props!

PR #1480, opened 2023-08-31
feat: aria-value* props support

mdjastrzebski:


src/helpers/accessiblity.ts
```diff
@@ -160,3 +164,33 @@ export function getAccessibilityState(element: ReactTestInstance) {
...
+): AccessibilityValue | undefined {
+  const {
+    accessibilityValue,
+    'aria-valuemax': ariaValueMax,
+    'aria-valuemin': ariaValueMin,
+    'aria-valuenow': ariaValueNow,
+    'aria-valuetext': ariaValueText,
+  } = element.props;
+
+  const hasAnyAccessibilityValueProps =
```
pierrezimmermannbam, line 179: nit: here I think that a condition hasNoAccessibilityValueProps would be a bit clearer because it would remove the negation in the next if

pierrezimmermannbam (approved):
Just a small comment but otherwise this looks great!

PR #1490, opened 2023-09-06
feat: default accessibility roles

mdjastrzebski:
### Summary

This PR improves accessibility role handling by supporting default following default roles:
* `text` role for Text elements
* `textbox`* role for TextInput elements
* `none` role for all other elements

Note: `textbox` is not an official RN role, you cannot set it explicitly on elements. However, returning `none` role for TextInput elements does not feel correct as they have significantly different default accessibility behavior that default accessible Views. Therefore, I propose to use [official ARIA role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/textbox_role), which despite not being supported by RN, better describes the characteristics of TextInput elements.

### Test plan

Updated `*ByRole` tests to check for default role.


MattAgn:
I'm not sure what to think of the `textbox` role. I find it a bit misleading to have an API that 99% of the time returns React Native roles but for one case does not, and instead returns a role based on web standards. 
I'm not sure being able to do ` expect(screen.getByRole('textbox').props.testID).toBe('text-input-default')` is so important that we need to introduce web behavior in our RN library.
however i understand your point that "none" is not that great either for text box 😅

mdjastrzebski:
@pierrezimmermannbam, @MattAgn thanks for your feedback, I'll work on explaining my reasoning properly after RNEU

mdjastrzebski:
@pierrezimmermannbam, @MattAgn Let me answer the easier part of your feedback first: do we need default roles for `View` and `Text` and why not just require explicit roles. 

I've done a11y experiments on both iOS and Android and verified that:
1) `View` with `none` role behaves the same as `View` without explicit role. Moreover, assigning explicit `none` role to `Text` also makes it behave the same.
2) `Text` with `text` role behaves the same as `Text` without explicit role. Moreover, assigning explicit `text` role to `View` also makes it behave the same (+ adding `accessibilityLabel` as a mean of passing "text" to a `View`)

I'll do some more experiments for `TextInputs` next and post the results and thoughs in this PR.

deansimcox (approved):
Appreciate your work on this 🙏

mdjastrzebski:
@pierrezimmermannbam @MattAgn I've reduced the scope of this PR to default `text` role for `Text` component, and `none` role for all other components without explicit role. Pls review.

PR #1491, opened 2023-09-07
Fix: wrap wait with act

pierrezimmermannbam:
### Summary

I noticed on one of my project where I migrated `fireEvent` to `userEvent` that I got an act warning once when using `userEvent.press`. What I found is that it was due to the fact that the `wait` function does not use `act`, although it advances timers. This PR fixes that issue by using `act` in `wait`.

### Test plan

I added a test case but it's not perfect. Before the fix there was a warning in the console but the test still worked. Because `wait` is async, the useEffect is still run even when not using act . The only solution I found to get the test to fail was to spy on `console.error` and make sure it's not called.


mdjastrzebski:
@pierrezimmermannbam it seems to make sense to wrap `wait` in async `act`, similarly as we do e.g. with `fireEvent`, etc. 

However, I've noticed that original [User Event](https://github.com/testing-library/user-event) actually does not seem to use `act` at all, not for `wait` and not for dispatching events. I wonder how is that possible. Do you have any thoughts on that?

mdjastrzebski:
I've found some clues about the original User Event, so since UE is not React specific it does not use `act` function but RTL passed `eventWrapper`(https://github.com/testing-library/react-testing-library/blob/c04b8f006c5a683d05c460c8ee1e2248d6f74350/src/pure.js#L59C7-L59C7) option to DTL.

The `act` function is later applied using `wrapEvent` helper to [`dispatchEvent`](https://github.com/testing-library/user-event/blob/7a305dee9ab833d6f338d567fc2e862b4838b76a/src/event/dispatchEvent.ts#L54). However, curiously UE does NOT wrap `wait` helper, and as far as I observer there is no code that would look like their `wait` is wrapped in `act` either. 🤔

pierrezimmermannbam:
@mdjastrzebski that's interesting, I'll try to play with RTL's user event and see if I can reproduce the issue I encountered

mdjastrzebski:
@pierrezimmermannbam is the issue you are trying to fix caused by this weird 130 ms delay introduced by React Native? If so then perhaps it would be good idea to just wrap that one `wait` call with `async act`. I

pierrezimmermannbam:
I've done some testing with RTL using about the same test:

```tsx
it.only('is wrapped by act', async () => {
  const consoleErrorSpy = jest.spyOn(console, 'error');
  jest.useRealTimers();
  const TestComponent = () => {
    const [isVisible, setIsVisible] = React.useState(false);

    React.useEffect(() => {
      setTimeout(() => {
        setIsVisible(true);
      }, 100);
    }, []);

    return (
      <>
        <input placeholder="toto" />
        {isVisible && <span>Visible</span>}
      </>
    );
  };

  render(<TestComponent />);

  await userEvent.type(screen.getByPlaceholderText('toto'), 'al', {
    delay: 200,
  });
});
```

There is no act warning and what I've found is that the `asyncWrapper` is used so my guess is that every async operation is wrapped by act so the wait will be wrapped as well.

In our case it's not linked to the delay for the press out, any async code could be run while the user performs an action, as demonstrated in the test case I wrote in this PR, so I think it's a valid use case for using `act`

mdjastrzebski:
@pierrezimmermannbam `asyncWrapper` ([drain microtask queue](https://github.com/testing-library/react-testing-library/blob/c04b8f006c5a683d05c460c8ee1e2248d6f74350/src/pure.js#L36) while preserving `act` env) and `eventWrapper` ([wrap with `act`](https://github.com/testing-library/react-testing-library/blob/c04b8f006c5a683d05c460c8ee1e2248d6f74350/src/pure.js#L59)) seem to be different tools. As far as I understand they wrap each interaction in `asyncWrapper` ([here](https://github.com/testing-library/user-event/blob/d7483f049a1ec2ebf1ca1e2c1f4367849fca5997/src/setup/setup.ts#L163)).

More interesting, they seem to add a dummy `wait` call after each interaction: [here](https://github.com/testing-library/user-event/blob/d7483f049a1ec2ebf1ca1e2c1f4367849fca5997/src/setup/setup.ts#L130)

pierrezimmermannbam:
@mdjastrzebski you're right, maybe we could try a similar approach with userEvent? This way we could make sure that every interaction is wrapped by act and prevent similar bugs from happening. Interesting also that they add a call to `wait` at the end, I guess we could mimic that behavior as well, wdyt?

mdjastrzebski:
@pierrezimmermannbam afaik each interaction in UE is not wrapped by `act` (`eventWrapper`) but by `asyncWrapper` which drains microtask queue without invoking `act`.

Not sure about the purpose of that `wait` at the end.

pierrezimmermannbam:
@mdjastrzebski, looking again at this [RFC](https://github.com/reactwg/react-18/discussions/102) on `IS_REACT_ACT_ENVIRONMENT `, it looks like we should only use `act` for interactions and not when waiting (e.g. when using `waitFor` or `wait`). Also I saw an `act` warning being thrown in a codebase when using `userEvent.type` and removing the calls to `wait` didn't fix the issue so the previous fix was not really relevant. 

I updated the PR to mimic the behavior from RTL. We wrap every API from `userEvent` with the same async wrapper we use for `waitFor` which would be similar to their `asyncWrapper` and we keep `act` in dispatch event, which would play the role of RTL's `eventWrapper`.

I was not too sure how to test this, I added a test case where an act warning would have been thrown if act environment wasn't set to false. I doesn't test every API but I don't really think it's worthwhile to write a test for every API. If this doesn't work as expected then we could add more tests if we feel they're valuable.

I also kept the `wait` in `wrapAndBindImpl`. It makes sense to wait between each user interaction imo and the closer our API is from RTL's the better

experiments-rtl/src/app/__tests__/click.test.tsx
```diff
@@ -25,7 +25,36 @@ test("fireEvent.click()", () => {
     </button>
   );
 
-  const button = screen.getByText("Click");
+  const button = screen.getByText('Click');
   fireEvent.click(button);
   expect(handleClick).toHaveBeenCalledTimes(1);
 });
+
+it.only('is wrapped by act', async () => {
```
mdjastrzebski, line 33: pls remove `.only`
```suggestion
it('is wrapped by act', async () => {
```

src/user-event/__tests__/act.test.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { Pressable, Text } from 'react-native';
+import render from '../../render';
+import { userEvent } from '..';
+import { screen } from '../../screen';
+
+test('user event disables act environmennt', async () => {
```
mdjastrzebski, line 7: test name does not seem to match assertion, could you adjust either of these so they match, alternatively add code comment to explain what is happening

src/user-event/setup/setup.ts
```diff
@@ -143,13 +145,33 @@ function createInstance(config: UserEventConfig): UserEventInstance {
...
+    scrollTo: wrapAndBindImpl(instance, scrollTo),
   };
 
   Object.assign(instance, api);
   return instance;
 }
+
+// This implementation is sourced from testing-library/user-event
+// https://github.com/testing-library/user-event/blob/7a305dee9ab833d6f338d567fc2e862b4838b76a/src/setup/setup.ts#L121
+function wrapAndBindImpl<Impl extends (...args: any) => Promise<any>>(
```
mdjastrzebski, line 161: Seems like we could move it to utils next to `asyncWrapper`

src/user-event/utils/asyncWrapper.ts
```diff
@@ -0,0 +1,35 @@
...
+    }
+  }
+
+  if (!checkReactVersionAtLeast(16, 9)) {
+    return callback();
+  }
+
+  let result: T;
+
+  await act(async () => {
```
mdjastrzebski, line 29: So, we only wrap in `act` from React 16.9 to 17.x. Pls add code comment clarifing this

src/waitFor.ts
```diff
@@ -1,14 +1,13 @@
...
-import { flushMicroTasks, flushMicroTasksLegacy } from './flush-micro-tasks';
+import { flushMicroTasks } from './flush-micro-tasks';
 import { ErrorWithStack, copyStackTrace } from './helpers/errors';
 import {
   setTimeout,
   clearTimeout,
   jestFakeTimersAreEnabled,
 } from './helpers/timers';
-import { checkReactVersionAtLeast } from './react-versions';
+import { asyncWrapper } from './user-event/utils/asyncWrapper';
```
mdjastrzebski, line 10: If it's RNTL-wide helper, move it to general `utils` from `user-event/utils`

src/user-event/utils/asyncWrapper.ts
```diff
@@ -0,0 +1,35 @@
...
+  callback: () => Promise<T>
+): Promise<T> => {
+  if (checkReactVersionAtLeast(18, 0)) {
+    const previousActEnvironment = getIsReactActEnvironment();
+    setReactActEnvironment(false);
+
+    try {
+      const result = await callback();
+      // Flush the microtask queue before restoring the `act` environment
+      await flushMicroTasksLegacy();
```
mdjastrzebski, line 16: RTL has addtionally jest advance timers call: https://github.com/testing-library/react-testing-library/blob/c04b8f006c5a683d05c460c8ee1e2248d6f74350/src/pure.js#L49

Do you think we should add this as well?

mdjastrzebski:
@pierrezimmermannbam thanks for investigating that. I like that new implementation got in line with RTL,s one. I've left some minor comments to be addressed before merging.

NB: From reading the code it seems like we are only focusing on supporting React 18, with React 17 logging some errors in the console, so it would be worth considering dropping official React 17 support in RNTL v13 #1505, since we aren't actually supporting it.

experiments-rtl/src/app/__tests__/click.test.tsx
pierrezimmermannbam: I deleted the test, it was just an experiment I was conducting and I forgot to remove it

src/user-event/__tests__/act.test.tsx
pierrezimmermannbam: I added a comment explaining what happens here

src/user-event/setup/setup.ts
pierrezimmermannbam: Actually I moved the `asyncWrapper` cause it's used also by `waitFor` so I don't think it should be in the `userEvent` folder. I don't see `wrapAndBindImpl` being used elsewhere as it's directly tied to user event so I think it's in the right place

src/user-event/utils/asyncWrapper.ts
pierrezimmermannbam: Not sure at all about that, I have no clue what it does. My thought process here was that we already had pretty much our own version of `asyncWrapper` that we use for `waitFor` and it seems to be working pretty well so I figured we might as well stick to that version. Maybe it could be worth investigating what difference it could make to get closer to RTL's implementation but I'd rather do that in another PR

pierrezimmermannbam:
@mdjastrzebski thanks for the review! I addressed your comments. I agree with you on React 17 support, there are some tests breaking and some errors logged as well on the main branch. And this PR also makes things a bit worse because it introduces overlapping act calls. We could fix that however by removing `act` in `dispatchEvent` for versions < 18, but at this point I'm not even sure it'd be worth it, wdyt?

pierrezimmermannbam:
@mdjastrzebski could you please review when you have the time? It would be nice to have this pr in a release soon because this is currently a big blocker for user event

mdjastrzebski:
@pierrezimmermannbam Thanks for pinging me, I've been pretty busy lately, but I'll try to review this in the following days.

mdjastrzebski:
(rebased on main)

mdjastrzebski (approved):
Looks good to me!

mdjastrzebski:
This PR has been released [v12.3.1](https://github.com/callstack/react-native-testing-library/releases/tag/v12.3.1) 🚀

PR #1492, opened 2023-09-07
refactor: use react jsx transform

pierrezimmermannbam:
### Summary

Use jsx transform to get rid of all React imports

### Test plan

Tests pass so the babel config should work fine and the `yarn build` command still runs successfully

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,9 +1,10 @@
 module.exports = {
   presets: [
     '@babel/preset-typescript',
-    '@babel/preset-react',
+    ['@babel/preset-react', { runtime: 'automatic' }],
     [
       '@babel/preset-env',
+
       {
         targets: {
           node: '14',
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,7 @@
     "react-native/no-inline-styles": "off",
     "react-native-a11y/has-valid-accessibility-descriptors": "off",
     "react-native-a11y/has-valid-accessibility-ignores-invert-colors": 0,
-    "react-native-a11y/has-valid-accessibility-value": "off"
+    "react-native-a11y/has-valid-accessibility-value": "off",
+    "react/react-in-jsx-scope": "off"
   }
 }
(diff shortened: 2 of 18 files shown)


mdjastrzebski:
@pierrezimmermannbam does the change in this PR affects also generated code? Most of the changes files are test files, but how about the `build` output? Could you generate `build` output with and without this PR and see what is the difference there? I wonder if that could affect the users of RNTL in any way.

babel.config.js
```diff
@@ -1,9 +1,10 @@
 module.exports = {
   presets: [
     '@babel/preset-typescript',
-    '@babel/preset-react',
```
mdjastrzebski, line 4: I wonder if this change will affect the output generated in the `build` folder and hence what's get packaged into NPM package. How would that affect our users.

mdjastrzebski: We are shipping some minimal JSX e.g. in `detectHostComponents`.

pierrezimmermannbam:
@mdjastrzebski it does affect the generated code. Compiled files now use `react/jsx-runtime` instead of `React` so it would have an impact on end users. It would be a breaking change because jsx transform was introduced with react 17 so we'd have to update the peer deps. For users that have react >= 17 it would work regardless of if they have this transform enabled in their babel config or not. 

So I guess the question is do we want to drop React 16 support? I'm not too sure on this. I would assume most users have at least version 17. Also we're not running tests anymore on React 16 so we can't be sure RNTL still works fine with it and that would be the main reason to drop react 16 support imo. 

On the other hand this change doesn't bring much value to the users. It's a small improvement for the DX in RNTL but that's pretty much it so it's something that we should do if we drop react 16 support but it's not really the main reason to do so

mdjastrzebski:
@pierrezimmermannbam I think we could do this in RNTL v13. I'm not event sure we support React 16 anymore due to lack of any testing, so RNTL v13 should give clean expectation in this area, perhaps our guideline could be to support the same versions as RN itself: https://reactnative.dev/versions

mdjastrzebski:
Closing for now, can be brought back with v13. #1505

Issue #1485, opened 2023-09-04
feature: implement `toHaveAccessibleValue` Jest matcher


mdjastrzebski:
## Describe the Feature

[`toHaveAccessibleName`](https://github.com/testing-library/jest-dom#tohaveaccessiblename) is a Jest DOM matcher asserting that given element has certain accessible name: text content or a11y label.

RNTL already contains some code working with accessible name, namely `*ByRole` query accepts an optional [`name` param](https://github.com/callstack/react-native-testing-library/blob/a9f32a69e46bedc9413dd0bdf74637f7a3b2b579/src/queries/role.ts#L37). This matcher should compare expect name parameter with all of these:
1. text content of element referenced by `accessibilityLabelledBy` / `aria-labelledby` prop
2. `accessbilityLabel` / `aria-label` prop
3. text content of given element

These should be evaluated in that particular order.

## Possible Implementations
Proposed API:

```ts
export function toHaveAccessibleName(
  this: jest.MatcherContext,
  element: ReactTestInstance,
  expectedName?: TextMatch,
) 
```

The matcher should:
1) Validate that it is invoked on a host element
2) When `expectedName` param is provided check if it matches element's accessible name
3) When `expectedName` is not provided check if element has accessible name

Each matcher should have a fairly comprehensive test suite.

## Related Issues
#1454

tarunrajput:
@mdjastrzebski, I'd like to implement this

mdjastrzebski:
@tarunrajput go ahead, it's yours now!

mdjastrzebski:
Resolved in #1496!

mdjastrzebski:
🎉 Release in version: [v12.4.0](https://github.com/callstack/react-native-testing-library/releases/tag/v12.4.0)

PR #1496, opened 2023-09-11
Implement toHaveAccessibilityValue matcher

tarunrajput:
### Summary

#1485 - Implement [`toHaveAccessibilityValue`](https://github.com/testing-library/jest-native/blob/main/src/to-have-accessibility-value.ts) matcher

### Test plan
Add unit tests and ensure all test passes


src/matchers/__tests__/to-have-accessibility-value.tsx
```diff
@@ -0,0 +1,149 @@
...
+      />
+      <View
+        testID="text"
+        accessibilityValue={{ text: 'accessibility value' }}
+      />
+    </>
+  );
+}
+
+test('toHaveAccessibilityValue() on matching accessibility value', () => {
```
pierrezimmermannbam, line 29: Imo these tests are not very readable for two reasons: 

- the first one is because of the `renderViewsWithAccessibilityValue` util. It reduces code duplication but code duplication in tests is not a big deal and here to understand every test it is necessary to look its definition to understand the test which is not great
- the second is the use of variables in the test, e.g. `min` or `minMax`, for the same reason. The variable names actually have less meaning than how they were queried so I'd rather use queries directly.

The test cases could be organized differently to avoid rendering so many components in each tests. You could basically write the same tests than in the [tests or the role query ](https://github.com/callstack/react-native-testing-library/blob/main/src/queries/__tests__/role-value.test.tsx). Also what is missing in those tests and that's present in the role query tests is the support for `aria props`

pierrezimmermannbam (changes requested):
@tarunrajput  Thanks for the pr! Implementation looks good, I just have some feedback on the tests that needs to be addressed  before we merge this

src/matchers/__tests__/to-have-accessibility-value.tsx
mdjastrzebski: [nit] Yeah, I understand that the tests resemble what Jest Native had, and I think that @pierrezimmermannbam has a point here, that by grouping the scenarios into more smaller tests the tests would be more readable.

mdjastrzebski (approved):
Looks good! Let's merge 🔀 this!

mdjastrzebski:

PR #1493, opened 2023-09-07
feat: `toBeBusy()` matcher

hduprat:
### Summary

Implement new matcher `toBeBusy()`

### Test plan

`yarn test`

PR #1497, opened 2023-09-13
feat: `toBeExpanded` matcher

siepra:
### Summary

Part of Jest matchers migration
Following the example of https://github.com/callstack/react-native-testing-library/pull/1493

### Test plan

`yarn test`

### DIFF
--- a/src/helpers/accessiblity.ts
+++ b/src/helpers/accessiblity.ts
@@ -213,11 +213,11 @@ export function isElementBusy(
   return ariaBusy ?? accessibilityState?.busy ?? false;
 }
 
-export function isElementSelected(
+export function isElementCollapsed(
   element: ReactTestInstance
-): NonNullable<AccessibilityState['selected']> {
-  const { accessibilityState, 'aria-selected': ariaSelected } = element.props;
-  return ariaSelected ?? accessibilityState?.selected ?? false;
+): NonNullable<AccessibilityState['expanded']> {
+  const { accessibilityState, 'aria-expanded': ariaExpanded } = element.props;
+  return (ariaExpanded ?? accessibilityState?.expanded) === false;
 }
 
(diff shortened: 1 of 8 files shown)


mdjastrzebski:
@siepra thanks for submitting this PR. During our experiments with iOS/Android RN runtime we've noticed that `expanded: false` state is behaving different than omitting the `expanded` entry. Detailed behavior here: https://github.com/callstack/react-native-testing-library/wiki/Accessibility:-State#expanded-state

Basically, RN treats explicit `expanded: false` as "collapsed" state, where as lack of `expanded` state means that this is neither expanded nor no collapsed. Therefore, it would make sense to introduce, in this PR, additional matcher of `toBeCollapsed` which would check `expanded === false` condition. Pls add `isElementCollapsed` as well + relevant tests.

src/helpers/accessiblity.ts
```diff
@@ -226,3 +226,14 @@ export function isElementExpanded(
   const { accessibilityState, 'aria-expanded': ariaExpanded } = element.props;
   return ariaExpanded ?? accessibilityState?.expanded ?? false;
 }
+
+export function isElementCollapsed(
+  element: ReactTestInstance
+): NonNullable<AccessibilityState['expanded']> {
+  const { accessibilityState, 'aria-expanded': ariaExpanded } = element.props;
+  return ariaExpanded !== undefined
```
mdjastrzebski, line 234: Why not `return (ariaExpanded ?? accessibilityState?.expanded) === false`

siepra: It's the same, I just wanted to make sure the syntax will be correct for the default case. Changing it now

mdjastrzebski (approved):
Looks good! ✅

mdjastrzebski:

PR #1500, opened 2023-09-18
fix: string validation type checking

mdjastrzebski:


mdjastrzebski:
Closing, as this is not an issue with our code, but rather recent changes to `@types/react-test-renderer` are incorrect which I'verified by running check locally and also analysis the React Test Renderer source code. (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/65855)

PR #1504, opened 2023-09-20
fix: disable React 17 tests

mdjastrzebski:
### Summary

Disable React 17 tests on CI, as they were always failing due to snapshot miss match as well of some unsupported features like the new ARIA props. On the other hand the way the test script has been written quietly consumed this failure, so that React 17 tests always indicated success.

I've tried to disable just this tests on the CI but could not overcome following error (local & CI):
```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  MESSAGEPORT

      1 | // This file and the act() implementation is sourced from react-testing-library
      2 | // https://github.com/testing-library/react-testing-library/blob/c80809a956b0b9f3289c4a6fa8b5e8cc72d6ef6d/src/act-compat.js
    > 3 | import { act as reactTestRendererAct } from 'react-test-renderer';
        | ^
      4 | import { checkReactVersionAtLeast } from './react-versions';
      5 |
      6 | type ReactAct = typeof reactTestRendererAct;

      at node_modules/scheduler/cjs/scheduler.development.js:178:17
      at Object.<anonymous> (node_modules/scheduler/cjs/scheduler.development.js:645:5)
      at Object.<anonymous> (node_modules/scheduler/index.js:6:20)
      at node_modules/react-test-renderer/cjs/react-test-renderer.development.js:19:19
      at Object.<anonymous> (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:17382:5)
      at Object.<anonymous> (node_modules/react-test-renderer/index.js:6:20)
      at Object.require (src/act.ts:3:1)
      at Object.require (src/pure.ts:1:1)
      at Object.require (jest-setup.ts:1:1)
```

React 17 tests could still be run locally using `yarn test:react-17` script.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,7 +61,7 @@ jobs:
   test:
     needs: [install-cache-deps]
     runs-on: ubuntu-latest
-    name: Test React 18
+    name: Test
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


pierrezimmermannbam (approved):
Since they didn't work it's better that way but it means that we're only testing against react 18 now, I wonder if it means we should update the peer deps accordingly

mdjastrzebski:
@pierrezimmermannbam when running tests locally they pass except "open handle error" that makes them return non-zero exit code. So I think we can say we support React 17/ older versions of RN for now. That being said I think we should offically drop support for older React/RN with RNTL v13 (see #1505)

PR #1523, opened 2023-11-13
docs: expose jest matchers publically (docs, examples, etc)

mdjastrzebski:
### Summary

All of the steps to expose built-in Jest matchers:
- [x] Update docs (references to Jest Native)
- [x] Update error messages / deprecations
- [x] Update example apps  

### Test plan

All checks should pass, all example apps checks should pass as well.


website/docs/GettingStarted.md
```diff
@@ -40,27 +40,10 @@ In order to properly use helpers for async tests (`findBy` queries and `waitFor`
 
 ### Additional Jest matchers
 
-In order to use additional React Native-specific Jest matchers from [@testing-library/jest-native](https://github.com/testing-library/jest-native) package add it to your project:
+In order to use additional React Native-specific Jest matchers add the following line to your `jest-setup.ts` file:
```
pierrezimmermannbam, line 43: It needs to be imported in a setup after env file, I think it may not be very clear here, we could say add the following line to your `jest-setup-after-env.ts`?

website/docs/JestMatchers.md
```diff
@@ -7,15 +7,34 @@ title: Jest Matchers
...
+This guide describes built-in Jest matchers, we recommend using these matchers as they provide readable tests, accessibility support, and a better developer experience.
 
 import TOCInline from '@theme/TOCInline';
 
 <TOCInline toc={toc} />
 
-## Element Existence
+## Setup
+
+You can use the built-in matchers by adding the following line to your `jest-setup.ts` file:
```
pierrezimmermannbam, line 18: Same remark here

pierrezimmermannbam (approved):

website/docs/GettingStarted.md
mdjastrzebski: Good point, I think usually projects should do setup in setupFilesAfterEnv, so I wil clarify this.

PR #1531, opened 2023-11-21
docs: rewrite How should I query? guide

mdjastrzebski:
### Summary

Rewrite How should I query? guide

### Test plan

Proof-read the docs, check the internal and external links.

### DIFF
--- a/website/docs/HowShouldIQuery.md
+++ b/website/docs/HowShouldIQuery.md
@@ -63,7 +63,7 @@ The query predicate describes how you decide whether to match the given element.
 
 Choosing the proper query predicate helps better express the test's intent and make the tests resemble how users interact with your code (components, screens, etc.) as much as possible following our [Guiding Principles](https://testing-library.com/docs/guiding-principles). Additionally, most predicates promote the usage of proper accessibility props, which add a semantic layer on top of an element tree composed primarily of [`View`](https://reactnative.dev/docs/view) elements.
 
-It is recommended to use queries in the following order of priority:
+It is recommended to use query predicates in the following order of priority:
 
 ### 1. By Role query {#by-role-query}
 
--- a/website/docs/HowShouldIQuery.md
+++ b/website/docs/HowShouldIQuery.md
@@ -49,15 +49,15 @@ Do not use `queryAllBy*` as it provides no assertions on the number of matching 
 
 The query predicate describes how you decide whether to match the given element.
 
-| Predicate                                               | Supported elements                                    | Inspected props                                                                             |
-| ------------------------------------------------------- | ----------------------------------------------------- | ------------------------------------------------------------------------------------------- |
-| [`*ByRole`](api-queries#by-role)                        | all host elements                                     | `role`, `accessibilityRole`,<br /> optional: accessible name, accessibility state and value |
-| [`*ByLabelText`](api-queries#by-label-text)             | all host elements                                     | `aria-label`, `aria-labelledby`,<br /> `accessibilityLabel`, `accessibilityLabelledBy`      |
-| [`*ByDisplayValue`](api-queries#by-display-value)       | [`TextInput`](https://reactnative.dev/docs/textinput) | `value`, `defaultValue`                                                                     |
-| [`*ByPlaceholderText`](api-queries#by-placeholder-text) | [`TextInput`](https://reactnative.dev/docs/textinput) | `placeholder`                                                                               |
-| [`*ByText`](api-queries#by-text)                        | [`Text`](https://reactnative.dev/docs/text)           | `children` (text content)                                                                   |
-| [`*ByHintText`](api-queries#by-hint-text)               | all host elements                                     | `accessibilityHint`                                                                         |
-| [`*ByTestId`](api-queries#by-test-id)                   | all host elements                                     | `testID`                                                                                    |
+| Predicate                                               | Supported elements | Inspected props                                                                             |
+| ------------------------------------------------------- | ------------------ | ------------------------------------------------------------------------------------------- |
+| [`*ByRole`](api-queries#by-role)                        | all host elements  | `role`, `accessibilityRole`,<br /> optional: accessible name, accessibility state and value |
+| [`*ByLabelText`](api-queries#by-label-text)             | all host elements  | `aria-label`, `aria-labelledby`,<br /> `accessibilityLabel`, `accessibilityLabelledBy`      |
+| [`*ByDisplayValue`](api-queries#by-display-value)       | `TextInput`        | `value`, `defaultValue`                                                                     |
+| [`*ByPlaceholderText`](api-queries#by-placeholder-text) | `TextInput`        | `placeholder`                                                                               |
+| [`*ByText`](api-queries#by-text)                        | `Text`             | `children` (text content)                                                                   |
+| [`*ByHintText`](api-queries#by-hint-text)               | all host elements  | `accessibilityHint`                                                                         |
+| [`*ByTestId`](api-queries#by-test-id)                   | all host elements  | `testID`                                                                                    |
 
 ### Idiomatic query predicates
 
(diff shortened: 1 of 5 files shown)


MattAgn:
This is much more more explicit and clear than before! 🤩

website/docs/HowShouldIQuery.md
```diff
@@ -3,19 +3,143 @@ id: how-should-i-query
...
+| [`*ByHintText`](api-queries#by-hint-text)               | all host elements  | `accessibilityHint`                                                                         |
+| [`*ByTestId`](api-queries#by-test-id)                   | all host elements  | `testID`                                                                                    |
+
+### Idiomatic query predicates
+
+Choosing the proper query predicate helps better express the test's intent and make the tests resemble how users interact with your code (components, screens, etc.) as much as possible following our [Guiding Principles](https://testing-library.com/docs/guiding-principles).
+
+Additionally, most predicates promote the usage of proper accessibility props, which add a semantic layer on top of an element tree composed primarily of [`View`](https://reactnative.dev/docs/view) elements.
+
+### 1. By Role query {#by-role-query}
```
MattAgn, line 68: i'm curious, what's the use of `{#by-role-query}`? it's the first time i see this

website/docs/HowShouldIQuery.md
```diff
@@ -3,19 +3,143 @@ id: how-should-i-query
...
+Using idiomatic query variants helps better express your test's intent and expectations about the number of matching elements. Using other query variants might work but could make it harder to reason about the test.
+
+Here are general guidelines for picking idiomatic query variants:
+
+1. Use `getBy*` in the most common case when you expect a **single matching element**. Use other queries only in more specific cases.
+2. Use `findBy*` when an element is not yet in the element tree, but you expect it to be there as a **result of some asynchronous action**.
+3. Use `getAllBy*` (and `findAllBy*` for async) if you expect **more than one matching element**, e.g. in a list.
+4. Use `queryBy*` variant only when element **should not exist** to use it together with e.g. [`not.toBeOnTheScreen()`](jest-matchers#tobeonthescreen) matcher.
+
+Do not use `queryAllBy*` as it provides no assertions on the number of matching elements.
```
MattAgn, line 46: I find it a bit weird that we tell people not to use `queryAllBy` and at the same time give them the ability to use it.
It think we could either: 
- say there might be exceptions and give an example of such exception
- or simply remove the query from RNTL

website/docs/HowShouldIQuery.md
mdjastrzebski: This is definition of anchor for http links in Docosaurus

website/docs/HowShouldIQuery.md
mdjastrzebski: Some thoughts on reasons why to keep it:
* RTL compatibility
* support legacy code
* building block for custom wrappers, etc

website/docs/HowShouldIQuery.md
```diff
@@ -3,19 +3,143 @@ id: how-should-i-query
...
+| [`queryBy*`](api-queries#query-by)        | Zero or one matching element  | <code>ReactTestInstance &#124; null</code> | No        |
+| [`queryAllBy*`](api-queries#query-all-by) | No assertion                  | `Array<ReactTestInstance>`                 | No        |
+| [`findBy*`](api-queries#find-by)          | Exactly one matching element  | `Promise<ReactTestInstance>`               | Yes       |
+| [`findAllBy*`](api-queries#find-all-by)   | At least one matching element | `Promise<Array<ReactTestInstance>>`        | Yes       |
+
+Queries work as implicit assertions on the number of matching elements and will throw an error when the assertion fails.
+
+### Idiomatic query variants
+
+Using idiomatic query variants helps better express your test's intent and expectations about the number of matching elements. Using other query variants might work but could make it harder to reason about the test.
```
pierrezimmermannbam, line 37: It also improves the error messages because it's the query itself that throws and not the assertion around

website/docs/HowShouldIQuery.md
```diff
@@ -3,19 +3,143 @@ id: how-should-i-query
...
+
+### Idiomatic query predicates
+
+Choosing the proper query predicate helps better express the test's intent and make the tests resemble how users interact with your code (components, screens, etc.) as much as possible following our [Guiding Principles](https://testing-library.com/docs/guiding-principles).
+
+Additionally, most predicates promote the usage of proper accessibility props, which add a semantic layer on top of an element tree composed primarily of [`View`](https://reactnative.dev/docs/view) elements.
+
+### 1. By Role query {#by-role-query}
+
+The first and most versatile predicate is [`*ByRole`](api-queries#by-role), which starts with the semantic role of the element and can be further narrowed down with additional options. React Native has two role systems, the web/ARIA-compatible one based on [`role`](https://reactnative.dev/docs/accessibility#role) prop and the traditional one based on [`accessibilityRole`](https://reactnative.dev/docs/accessibility#accessibilityrole) prop, you can use either of these.
```
pierrezimmermannbam, line 70: Imo there is a part of this doc that conflicts with the `Queries` section of the doc. I believe that how queries work and their API should be detailed there and that doc should be referenced here. I think this specific doc should focus on when to use which query and what are the reasons for it

website/docs/HowShouldIQuery.md
```diff
@@ -3,19 +3,143 @@ id: how-should-i-query
...
+- [`toBeBusy()`](jest-matchers#tobebusy)
+- [`toHaveAccessibilityValue()`](jest-matchers#tohaveaccessibilityvalue)
+
+### 2. Text input queries {#text-input-queries}
+
+Querying [`TextInput`](https://reactnative.dev/docs/textinput) elements presents a unique challenge as there is no separate role for `TextInput` elements. There is a `searchbox`/`search` role, which can be assigned to `TextInput`, but it should be only used in the context of search inputs, leaving other text inputs without a role to query with.
+
+Therefore, you can use the following queries to find relevant text inputs:
+
+1. [`*ByLabelText`](api-queries#by-label-text) - will match the accessibility label of the element. This query will match any host elements, including `TextInput` elements.
```
pierrezimmermannbam, line 127: Is it a good practice to use an a11y label on a text input? You'll need to change it depending on the value it's not very practical. I personally don't and use `ByDisplayValue ` and `ByPlaceholderText` almost all of the time. I also like those more because they're specific to `TextInput` so they make tests clearer imo

pierrezimmermannbam (changes requested):
Huge improvement on the previous docs and very nice to finally see `role` queries promoted to # 1 !

website/docs/HowShouldIQuery.md
mdjastrzebski: Good question. I think that every TextInput needs to have a label identifying its purpose. For visual users it might be a adjacent `Text` label, however for users using a11y this might not so evident, and hence setting a11y label would benefit them, as well as improve the testing. 
(Note to self: I need to verify whether `placeholder` prop is pronounced under screen reader.)

Regarding `*ByPlaceholder` vs `*ByDisplayValue`, I would prefer using placeholder query as it identifies the input by it's function, regardless of current value which might change.

website/docs/HowShouldIQuery.md
mdjastrzebski: I see what you mean, I do not intend to duplicate the detailed API doc for `*ByRoles` query, and at the same time highlight popular query techniques, so introducing some common roles and mentioning use of `name` param as an essential option for this query.

website/docs/HowShouldIQuery.md
MattAgn: good for me!

website/docs/HowShouldIQuery.md
```diff
@@ -3,19 +3,142 @@ id: how-should-i-query
...
+| [`*ByHintText`](api-queries#by-hint-text)               | all host elements  | `accessibilityHint`                                                                         |
+| [`*ByTestId`](api-queries#by-test-id)                   | all host elements  | `testID`                                                                                    |
+
+### Idiomatic query predicates
+
+Choosing the proper query predicate helps better express the test's intent and make the tests resemble how users interact with your code (components, screens, etc.) as much as possible following our [Guiding Principles](https://testing-library.com/docs/guiding-principles).
+
+Additionally, most predicates promote the usage of proper accessibility props, which add a semantic layer on top of an element tree composed primarily of [`View`](https://reactnative.dev/docs/view) elements.
+
+### 1. By Role query {#by-role-query}
```
MattAgn, line 68: I'm not sure if this is needed, but i would maybe add a sentence explaining that the list of query predicates are ordered by how much people should use them

Issue #1534, opened 2023-11-28
Typescript fails in `extend-expect.d.ts`


erheron:
## Describe the bug
We use Typescript in our project. We do not have `skipLibCheck` enabled. 
After installation of RNTL v 12.4.0 and following [RNTL Docs on installation](https://callstack.github.io/react-native-testing-library/docs/getting-started#installation) it turns out that Typescript fails in RNTL's `node_modules/@testing-library/react-native/src/matchers/extend-expect.d.ts` file. 

## Expected behavior
Typescript does not fail

## Steps to Reproduce

Here's a reproduction repo: TBD

## Screenshots

<img width="1245" alt="image" src="https://github.com/callstack/react-native-testing-library/assets/36896640/6a24c75a-c4e2-4b11-80b6-3b768769552c">

## Versions

```bash
npmPackages:
    @testing-library/react-native: ^12.4.0 => 12.4.0
    react: 18.2.0 => 18.2.0
    react-native: ^0.72.7 => 0.72.7
    react-test-renderer: 18.2.0 => 18.2.0
```

erheron:
Weirdly enough, Typescript error is reproducible in our original project, but not in the smaller repro I shared. I'll take another look soon and maybe try to update the repro. But TSC error on `toHaveStyle` is concerning too 🤔

mdjastrzebski:
@erheron try having separate `jest-setup.ts` file as in our [example](https://github.com/callstack/react-native-testing-library/blob/01c770d3421eaaed28bc8a13d179d23a96ea1c50/examples/basic/jest-setup.ts#L4).

erheron:
Yeah, clone the repro repo, check out [this commit](https://github.com/erheron/rntl-1534-repro/commit/fa81d120cd3354526ceeb272d2f0f695dfcfccf1) and run `yarn tsc`. What you should see is exactly this:

```bash
node_modules/@testing-library/react-native/src/matchers/extend-expect.d.ts:3:50 - error TS2307: Cannot find module '../matches' or its corresponding type declarations.

3 import type { TextMatch, TextMatchOptions } from '../matches';
                                                   ~~~~~~~~~~~~

node_modules/@testing-library/react-native/src/matchers/extend-expect.d.ts:4:48 - error TS2307: Cannot find module '../helpers/matchers/match-accessibility-value' or its corresponding type declarations.

4 import type { AccessibilityValueMatcher } from '../helpers/matchers/match-accessibility-value';
[...]

PR #1537, opened 2023-11-29
fix: matcher Jest extensions

mdjastrzebski:
### Summary

Clean up TS issues reported in #1534 without duplicating type declarations.

This solution is modeled on approach used in Jest Expect package: https://github.com/jestjs/jest/blob/main/packages/jest-expect/src/types.ts

### DIFF
--- a/src/matchers/types.ts
+++ b/src/matchers/types.ts
@@ -5,7 +5,11 @@ import { TextMatch, TextMatchOptions } from '../matches';
 import { Style } from './to-have-style';
 
 export interface JestNativeMatchers<R> {
+  /**
+   * Assert whether an element is present in the element tree or not.
+   */
   toBeOnTheScreen(): R;
+
   toBeChecked(): R;
   toBeCollapsed(): R;
   toBeDisabled(): R;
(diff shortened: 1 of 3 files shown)


src/matchers/types.ts
```diff
@@ -31,26 +38,8 @@ declare global {
 }
 
 // Explicit `@jest/globals` `expect` matchers.
+// @ts-ignore
```
erheron, line 41: Wouldn't this effectively force all your users to set `skipLibCheck: true` in tsconfig? 😔

mdjastrzebski: Do you have ideas how to resolve any other way?
image

mdjastrzebski: idk when testing against your repro repo the mentioned `@ts-ignore` seems to work fine. Seems like it might just be suppring check for `@jest/expect` existence which does not prevent it from using in test app with `@jest/expect`.

Note: In order to trigger the `@jest/globals` module typing you need to use `expect` from `@jest/globals` and not a global one:

image

Which BTW triggers a lot other TS errors when invoking `tsc` but not from RNTL itself, but things like:
* node_modules/@types/node/buffer.d.ts
* node_modules/react-native/types/modules/globals.d.ts

PR #1545, opened 2023-12-15
Disable prettier error alert

tiago-edu:
Prettier alerts disabled. Focusing on disabling unnecessary errors.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "@callstack",
   "rules": {
+    "prettier/prettier": 0, 
     "flowtype/no-weak-types": 0,
     "react-native/no-raw-text": 0,
     "no-console": 1,


mdjastrzebski:
@tiago-edu What is the motivation behind this PR?

tiago-edu:
> @tiago-edu What is the motivation behind this PR?

For beginner programmers, errors can often seem 'scary'. eslint, depending on each machine, changes the errors present based on the user's presets. I think it's viable to disable it, so that you don't have unnecessary error alerts. Furthermore, because it is just a line of code, I believe it will bring much more benefits than harm.

mdjastrzebski:
Closing as we use ESLint + Prettier integration to make sure that submitted PR-s are coherent with code formatting of the rest of this code. Disabling it would create constant re-formatting issues between PRs.

PR #1603, opened 2024-05-03
[v13] chore: remove deprecated stuff

mdjastrzebski:
### Summary

Remove deprecated stuff:
* `*ByAccessibilityState` queries
* `*ByAccessibilityValue` queries
* use modern `flushMicroTasks` instead of `flushMicroTasksLegacy`

### Test plan

All checks should pass.


pierrezimmermannbam:
Should the target branch be something other than main though? I thought we'd work on a separate branch for react 19 or do we still want to remove those on main?

mdjastrzebski:
@pierrezimmermannbam you're right, it should be `v13`. Fixed it.

PR #1608, opened 2024-05-06
chore: remove stale examples

mdjastrzebski:
### Summary

Remove stale examples for Redux
Remove stale docs for Redux and React Navigation


friederbluemle:
I think the Redux examples could still be useful, even if "stale". Is there a chance to get the outdated examples updated instead of removed altogether?

mdjastrzebski:
I definitely would like to have a proper Redux example (as well as examples for other state management solutions). However, the current "example" goes against our recommended testing best practices 🤷‍♂️

@friederbluemle would you be interested in contributing a small Redux setup example for RNTL that could be added to this repo? I offer my guidance and code review.

Issue #1611, opened 2024-05-20
unstable_validateStringsRenderedWithinText immediate setState in useEffect 


precious-void:
## Describe the bug
When immediately calling setState in useEffect and the component is rendered with `unstable_validateStringsRenderedWithinText` parameter, the following error is thrown:
```
    `render` method has not been called

       5 |
       6 | const notImplemented = () => {
    >  7 |   throw new Error(SCREEN_ERROR);
         |         ^
       8 | };
       9 |
      10 | const notImplementedDebug = () => {

      at Object.notImplemented (src/screen.ts:7:9)
      at toJSON (src/render.tsx:71:48)
      at commitLayoutEffectOnFiber (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:13282:15)
      at commitLayoutMountEffects_complete (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:14330:9)
      at commitLayoutEffects_begin (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:14316:7)
      at commitLayoutEffects (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:14301:3)
      at commitRootImpl (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:16062:5)
      at commitRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:15933:5)
      at performSyncWorkOnRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:15454:3)
      at flushSyncCallbacks (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2597:22)
      at flushPassiveEffectsImpl (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:16261:3)
      at flushPassiveEffects (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:16197:14)
      at node_modules/react-test-renderer/cjs/react-test-renderer.development.js:16012:9
      at flushActQueue (node_modules/react/cjs/react.development.js:2667:24)
      at act (node_modules/react/cjs/react.development.js:2521:11)
      at actImplementation (src/act.ts:30:25)
      at renderWithAct (src/render-act.ts:12:11)
      at renderWithStringValidation (src/render.tsx:81:33)
      at renderWithStringValidation (src/render.tsx:52:12)
      at renderInternal (src/render.tsx:29:10)
```

## Expected behavior

[...]

PR #1612, opened 2024-05-20
fix: immediate setState in useEffect for unstable_validateStringsRenderedWithinText

precious-void:
### Summary

Referencing: #1611 
Fixed by checking for the `renderer` to be initialized for the first update received from the Profiler `onRender`

### Test plan

Implemented test for that case, that was failing before this implemented fix.


mdjastrzebski (approved):

mdjastrzebski:
This PR has been released with [v12.5.1](https://github.com/callstack/react-native-testing-library/releases/tag/v12.5.1) 🎉

PR #1641, opened 2024-08-07
fix: Respect maxLength prop in type() function

### LINKED ISSUE #1239: Max length prop of TextInput is ignored
## Describe the bug

It is possible to change the value of a TextInput that has a maxLength prop. The following test fails for instance :

```tsx
test('it should not allow to write a text longer than maxlength in a text input', () => {
  const TestComponent = () => {
    const [value, setValue] = useState('');

    return (
      <RNTextInput placeholder="placeholder" maxLength={1} onChangeText={setValue} value={value} />
    );
  };

  render(<TestComponent />);

  fireEvent.changeText(screen.getByPlaceholderText('placeholder'), 'text');

  screen.debug();
  expect(screen.queryByDisplayValue('text')).toBeFalsy();

});

```

## Expected behavior

I'd expect that only the substring of length equal to maxlength is set as value.

Regarding the implementation I'm not sure though if this should be done by fireEvent because it just calls props and we'd have to check what the event is and have special checks in that case so maybe it's something that should be done once we have an implementation of userEvent.type. So actually I'm saying this is not really a bug but rather a use case that userEvent.type  should support

joshuayoes:
### Summary
This PR modifies the `type()` function to respect the `maxLength` prop of the input element. If the current text length exceeds the `maxLength` value, typing events will not be emitted. This ensures that the input value does not exceed the specified maximum length.

Resolves #1239

### Test plan
1. Create an input element with a `maxLength` prop.
2. Use the `type()` function to input text that exceeds the `maxLength` value.
3. Verify that the input value does not exceed the specified maximum length.
4. Check that no typing events are emitted once the `maxLength` is reached.

Additionally, see that `yarn test` passes with an additional test added to `src/user-event/type/__tests__/type.test.tsx`

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,8 +2,12 @@ coverage:
   precision: 2
   round: down
   range: 70...100
+  status:
+    project:
+      default:
+        threshold: 0.5%  # Allowable coverage drop in percentage points
 
 comment:
   behavior: default
   require_changes: false
-  require_base: no
+  require_base: false
--- a/src/test-utils/events.ts
+++ b/src/test-utils/events.ts
@@ -19,6 +19,10 @@ export function createEventLogger() {
   return { events, logEvent };
 }
 
-export function getEventsName(events: EventEntry[]) {
+export function getEventsNames(events: EventEntry[]) {
   return events.map((event) => event.name);
+}
+
+export function lastEventPayload(events: EventEntry[], name: string) {
+  return events.filter((e) => e.name === name).pop()?.payload;
 }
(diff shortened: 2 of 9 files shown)


mdjastrzebski:
Hi Josh, thanks for submitting this PR. I've done some experiments on iOS, Android and RN Web to check what is the actual behavior of user typing while reaching max chars limit. Here are my findings:
- iOS: emits only `keyPress` event s
- Android: no events emitted
- RN Web: emits only `keyPress` events

Based on this, I'll tweak your PR to match the iOS/RN Web behavior, as RNTL does not differentiate between platform behaviors but tries to "merge" them in some "reasonable" way.

mdjastrzebski:
This PR has been released in [v12.5.3](https://github.com/callstack/react-native-testing-library/releases/tag/v12.5.3) 🎉

joshuayoes:
Thanks for tweaking this and getting it merged @mdjastrzebski!

joshuayoes:
I bumped to the updated version in my project and I was able to remove my local patch!

PR #1653, opened 2024-08-27
feat: simulated native TextInput state

mdjastrzebski:
### Summary

This PR introduces the concept of simulated native view state. The PR supports tracking the value of host `TextInput` elements. 

```
test('sets native state value for unmanaged text inputs', async () => {
  render(<TextInput testID="input" />);

  const user = userEvent.setup();
  const input = screen.getByTestId('input');
  expect(input).toHaveDisplayValue('');

  await user.type(input, 'abc'); // or fireEvent.changeText(input, 'abc')

  // Below assertion now passes
  expect(input).toHaveDisplayValue('abc'); 
});
```

This PR supports setting native state value through `fireEvent.changeText` and  User Event `type()`, `clear()` and `paste()` methods.

When `TextInput` is managed through `value` prop, that value takes precedence over simulated native value.

### Test plan

Add

### DIFF
--- a/src/__tests__/fire-event.test.tsx
+++ b/src/__tests__/fire-event.test.tsx
@@ -140,7 +140,6 @@ test('fireEvent.scroll', () => {
 
 test('fireEvent.changeText', () => {
   const onChangeTextMock = jest.fn();
-  const CHANGE_TEXT = 'content';
 
   render(
     <View>
@@ -148,9 +147,19 @@ test('fireEvent.changeText', () => {
     </View>,
   );
 
-  fireEvent.changeText(screen.getByPlaceholderText('Customer placeholder'), CHANGE_TEXT);
+  const input = screen.getByPlaceholderText('Customer placeholder');
+  fireEvent.changeText(input, 'content');
+  expect(onChangeTextMock).toHaveBeenCalledWith('content');
+});
 
-  expect(onChangeTextMock).toHaveBeenCalledWith(CHANGE_TEXT);
+it('sets native state value for unmanaged text inputs', () => {
+  render(<TextInput testID="input" />);
+
+  const input = screen.getByTestId('input');
+  expect(input).toHaveDisplayValue('');
+
+  fireEvent.changeText(input, 'abc');
+  expect(input).toHaveDisplayValue('abc');
 });
 
 test('custom component with custom event name', () => {
@@ -443,15 +452,5 @@ describe('native events', () => {
 
     fireEvent(screen.getByTestId('test-id'), 'onMomentumScrollEnd');
     expect(onMomentumScrollEndSpy).toHaveBeenCalled();
-  });
-
-  it('sets native state value for unmanaged text inputs', () => {
-    render(<TextInput testID="input" />);
-
-    const input = screen.getByTestId('input');
-    expect(input).toHaveDisplayValue('');
-
-    fireEvent.changeText(input, 'abc');
-    expect(input).toHaveDisplayValue('abc');
   });
 });
--- a/src/__tests__/fire-event.test.tsx
+++ b/src/__tests__/fire-event.test.tsx
@@ -9,6 +9,7 @@ import {
   View,
 } from 'react-native';
 import { fireEvent, render, screen } from '..';
+import '../matchers/extend-expect';
 
 type OnPressComponentProps = {
   onPress: () => void;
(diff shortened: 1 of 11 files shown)


src/__tests__/fire-event.test.tsx
```diff
@@ -443,4 +444,14 @@ describe('native events', () => {
     fireEvent(screen.getByTestId('test-id'), 'onMomentumScrollEnd');
     expect(onMomentumScrollEndSpy).toHaveBeenCalled();
   });
+
+  it('sets native state value for unmanaged text inputs', () => {
```
MattAgn, line 448: [nit] it could be nice to put this test next to the other TextInput test to quickly see the difference

src/user-event/type/type.ts
```diff
@@ -94,6 +95,7 @@ export async function emitTypingEvents(
     return;
   }
 
+  nativeState?.elementValues.set(element, text);
```
MattAgn, line 98: i'm curious, why do you set the native state on each key press and not at the end (after the `  for (const key of keys)`, if i understand correctly, for each key, of the text, we dispatch an event and we set the initial value right ?

MattAgn (approved):

src/__tests__/fire-event.test.tsx
```diff
@@ -9,6 +9,7 @@ import {
   View,
 } from 'react-native';
 import { fireEvent, render, screen } from '..';
+import '../matchers/extend-expect';
```
pierrezimmermannbam, line 12: It should be possible to import that file in a jest setup file so that matchers are available globally no?

src/fire-event.ts
```diff
@@ -143,3 +146,14 @@ fireEvent.scroll = (element: ReactTestInstance, ...data: unknown[]) =>
   fireEvent(element, 'scroll', ...data);
 
 export default fireEvent;
+
+function setNativeStateIfNeeded(element: ReactTestInstance, eventName: string, value: unknown) {
```
pierrezimmermannbam, line 150: Should it be done for fireEvent at all? I think it's a very good idea for userEvent but I think fireEvent should remain as minimal as possible, even though there is the shorthand for `fireEvent.changeText`, making it do more than calling the `onChangeText` nurtures the confusion around the differences between fireEvent and userEvent imo. And here it would not work with fireEvent(element, 'change') which also adds confusion

pierrezimmermannbam (commented):
That's a great idea! I like the implementation as well but I have one concern, when using debug or snapshots, native state would not be visible right? I don't know how it could be visible unless the element props we directly updated. Have you considered this solution?

mdjastrzebski:
@pierrezimmermannbam the native state by it's nature it's not touching props (like `TextInput` `value`), so snapshot won't showcase them. If you have any idea how to augment the snapshots by native state let me know.

src/fire-event.ts
mdjastrzebski: I would keep it, as many existing codebases rely on Fire Event, so restricting this to User Event, would prevent them from accessing the benefits of native state simulation.

I would guess that 90% of Fire Event users use `fireEvent.changeText` to change text instead of building custom event object for `fireEvent(element, "change", { event... })`. If someone complains that we do not support it, then we can add it here.

src/user-event/type/type.ts
mdjastrzebski: If we make the native state accessible through refs (not there yet, but could be). Then the event handler could theoretically query the new state. Otherwise, it's practially zero cost so why not.

src/fire-event.ts
pierrezimmermannbam: Existing codebases indeed rely on `fireEvent` but they do not use this feature so it would not break anything for them. You're right that it would mean they don't have access to that feature but that's only true if they don't start using `userEvent`. I think we should see `fireEvent` as a deprecated API, not really deprecated but only to be used for specific use cases that `userEvent` does not cover and as such I would not add new features to it, otherwise I don't know if it would be clear for users which API they should use and why they should migrate to `userEvent`

pierrezimmermannbam:
> @pierrezimmermannbam the native state by it's nature it's not touching props (like `TextInput` `value`), so snapshot won't showcase them. If you have any idea how to augment the snapshots by native state let me know.

Oh I see so you don't want to modify props directly because they're not modified in reality, that's a fair point. In that case the only way to make them visible would be to implement a custom snapshot matcher, which could be interesting

src/fire-event.ts
mdjastrzebski: Both Fire Event and User Event are supported APIs, just like in RTL. While User Event provides more realistic testing by emitting multiple events, it's still fine to stay on Fire Event, e.g. when having a large test base using it. There is no cost for us to keep supporting Fire Event and add small enhancements like this. Shipping simulated native state for only User Event and Jest matchers but not Fire Event would cause confusion.

PR #1654, opened 2024-08-27
refactor: remove real timers warning in User Event

mdjastrzebski:
### Summary

Remove real times warning in User Event `press`/`longPress` due to feedback from community.

### Test plan

Automated unit test should pass.


thiagobrez:
LGTM!

Issue #1624, opened 2024-05-28
RFC: RNTL Cookbook


mdjastrzebski:
## RNTL Cookbook

RNTL is pretty much feature complete, so the next big frontier is documenting and sharing best practices so our users could gain following benefits:
* quicker start
* not needing to reinvent the wheel
* access best practices and discuss them

For that reason I propose creating a RNTL Cookbook. It would be a documentation section where we would gather relatively short "recipes" for handing different concerns, tips & tricks, etc

#### Basic
- [ ] Custom Render functions
- [ ] Testing async code

#### Navigation
- [ ] React Navigation
- [ ] Expo Router (?)

#### State management
- [ ] Redux (classic)
- [ ] Redux (RTK)
- [ ] Zustand
- [ ] Jotai
- [ ] Valtio
- [ ] ...

pierrezimmermannbam:
I love the idea! The topics all seem relevant, I would add a few more I think:

- jest config: there are some areas to explore here but I'm thinking about:
    -  which preset to use
    - using clearMocks option
    - using fake timers by default
    - babel or alternatives such as sucrase?
    - config for universal apps
    - transformIgnorePatterns (issues around non transpiled code are very common and very misunderstood)
- custom component snapshot matcher? This is something that we usually use at BAM to have more succinct and more readable snapshots, we could even discuss about adding one to RNTL's custom matchers
- testing event emitters / subscribers

For librairies I'm not sure if it should be in our cookbook or on the librairies's side. We could contribute to other's libraries' documentation though. I think it makes more sense for instance that the recommended setup for testing reanimated is in their documentation

mdjastrzebski:
@pierrezimmermannbam it would be awesome if you could share some of your practical knowledge in the topics you mentioned (or other). 

As for libraries, the idea is not to replace their documentation, but rather link to that, with a brief TLDR version that would simplify the lives of most developers.

MattAgn:
Love the idea and the topics as well! I'd be glad to help if i get the time.
One area where i see lots of team struggle is testing strategy, where to place your integration test VS unit test VS E2E tests. What type of integration tests should you write etc (I often see integration tests that are very short and only test a very small part of feature)

mdjastrzebski:
@MattAgn come and join the fun :-)

vanGalilea:
Great idea, I'd love to think along :D

vanGalilea:
[...]

PR #1655, opened 2024-08-28
docs(cookbook): network requests recipes

vanGalilea:
### Summary

As part of our [RNTL Cookbook serie](https://github.com/callstack/react-native-testing-library/issues/1624) I've composed 'Network Request' recipes for fetch  & Axios

#### Sanity check Cookbook app
- Run the Cookbook app > Click on "Phonebook.."
- Verify rendering of contact and favorites list

#### Sanity check recipes docs
- Run the website app 
- Ensure you can navigate to "Network Requests Recipes"
- Read and review "Axios" & "Fetch" recipes

#### Visuals

| iOS | Android |
| --- | --- |
| simulator_screenshot_97B8088A-0622-4AE4-81A6-A6EEE7ABF527 | Screenshot_1724824468 |

### Possible spin offs
- [ ] Add an ['msw' ](https://v1.mswjs.io/docs/)recipe under the "Network Requests Recipes" sec.


examples/cookbook/app/network-requests/__tests__/PhoneBook.test.tsx
```diff
@@ -0,0 +1,119 @@
+import { render, screen, waitForElementToBeRemoved } from '@testing-library/react-native';
+import React from 'react';
+import axios from 'axios';
+import PhoneBook from '../PhoneBook';
+import { User } from '../types';
+
+jest.mock('axios');
+
+jest.setTimeout(10000);
```
vanGalilea, line 9: When this test suit runs in CI, the default 5s is not sufficient.
Locally it's a matter of 0.91 s.

Any idea what might be causing this on CI? @mdjastrzebski?

vanGalilea: According to my measuring:

```
renders an empty task list ran in 6612ms
PhoneBook fetches contacts successfully and renders in list ran in 7190ms
Top 8 slowest examples (14.071 seconds, 100.0% of total time):
  PhoneBook fetches contacts successfully and renders in list
    7.19 seconds ./app/network-requests/__tests__/PhoneBook.test.tsx
  renders an empty task list
    6.612 seconds ./app/state-management/jotai/__tests__/TaskList.test.tsx
  renders WelcomeScreen in light theme
    0.078 seconds ./app/custom-render/__tests__/index.test.tsx
  PhoneBook fetches favorites successfully and renders all users avatars
    0.0[5](https://github.com/callstack/react-native-testing-library/actions/runs/10590927775/job/29347478417?pr=1655#step:6:6)6 seconds ./app/network-requests/__tests__/PhoneBook.test.tsx
  PhoneBook fails to fetch contacts and renders error message
    0.052 seconds ./app/network-requests/__tests__/PhoneBook.test.tsx
  PhoneBook fails to fetch favorites and renders error message
    0.052 seconds ./app/network-requests/__tests__/PhoneBook.test.tsx
  renders a to do list with 1 items initially, and adds a new item
    0.02[6](https://github.com/callstack/react-native-testing-library/actions/runs/10590927775/job/29347478417?pr=1655#step:6:7) seconds ./app/state-management/jotai/__tests__/TaskList.test.tsx
  renders WelcomeScreen in dark theme
    0.005 seconds ./app/custom-render/__tests__/index.test.tsx
```

vanGalilea: It seems like the spec. tests "renders an empty task list" and "PhoneBook fetches contacts successfully and renders in list" are the slowest as they're the 1st tests in the file 🤷🏻

examples/cookbook/__mocks__/axios.ts
```diff
@@ -0,0 +1,13 @@
+const chuckNorrisError = () => {
```
pierrezimmermannbam, line 1: I don't think mocking axios is the right way to do this, imo you want to mock only api calls, not the whole library. There are librairies that are made for this such as msw, nock or shm which allow you to do that and also make tests fail if there are unmocked api calls, I think we should recommend those instead, wdyt?

examples/cookbook/app/network-requests/__tests__/PhoneBook.test.tsx
```diff
@@ -0,0 +1,118 @@
...
+import React from 'react';
+import axios from 'axios';
+import PhoneBook from '../PhoneBook';
+import { User } from '../types';
+
+jest.setTimeout(10000);
+
+describe('PhoneBook', () => {
+  it('fetches contacts successfully and renders in list', async () => {
+    (global.fetch as jest.Mock).mockResolvedValueOnce({
```
pierrezimmermannbam, line 11: One problem with this implementation is that we don't test that the correct url is called here. Also if there were multiple api calls on the same screen and you wanted to test that screen, how would you test that? I feel like eventually you'd end up with a custom api mock library instead of using one

vanGalilea: Thanks for the review @pierrezimmermannbam! I understand the importance of mocking correctly, especially ensuring the URL matches the expected one. I've gone ahead and added that check in the latest commit.

Regarding the mocking approach, my intention was to present it in a basic and bare-bones way to keep things simple for those new to mocking. However, I appreciate the feedback, and I’ve expanded the documentation to include a "Further Reading and Alternatives" section. This section now covers various ways to mock differently, which should provide more options and context for developers with varying experience levels.

Let me know if there are any other changes you'd like to see!

vanGalilea: Regarding the mocking approach, my intention was to present it in a basic and bare-bones way to keep things simple for those new to mocking. However, I appreciate the feedback, and I’ve expanded the documentation to include a "Further Reading and Alternatives" section. This section now covers various ways to mock differently, which should provide more options and context for developers with varying experience levels.

Let me know if there are any other changes you'd like to see!

examples/cookbook/__mocks__/axios.ts
vanGalilea: Appreciate the suggestion! I agree that mocking the entire axios library might be overkill, so I've revised the approach to override only the parts of the library that handle API calls—such as get, post, etc. This should ensure we're targeting just the API interactions while keeping the rest of the library intact. This approach also serves as a simple way to show developers how to guard themselves against unexpected API calls.

vanGalilea: Regarding the libraries you mentioned (msw, nock, shm), I appreciate the recommendation. For more context, I've addressed this in a similar way to a previous comment [here](https://github.com/callstack/react-native-testing-library/pull/1655#discussion_r1738028031). I aimed to keep the mocking approach basic and straightforward for beginners, but I’ve also updated the documentation to include a "Further Reading and Alternatives" section. This provides additional options for more advanced mocking scenarios, which should be helpful for developers looking for different tools.

Next PR I'll raise will touch MSW specifically, I've mentioned it in the PR's description 👍🏻

examples/cookbook/app/network-requests/__tests__/PhoneBook.test.tsx
pierrezimmermannbam: I understand your approach but I personally would not recommend not using a library for this because libraries help enforcing good practices and are also easier to use than a custom setup imo and I think we should only document what we consider to be good practices, wdyt @mdjastrzebski ?

vanGalilea: Thanks for your input, @pierrezimmermannbam. I completely agree that following best practices is crucial, and libraries can definitely help enforce those. That said, I also believe that there are multiple ways to achieve a goal, and implementing a custom solution is a valid approach, especially when starting with something basic and straightforward. In my examples, I aimed to leverage existing APIs to keep things simple and accessible for beginners. Using a library can certainly be a valuable expansion, and it's something that developers can consider as they grow more comfortable with mocking. I've also have mentioned this in the "Further Reading and Alternatives" section.

mdjastrzebski (changes requested):
Thank you for contributing this PR. Mocking network requests is an essential part of each RN app, so having a clear recipe should really benefit our users. 

When mocking `fetch` or Axios they way you did it, by mocking `get`/`post` methods there is an issue where there are 2+ API calls using the same verb and serving proper response based on the URL and/or request body. It seems that the current version of the PR skips this problem by handing two API calls separate, one through `fetch` and other through `axios`, while beneficial to showcase these two popopular approaches, it does not show to the user how to solve it.

@pierrezimmermannbam mentioned MSW, which has an elegant solution for this with it's request to response map:
image

Based on [this article](https://kentcdodds.com/blog/stop-mocking-fetch) by Kent C. Dodds, I think mocking `fetch` is suboptimal and we should explore `msw` as a primary approach.

examples/cookbook/__mocks__/axios.ts
```diff
@@ -0,0 +1,14 @@
...
+  throw Error(
+    "Please ensure you mock 'Axios' - Only Chuck Norris is allowed to make API requests when testing ;)",
+  );
+};
+
+export default {
+  ...jest.requireActual('axios'),
+  get: jest.fn(chuckNorrisError),
```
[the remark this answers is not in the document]
vanGalilea: This won't be relevant anymore as MSW warns when API requests don't have corresponding handlers 👍🏻

examples/cookbook/app/network-requests/__tests__/test-utils.ts
```diff
@@ -0,0 +1,113 @@
...
+/**
+ * Ensures that the URL matches the base URL of the API.
+ * @param url
+ * @throws {MismatchedUrlError}
+ */
+const ensureUrlMatchesBaseUrl = (url: string) => {
+  if (!url.includes('https://randomuser.me/api')) throw new MismatchedUrlError(url);
+};
+
+export const mockFetchWithSuccessResponse = () => {
```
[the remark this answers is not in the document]
vanGalilea: Thanks for the review! I'll rewrite this, to primarily use MSW, as it's more realistic. I'll also try to showcase a more real-world scenario with 2+ API calls 👍🏻

examples/cookbook/app/network-requests/PhoneBook.tsx
```diff
@@ -0,0 +1,52 @@
...
+import FavoritesList from './components/FavoritesList';
+import getAllContacts from './api/getAllContacts';
+import getAllFavorites from './api/getAllFavorites';
+
+export default () => {
+  const [usersData, setUsersData] = useState<User[]>([]);
+  const [favoritesData, setFavoritesData] = useState<User[]>([]);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
```
mdjastrzebski, line 14: thinking out loud, perhaps we should showcase usage with TanStack Query instead of manual promise fetching. Implementing data fetching using useEffect to avoid race conditions is [verbose](https://react.dev/reference/react/useEffect#fetching-data-with-effects). Wdyt @vanGalilea ?

examples/cookbook/app/network-requests/__tests__/test-utils.ts
```diff
@@ -0,0 +1,109 @@
...
+  server.use(
+    http.get('https://randomuser.me/api/', ({ request }) => {
+      // Construct a URL instance out of the intercepted request.
+      const url = new URL(request.url);
+      // Read the "results" URL query parameter using the "URLSearchParams" API.
+      const resultsLength = url.searchParams.get('results');
+      // Simulate a server error for the get all contacts request.
+      // We check if the "results" query parameter is set to "25"
+      // to know it's the correct request to mock, in our case get all contacts.
+      if (resultsLength === '25') {
```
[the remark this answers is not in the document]
mdjastrzebski: ok, now i get the idea, so this is because `https://randomuser.me/api` serves only a single endpoint, so you need to differentiate between different calls using the results param. Perhaps we can use `?seed=favorites` to simulate 2 different endpoints. Alternatively use some other test API that has 2 diff endpoints, e.g. swapi.dev. Wdyt?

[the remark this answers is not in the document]
vanGalilea: Good idea, I have made it more explicit!

examples/cookbook/app/network-requests/PhoneBook.tsx
vanGalilea: Showcase is a very good idea, as it is become and industry standard the last years.
With my approach I intended to keep things more straight-forward.

vanGalilea: Let me add it up, this week 👍🏻 In the meanwhile, you can review the rest ;)

mdjastrzebski (approved):
Looks good! Great work @vanGalilea.

Do you want to change anything or can we merge it and publish?

vanGalilea:
> Looks good! Great work @vanGalilea.
> 
> Do you want to change anything or can we merge it and publish?

Let's merge and I might add [Tanstack](https://github.com/callstack/react-native-testing-library/pull/1655#discussion_r1751930029) in the near future to it.

Thanks for the extensive review and thinking along 👍🏻

PR #1663, opened 2024-09-10
feat: add support for alt text queries

deecewan:
### Summary

`<Image />` has an `alt` prop, and this is what the screen reader uses to describe the image. There was previously no way to match on the alt text.

I've added matchers for alt text in this PR, so that we can find images using something accessible by screen readers.

### Test plan

automated tests are included. these are based on the `test-id` tests.

### DIFF
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -36,6 +36,7 @@ test('resetToDefaults() resets internal config to defaults', () => {
       switch: 'A',
       scrollView: 'A',
       modal: 'A',
+      image: 'A',
     },
   });
   expect(getConfig().hostComponentNames).toEqual({
@@ -44,6 +45,7 @@ test('resetToDefaults() resets internal config to defaults', () => {
     switch: 'A',
     scrollView: 'A',
     modal: 'A',
+    image: 'A',
   });
 
   resetToDefaults();
--- a/src/__tests__/host-component-names.test.tsx
+++ b/src/__tests__/host-component-names.test.tsx
@@ -17,6 +17,7 @@ describe('getHostComponentNames', () => {
         switch: 'banana',
         scrollView: 'banana',
         modal: 'banana',
+        image: 'banana',
       },
     });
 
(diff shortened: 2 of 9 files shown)


mdjastrzebski (commented):
Hi @deecewan, thanks for submitting this PR. 

I've done some initial checks, and it seems that `alt` and `aria-label`/`accessibilityLabel` are closely related:

- https://github.com/facebook/react-native/blob/8dabed60f456e76a9e53273b601446f34de41fb5/packages/react-native/Libraries/Image/Image.ios.js#L173
- https://github.com/facebook/react-native/blob/8dabed60f456e76a9e53273b601446f34de41fb5/packages/react-native/Libraries/Image/Image.android.js#L188

As far as I understand, this means that having `alt` or `aria-label/accessibilityLabel` + `accessible` props in runtime in indistinguishable. The same should be reflected in our queries and matchers. 

Since thee above code is platform-specific code, not rendered in your snapshot tests, we would have to simulate that behavior.

In order to have proper support for querying images we would have to do following pieces:
- `getByRole` query to support querying images using `getRole('image' | 'img', { name: altText });`
- `toHaveAccessibleName` support checking `alt` text

As for `getByAltText` it should inspect `alt` props as now, but it should also match when all of the following conditions are met:
- view is a host image
- `accessible` prop is set
- `aria-label`/`accessibilityLabel` matches expected `alt` text

I am still on the fence for adding `*ByAltText` matcher as on one hand it increases RTL compatibility but on the other hand is ok but not really recommended way to query things in RLT (see [query priorities](https://testing-library.com/docs/queries/about/#priority), [KCD advice](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-byrole-most-of-the-time)) and has not be request for so far by the users (you're first).

@deecewan would you be willing to submit a separate PR for Image `alt` prop support in existing queries (`byRole`) and matchers (`toHaveAccessibleName`) as described above? In the mean time we can discuss whether to include `ByAltText` matcher.

mdjastrzebski:
Closing as #1663 seems to provide the required functionality without expanding API surface.

PR #1669, opened 2024-09-15
[POC] Custom test renderer

mdjastrzebski:
### Summary

Experimental test renderer replacing React Test Renderer, which is being deprecated.

### Overview

The new renderer (React Native Test Renderer?) is a simple renderer created using `react-reconciler` package. It operates only on host elements and does not expose composite elements at all. It's still very simple and event propagation is still handled by RNTL.

### Details

1. This simplified most of the code that was responsible for hiding composite components from the output.
2. FireEvent API relied on calling event handlers also on composite elements. This is no longer working and we only invoke host element event handlers.
3. In most cases this is not problematic, except for `Pressable`/`Touchable` where `onPress` exists only on composite level, while on host element level it's `onResponderGrant` + `onResponderRelease` pair. In order to minimise the impact on the existing tests `fireEvent.press` actually calls `onResponderGrant` + `onResponderRelease`  to trigger `onPress` in such case.
4. `UNSAFE_xxx` queries (`*byType` and `*byProps`) and props (`UNSAFE_container`) no longer return composite elements, and can be either removed (high time) or kept but without `UNSAFE_` prefix as they are now safe.
5. Rendering string outside `Text` component is now automatically detected on renderer error and triggers a throw, as it would in React Native renderer.

#### Test status

All tests passing. Some tests had been disabled due to no longer being relevant.


pierrezimmermannbam:
The fact that composite components aren't there anymore makes `fireEvent` very fragile I think because we cannot guarantee anymore that you can use it with the name of any react prop. Even with your fix on fireEvent.press, it would break if users used `fireEvent(element, 'press')` and it would also break for `pressOut` or `pressIn` on `Pressable`. 

I wonder if we should consider removing it completely? Or maybe making it deprecated at first to ease migration but still aim at removing it in the long term. I don't think `fireEvent` would make sense anymore without composite elements and it would also make things simpler for users since there would not be two similar APIs anymore (userEvent and fireEvent). There are some downsides to this though:

- there would be some tests that could not be written anymore, although userEvent should be able to fill the gap in most situations but these tests may be written differently, more from a user perspective
- using userEvent kind of requires using fake timers, which not everyone does. This would make migration even harder but at the same time it would make everybody's tests better imo

Overall it would be a very difficult migration but that would make user's tests better in the end I think so I do think it's a reasonable option. 

Wdyt @mdjastrzebski @MattAgn  ?

mdjastrzebski:
I think we need to keep `fireEvent` as a way to trigger any random host event handler. RN "host" event handlers seem pretty stable due to the fact how components are mocked in RN (there is no true composite `View`, just mock forwarding everything as is from composite `View` to pseudo-host `View`). As I have written before, if RN mocks change, we will have a lot more things to change in RNTL, yet they remained surprisingly stable over the years.

I deeply care about great many RN developers that wrote and/or now maintain large test bases full of `fireEvent` calls. We might consider soft-deprecating (= warning) `fireEvent.press`, `fireEvent.changeText`, `fireEvent.scroll` and keep `fireEvent(elem, eventName, payload)` form. That way would give them a path (and time) to gradually migrate either to recommended User Event interactions if they exist or just use `fireEvent(element, eventName)` if they do not exist. 

Regarding User Event requiring fake timers, I think that many people are actually using them without fake timers, accepting the delay on `press` event for pressables. It's kinda like writing `sleep(130)` in your tests. You can have that with real times, it won't be optimal, but it will work. That's why I do not think that UE makes fake timers a strict requirement, it just makes your tests run faster.

mdjastrzebski:
Closing as replaced by #1705.

PR #1685, opened 2024-10-24
feat: option for enabling concurrent rendering

mdjastrzebski:
### Summary

Add `concurrentRendering` option to both `render` and `configure` that allows enabling concurrent rendering in the underlying React Test Renderer. The option is disabled by default not to break user tests but should be enabled in the next major release.

### DIFF
--- a/src/render.tsx
+++ b/src/render.tsx
@@ -1,4 +1,8 @@
-import type { ReactTestInstance, ReactTestRenderer } from 'react-test-renderer';
+import type {
+  ReactTestInstance,
+  ReactTestRenderer,
+  TestRendererOptions,
+} from 'react-test-renderer';
 import * as React from 'react';
 import { Profiler } from 'react';
 import act from './act';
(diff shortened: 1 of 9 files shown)


src/render.tsx
```diff
@@ -39,11 +54,18 @@ export function renderInternal<T>(
...
     detectHostComponentNames = true,
     unstable_validateStringsRenderedWithinText,
-    ...testRendererOptions
+    ...rest
   } = options || {};
 
+  const testRendererOptions: TestRendererOptions = {
+    ...rest,
+    // @ts-expect-error incomplete typing on RTR package
+    unstable_isConcurrent: concurrentRoot ?? getConfig().concurrentRoot,
```
slorber, line 66: Just curious, with RTR being deprecated, do you plan to remove it?

https://github.com/facebook/react/tree/main/packages/react-test-renderer

mdjastrzebski: @slorber I've got a working prototype for alternative test renderer built using React Reconciler (the package that virtually all alt React renderers use) in #1669. It passes all of RNTL tests and would be a good replacement for RTR. I still need to modify it for React 19 (in #1690). While transition to the new render will happen, there is no immediate urgency to replace RTR, as it still works with React 19, although with deprecation warning (which is btw disabled for RN testing).

My plan is as follows:
- RNTL v13 (alpha version already published) will be based on RTR and support React 18 & 19. 
- RNTL v14 will be based on the new renderer and support probably React 19+

slorber: I see thanks 👍

mdjastrzebski:
This PR has been released in [v12.8.0](https://github.com/callstack/react-native-testing-library/releases/tag/v12.8.0) 🎉

PR #1690, opened 2024-10-25
feat: react 19 custom renderer [POC]

mdjastrzebski:


slorber:
looks like ConcurrentRoot is going to become the default: https://github.com/facebook/react/pull/31476

mdjastrzebski:
RNTL v13 (the last one on RTR) is going to use concurrent root by default too.

Issue #1687, opened 2024-10-24
React Native Testing Library + Expo Router Testing Library: Unable to find node on an unmounted component after press event simulation


winghouchan:
## Describe the bug

When using React Native Testing Library with Expo Router Testing Library, simulating a user press event on a `Link` triggers the following error:

```
Unable to find node on an unmounted component.

  at findCurrentFiberUsingSlowPath (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:574:13)
  at ReactTestInstance._currentFiber (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:18325:17)
  at ReactTestInstance.get (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:18397:19)
  at props (node_modules/@testing-library/react-native/src/user-event/utils/dispatch-event.ts:25:26)
  at getEventHandler (node_modules/@testing-library/react-native/src/user-event/utils/dispatch-event.ts:12:19)
  at emitTextPressEvents (node_modules/@testing-library/react-native/src/user-event/press/press.ts:125:16)
  at basePress (node_modules/@testing-library/react-native/src/user-event/press/press.ts:45:5)
  at Object.press (node_modules/@testing-library/react-native/src/user-event/press/press.ts:17:3)
  at wrapAsync (node_modules/@testing-library/react-native/src/helpers/wrap-async.ts:19:22)
```

More details, including a hypothesis and investigation notes, can be found in the original issue opened in Expo's GitHub: expo/expo/issues/32313. I'm unsure where the responsibility of the issue lies (here versus Expo), hence opening an issue here as well.

## Expected behavior

The above error does not occur. The press event simulations are successful.

## Steps to Reproduce

A minimal reproducible example is available [here](https://github.com/winghouchan/expo-router-testing-library-unable-to-find-node-mcve). After cloning and installing dependencies:

1. Run `npm test`.

## Versions

```
@testing-library/react-native: ^12.7.2 => 12.7.2 
react: 18.2.0 => 18.2.0 
react-native: 0.74.5 => 0.74.5 
react-test-renderer: 18.2.0 => 18.2.0 
```

mdjastrzebski:
Let's continue this in your original issue in Expo Router: https://github.com/expo/expo/issues/32313#issuecomment-2437042183

winghouchan:
Moving the convo here because there's probably not anything for Expo's team to do. Summary of the conversation so far:

[...]

PR #1696, opened 2024-10-25
fix: press event order

mdjastrzebski:
### Summary

Resolves: #1687 

Reverses `pressOut` and `press` event order so that it matches experimental data for regular presses:
 1. `pressIn`,
 2. (press duration)
 3. `pressOut`
 4. (no waiting) `press`. 

The previous setup matched edge case for very short (< 130ms) presses, when order was:
1. `pressIn`
2. (short press duration < 130 ms)
3. `press`
4. (wait till 130 ms - short press duration)
5. `pressOut`

## Scope
- [x] Fix event order
- [x] Prevent triggering events on unmounted components

See experiments: https://github.com/callstack/react-native-testing-library/wiki/Press-Events

### Test plan

All tests pass

### DIFF
--- a/src/user-event/press/__tests__/press.real-timers.test.tsx
+++ b/src/user-event/press/__tests__/press.real-timers.test.tsx
@@ -32,7 +32,7 @@ describe('userEvent.press with real timers', () => {
     );
     await user.press(screen.getByTestId('pressable'));
 
-    expect(getEventsNames(events)).toEqual(['pressIn', 'press', 'pressOut']);
+    expect(getEventsNames(events)).toEqual(['pressIn', 'pressOut', 'press']);
   });
 
   test('does not trigger event when pressable is disabled', async () => {
@@ -128,7 +128,7 @@ describe('userEvent.press with real timers', () => {
     );
     await user.press(screen.getByTestId('pressable'));
 
-    expect(getEventsNames(events)).toEqual(['pressIn', 'press', 'pressOut']);
+    expect(getEventsNames(events)).toEqual(['pressIn', 'pressOut', 'press']);
   });
 
   test('crawls up in the tree to find an element that responds to touch events', async () => {
--- a/src/user-event/press/__tests__/press.real-timers.test.tsx
+++ b/src/user-event/press/__tests__/press.real-timers.test.tsx
@@ -198,7 +198,7 @@ describe('userEvent.press with real timers', () => {
     );
     await userEvent.press(screen.getByText('press me'));
 
-    expect(getEventsNames(events)).toEqual(['pressIn', 'press', 'pressOut']);
+    expect(getEventsNames(events)).toEqual(['pressIn', 'pressOut', 'press']);
   });
 
   test('does not trigger on disabled Text', async () => {
@@ -240,7 +240,7 @@ describe('userEvent.press with real timers', () => {
     expect(events).toEqual([]);
   });
 
-  test('works on TetInput', async () => {
+  test('works on TextInput', async () => {
     const { events, logEvent } = createEventLogger();
 
     render(
@@ -255,7 +255,7 @@ describe('userEvent.press with real timers', () => {
     expect(getEventsNames(events)).toEqual(['pressIn', 'pressOut']);
   });
 
-  test('does not call onPressIn and onPressOut on non editable TetInput', async () => {
+  test('does not call onPressIn and onPressOut on non editable TextInput', async () => {
     const { events, logEvent } = createEventLogger();
 
     render(
@@ -270,7 +270,7 @@ describe('userEvent.press with real timers', () => {
     expect(events).toEqual([]);
   });
 
-  test('does not call onPressIn and onPressOut on TetInput with pointer events disabled', async () => {
+  test('does not call onPressIn and onPressOut on TextInput with pointer events disabled', async () => {
     const { events, logEvent } = createEventLogger();
 
     render(
--- a/src/user-event/press/__tests__/press.real-timers.test.tsx
+++ b/src/user-event/press/__tests__/press.real-timers.test.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {
+  Button,
   Pressable,
   Text,
   TextInput,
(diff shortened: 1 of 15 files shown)


mdjastrzebski:
@pierrezimmermannbam I'm messing with User Event `press` / `longPress` implementation. You might want to take a look,

winghouchan:
@mdjastrzebski: interesting observation in regards to 'short' presses. I was able to reproduce this in my own experimentation too. Do you know if this phenomenon is documented somewhere (whether that be actual documentation or a reference to source code)?

mdjastrzebski:
@winghouchan The best source is probably Pressability code: https://github.com/facebook/react-native/blob/33e1ae13f88979d8346e4f5ddd4cbce500a16999/packages/react-native/Libraries/Pressability/Pressability.js#L342

There is the code path when `pressOut` can be either executed immeditaly or delayed: https://github.com/facebook/react-native/blob/33e1ae13f88979d8346e4f5ddd4cbce500a16999/packages/react-native/Libraries/Pressability/Pressability.js#L788

Then there is the code path for `press` when there is no path for delay: https://github.com/facebook/react-native/blob/33e1ae13f88979d8346e4f5ddd4cbce500a16999/packages/react-native/Libraries/Pressability/Pressability.js#L757

src/user-event/press/__tests__/press.real-timers.test.tsx
```diff
@@ -198,7 +198,7 @@ describe('userEvent.press with real timers', () => {
     );
     await userEvent.press(screen.getByText('press me'));
 
-    expect(getEventsNames(events)).toEqual(['pressIn', 'press', 'pressOut']);
+    expect(getEventsNames(events)).toEqual(['pressIn', 'pressOut', 'press']);
```
MattAgn, line 201: there does not seem to be a similar test for the long press event. it would be nice to have one since a similar bug of order of events could occur with longPress

src/user-event/press/press.ts
```diff
@@ -1,13 +1,15 @@
...
 import { isTextInputEditable } from '../../helpers/text-input';
 import { isPointerEventEnabled } from '../../helpers/pointer-events';
 import { isHostText, isHostTextInput } from '../../helpers/host-component-names';
 import { EventBuilder } from '../event-builder';
 import { UserEventConfig, UserEventInstance } from '../setup';
 import { dispatchEvent, wait } from '../utils';
-import { DEFAULT_MIN_PRESS_DURATION } from './constants';
+
+// These are constants defined in the React Native repo
+export const DEFAULT_MIN_PRESS_DURATION = 130;
```
MattAgn, line 11: it could be useful to add the link to the file where those constants are used in the RN repo

MattAgn (approved):

src/user-event/press/press.ts
```diff
@@ -73,18 +75,14 @@ const emitPressablePressEvents = async (
 
   dispatchEvent(element, 'responderGrant', EventBuilder.Common.responderGrant());
 
-  await wait(config, options.duration);
+  // We apply minimum press duration here to ensure that `press` events are emitted after `pressOut`.
+  // Otherwise, pressables would emit them in the reverse order, which in reality happens only for
+  // very short presses (< 130ms) and contradicts the React Native docs.
+  // See: https://reactnative.dev/docs/pressable#onpress
+  let duration = Math.max(options.duration, DEFAULT_MIN_PRESS_DURATION);
+  await wait(config, duration);
```
pierrezimmermannbam, line 83: this will work but I think the previous implementation was closer to the reality and this feels kind of like a work around to me. Wouldn't it be easier to keep the implementation of `basePress` untouched and change the duration in `press` from 0 to `DEFAULT_MIN_PRESS_DURATION`? If someone did a longPress with a duration bellow 130ms (even though it may not be realistic) pressout should come out after press and it won't be the case with this change

pierrezimmermannbam (commented):
Nice work investigating the issue! The fix is correct but I suggested a slightly different approach

src/user-event/press/press.ts
mdjastrzebski: Makes sense for Pressables, but we would have to keep the Text/Input as we simulate pressability behavior there

mdjastrzebski:
This PR has been released in [v12.8.1](https://github.com/callstack/react-native-testing-library/releases/tag/v12.8.1) 🎉

cartok:
Since these changes, I have issues on some tests when using `userEvent()` i'd have to setup `userEvent()` with the `delay` option, like `{ delay: 100 }`. Was it expected from these changes? The `fireEvent()` still works as before.

mdjastrzebski:
@cartok What kind of issues are you having with user event?

friederbluemle:
Ah interesting, after doing a bit more investigation, I found the root cause is related to those 3 added lines that check if the element is mounted:

```ts
function fireEvent(element: ReactTestInstance, eventName: EventName, ...data: unknown[]) {
  if (!isElementMounted(element)) {
    return;
  }
// ...
```

Maybe there is something wrong with the way I query the element? Why would `scrollView.props.refreshControl` not be mounted when I call `fireEvent`?

mdjastrzebski:
Looks like you are trying to fire event on a prop (refresh control), which is a React.Element type (description of React component to be rendered),

Fire Event, and most other RNTL APIs expect to be called on ReactTestInstance, rendered React component.

friederbluemle:
Thanks for the quick reply @mdjastrzebski!

> Fire Event, and most other RNTL APIs expect to be called on ReactTestInstance, rendered React component.

Yes, that makes sense.

Regarding the workarounds, assigning a testID also does not work, it is not able to find the element (I assume it is never "really" rendered).
But - workaround #2 _does work_! Calling `onRefresh()` directly via `scrollView.props.refreshControl.props.onRefresh()` makes the test pass like before.

Using the chained props access really doesn't look good which is why I would prefer fireEvent syntax. I understand it doesn't make sense to use it on an "unmounted" element. The next and real question is why is RefreshControl never rendered/mounted.

mdjastrzebski:
WIP #1822

Issue #1579, opened 2024-03-28
RFC: impact of `detectHostComponentNames`


mdjastrzebski:
## Summary

I've been curious what is the impact of automatically detecting host component names. Due to the lack of better large test suite I've taken RNTL own tests.

Here is my process:
1. Disable tests that fail with hardcode `hostComponentNames` with `test.skip()`.
2. Measure execution time with `time yarn test` in relevant scenarios
3. Add `console.log()` to `detectHostComponentNames()` and count execution count (separate run from measuring time).

Note: we also have `beforeEach(() => resetToDefaults())`, so I measured it with and without it.

- Without `detectHostComponentNames`: 39,4 s / 0 runs
- With `detectHostComponentNames` + without `beforeEach(() => resetToDefaults())`: 44,6 s / 63 runs => 13% overhead
- With `detectHostComponentNames` + with `beforeEach(() => resetToDefaults())`: 47,1 s / 585 runs => 20% overhead

Seems like Jet by default clears the runtime state per each file. So we have `detectHostComponentNames()` being called for each file.

Here are my questions:
- if if my methodology is correct?
- did you observe the same/different effect on your own test-base? ([here](https://github.com/callstack/react-native-testing-library/issues/1423) is a link previous perf impact issue of `detect...`)
- is that number significant enough to warrant a research into alternative approaches?

thymikee:
It's expected from Jest to recreate the whole environment for each test suite (file).

pierrezimmermannbam:
You're running the whole RNTL test suite? On my machine it takes about 7s, does it take only 39,4ms on yours? From the tests I'v run on my machine (M3 pro) the difference is less significant, it takes on average 7.45s to run the test suite without the detect and 7.7s

Other than that yeah the approach is correct. One thing to note though is that I would expect the difference to be even less significant on a real project where tests per file are likely to be more heavy (I'll try to test it on my current project too). I don't know if the perf impact is high enough to look into alternative approaches, I guess if we find a good one it could be worth

mdjastrzebski:
@pierrezimmermannbam The unit was to be `s` not `ms` (sorry for confusion, I've run it on my MBP with Intel)

pierrezimmermannbam:
Oh I see, that makes sense. Strange that the difference in percentage would be so different on our machines. Also I'm just realizing this shows that calling `resetToDefaults` is useless no? Why were we doing this initially?

mdjastrzebski:
[...]

PR #1697, opened 2024-10-31
refactor(v13): remove detect host component names

mdjastrzebski:
### Summary

Remove host component name detection for v13+ release. Instead we hardcode the expected host component names & add relevant tests to make sure our `isHostXxx` function work correctly for RN components.

Rationale:
1. The host component detection code gives impression of RNTL being resilient to RN mocks changes. In the past there was a change from `RCTView`/`RCTText` to `View`/`Text`. In reality that past change wasn't only about the name, but also the level at which components are mocked, as the new mocks no longer execute JS code for View, Text. 
2.  RNTL is actually tightly coupled with RN mocks implementation beyond just rendered host component name, but also for their behavior. In order to properly simulate RN env, we recreate RN observed behavior inside RNTL, e.g. by checking `aria-*` props names, or handling pressability for `Text`/`TextInput`, as that part of the code got removed by the RN mocks. Therefore, potentially any RN mocks change will result in RNTL needing to adapt to these changes beyond the name.
3. The current detection code supported only bases case of Text component usage. While there are people using RCTText direction for performance reasons.
4. When switching to custom test renderer #1669 , we will no longer able to render the components before knowing host component names, as the renderer itself will require configuration for which components are allowed to hold string content directly.
 
### Test plan
 
All tests should pass.


thymikee (approved):

thymikee:
Would be great to get rid of that indirection between incomplete RN core mocks and RNTL

pierrezimmermannbam (approved):
I completely agree on this, this library is anyway coupled with RN mocks so it feels like unnecessary complexity in the end

rtrembecky:
hi, just a quick feedback that this is missing from the [v13.0.0-alpha.1](https://github.com/callstack/react-native-testing-library/releases/tag/v13.0.0-alpha.1) release notes, not sure if intended. I'm eager to try it to see tests performance impact

rtrembecky:
I've done a very quick testing on test suite of one of our RN libs - it went through fine. but unfortunately I didn't see any noticeable performance gain, which is what I'm currently interested in. (the biggest culprits for us are probably barrel files and transformation of RN flow-typed files... not related to this issue)

mdjastrzebski:
@rtrembecky Hmmm, that's a bit weird. I would expecte 10-20% speed improvement as in https://github.com/callstack/react-native-testing-library/issues/1579. Can you share some stats? Total run time before/after, number of tests, etc?

mdjastrzebski:
Here are the stats for RNTL own test base

Before (7905bb597998fece782c57871cbfd7932d86938b):
```
Test Suites: 72 passed, 72 total
Tests:       663 passed, 663 total
Snapshots:   222 passed, 222 total
Time:        9.159 s
Ran all test suites.
yarn test  52.06s user 11.12s system 555% cpu 11.366 total
```

After (13bc78a504fec403626a091f47d6129abd86073a):
```
Test Suites: 72 passed, 72 total
Tests:       663 passed, 663 total
Snapshots:   221 passed, 221 total
Time:        8.074 s
Ran all test suites.
yarn test  36.39s user 7.46s system 455% cpu 9.623 total
```

rtrembecky:
running a single RN test on my M1 Max takes 35s even it it's skipped, saying something about our setup - I think it's the barrel files handling (https://github.com/jestjs/jest/issues/11234), so RNTL performance is most probably overshadowed. I will definitely come back, if I manage to solve this, to report the performance impact.

PR #1726, opened 2025-01-07
feat: event debugging (WIP)

### LINKED ISSUE #1717: Log when `fireEvent` fails to fire an event
## Describe the Feature

I couldn't find a feature request for this, so apologies if this has already been requested. 

It would be hugely beneficial if `fireEvent` could be configured to log a warning if it is unable to perform the requested event. Unfortunately, I sometimes write tests that press on disabled buttons and spend a silly amount of time debugging why my tests aren't working. I usually end up debugging down into `react-native-testing-library` only to find that I'm a dummy and my button has a `disabled` prop or some other attribute preventing the button event from firing. I would be a huge time saver if `react-native-testing-library` could give me a hint when I do dumb things like this.

## Possible Implementations

One possibility might be to allow passing an additional debug param to `fireEvent` that when true will console log why `fireEvent` didn't work. It would be especially helpful to know what prop is preventing events from firing.

Another option that could work but isn't backwards compatible would be to return an object from `fireEvent` when it fails to fire an event.

## Related Issues

N/A

### LINKED ISSUE #1718: UX: feedback when event could not be triggered due to disabled state, etc
## Describe the Feature

Provide user feedback when user could not be triggered due to disabled state, missing event handler, etc

## Possible Implementations

Option on `configure` not to break existing Fire Event / User Event APIs. This option would probably have to be opt-in, as some users are relying on event not being triggered in disabled states.

## Related Issues

CC: @pierrezimmermannbam @MattAgn @AugustinLF wdyt?

mdjastrzebski:
### Summary

Resolves #1717 
Resolves #1718 

Introduces (event) debugging mode that outputs information about event handles not being called due to disabled state, etc.

### Details

Enabling: `configure({ debug: true })`. Cannot be configured on `render` as `fireEvent` and `userEvent` are independent modules, not returns from `render`.

Fire Event:
- logs warnings when event handler was found but disabled:
- logs warning when no event handler was invoked for given `fireEvent` call.

User Event:
- logs debug for events that have missing event handlers.

### To Do

- User Event: log warning if no event at all has been called for given `userEvent` call.
- Fire Event & User Event: consider logging debug for matches/called event handlers.

### DIFF
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -14,9 +14,10 @@ test('configure() overrides existing config values', () => {
   configure({ defaultDebugOptions: { message: 'debug message' } });
   expect(getConfig()).toEqual({
     asyncUtilTimeout: 5000,
+    concurrentRoot: true,
+    debug: false,
     defaultDebugOptions: { message: 'debug message' },
     defaultIncludeHiddenElements: false,
-    concurrentRoot: true,
   });
 });
 
--- a/src/__tests__/fire-event-debug.test.tsx
+++ b/src/__tests__/fire-event-debug.test.tsx
@@ -1,0 +1,69 @@
+import * as React from 'react';
+import { Pressable, Text, View } from 'react-native';
+
+import { configure, fireEvent, render, screen } from '..';
+import { _console } from '../helpers/logger';
+
+beforeEach(() => {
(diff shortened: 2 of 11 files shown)


mdjastrzebski:
Could you share your feedback on this potential new feature @pierrezimmermannbam @MattAgn @AugustinLF

MattAgn (approved):

PR #1802, opened 2025-07-24
fix: toBeDisabled now also evaluates disabled prop when element is Text

### LINKED ISSUE #1801: toBeDisabled() matcher behavior with Text component and onPress/disabled props
I am migrating from `@testing-library/jest-native/extend-expect` to `@testing-library/react-native/extend-expect`.

Previously, with `@testing-library/jest-native`, I could successfully use `toBeDisabled()` on a `Text` component that had an `onPress` prop and was passed a `disabled={true}` prop.

After migrating to `@testing-library/react-native`, this test now fails.

My observation is that if a `Text` component has an `onPress` handler and is also passed a `disabled={true}` prop, the `onPress` callback **does not fire on either Android or iOS**. While there's no visual feedback (e.g., dimming), the functional interaction is prevented cross-platform.

Despite this functional disabling, `toBeDisabled()` from `@testing-library/react-native` fails, reporting:

```
Error: expect(element).toBeDisabled()

Received element is not disabled:
  <Text>
    Add
  </Text>
```

Debugging shows `screen.getByText('Add').props.disabled` is indeed `true`.

#### Question:

Is this change in `toBeDisabled()` behavior for `Text` components (which functionally prevent `onPress` when `disabled={true}`) intended as part of the migration? I'm seeking clarification on whether `Text` components, when used interactively via `onPress` and then functionally disabled, are expected to be recognized by `toBeDisabled()`.

#### Reproduction Steps:

1.  Create a simple React Native component:
    ```typescript
    import React from 'react';
    import { View, Text, Alert } from 'react-native';

    const MyTextButton = ({ isDisabled }) => {
      return (
        <View>
          <Text
            onPress={() => Alert.alert('Tapped!')}
            disabled={isDisabled} // This prop makes onPress non-functional on both platforms
            style={{ padding: 10, borderWidth: 1, borderColor: 'black' }}
            testID="my-text-button"
          >
            Tap Me
          </Text>
        </View>
      );
    };

    export default MyTextButton;
    ```
2.  Create a test file:
    ```typescript
    import React from 'react';
    import { render, screen, userEvent } from '@testing-library/react-native';
    import MyTextButton from './MyTextButton';

    describe('MyTextButton', () => {
      it('should be functionally disabled when prop is true but toBeDisabled() fails', async () => {
        render(<MyTextButton isDisabled={true} />);
        const textButton = screen.getByTestId('my-text-button');

        const alertSpy = jest.spyOn(Alert, 'alert');
        await userEvent.press(textButton);
        expect(alertSpy).not.toHaveBeenCalled(); // This passes, confirming functional disablement
        alertSpy.mockRestore();

        expect(textButton).toBeDisabled(); // This assertion fails
      });
    });
    ```
3.  Run the tests.

#### Versions:

  * `@testing-library/react-native`: [Your Version, e.g., 12.9.0]
  * `@testing-library/jest-native`: 5.4.3 (context for previous behavior)
  * `react-native`: [Your Version, e.g., 0.77.2]

-----

rarenatoe:
### Summary

Resolves #1801 by ensuring `Text` components with the `disabled` prop are correctly identified as disabled by `computeAriaDisabled`.

Previously, `computeAriaDisabled` didn't check for the `disabled` prop on `Text` elements, causing queries like `getByRole('text', { disabled: true })` to fail. The change adds a check to return `true` when `isHostText(element)` and `props.disabled` are both true, aligning `react-native-testing-library`'s behavior with expected accessibility states.

### Test plan

New unit tests have been added for `computeAriaDisabled` to verify this change. These tests cover:

1.  **`Text` component with `disabled={true}`:** Asserts `computeAriaDisabled` returns `true`.

2.  **`Text` component without `disabled` prop or `disabled={false}`:** Asserts `computeAriaDisabled` returns `false`.

3.  **Existing `TextInput` behavior:** Ensures non-editable `TextInput` components are still correctly identified as disabled.

4.  **`aria-disabled` and `accessibilityState.disabled` props:** Verifies these existing checks continue to function.

The new `computeAriaDisabled` tests should pass, demonstrating correct handling of the `disabled` prop on `Text` components.

### DIFF
--- a/src/helpers/__tests__/accessiblity.test.tsx
+++ b/src/helpers/__tests__/accessiblity.test.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { Pressable, Switch, Text, TextInput, TouchableOpacity, View } from 'react-native';
 
 import { isHiddenFromAccessibility, isInaccessible, render, screen } from '../..';
-import { computeAriaLabel, isAccessibilityElement } from '../accessibility';
+import { computeAriaDisabled, computeAriaLabel, isAccessibilityElement } from '../accessibility';
 
 describe('isHiddenFromAccessibility', () => {
   test('returns false for accessible elements', () => {
@@ -278,11 +278,11 @@ describe('isHiddenFromAccessibility', () => {
   test('has isInaccessible alias', () => {
     expect(isInaccessible).toBe(isHiddenFromAccessibility);
   });
-});
 
-test('is not triggered for element with "aria-modal" prop', () => {
-  render(<View aria-modal testID="subject" />);
-  expect(isHiddenFromAccessibility(screen.getByTestId('subject'))).toBe(false);
+  test('is not triggered for element with "aria-modal" prop', () => {
+    render(<View aria-modal testID="subject" />);
+    expect(isHiddenFromAccessibility(screen.getByTestId('subject'))).toBe(false);
+  });
 });
 
 describe('isAccessibilityElement', () => {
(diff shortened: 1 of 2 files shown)


mdjastrzebski:
@rarenatoe You should be able to use v13 without serious effort.

1. Afaik `act` warning should be related do your `react-test-renderer`/`react` version rather than RNTL.
2. You can disable concurrent rendering with `configure({ concurrentRoot: false })` in your `jest-setup.ts`.

Which version of React & React Native are you on?

rarenatoe:
I am on react 18.3.1 and react-native 0.77.3
I'd prefer not to set `concurrentRoot: false` because otherwise it's something that is later forgotten forever and never migrated.

PR #1907, opened 2026-05-17
fix: support accessibilityLabelledBy arrays

sjh9714:
### Summary

Fixes accessible name matching when `accessibilityLabelledBy` is an array.

React Native types allow `accessibilityLabelledBy` to be `string | string[]`, but
the previous lookup compared `nativeID` directly with the prop value. That made
`getByRole('button', { name })` miss elements labelled by an array value.

This PR keeps the existing string behavior and normalizes the label id prop to
an array before matching label elements.

Fixes #1839

### Test plan

- `yarn test src/queries/__tests__/role.test.tsx -t accessibilityLabelledBy`
- `yarn test src/queries/__tests__/role.test.tsx`
- `yarn typecheck`
- `yarn lint`
- `yarn test`
- `yarn prettier`
- `git diff --check`

### DIFF
--- a/src/helpers/accessibility.ts
+++ b/src/helpers/accessibility.ts
@@ -153,10 +153,11 @@ export function computeAriaLabel(instance: TestInstance): string | undefined {
   const labelElementId =
     instance.props['aria-labelledby'] ?? instance.props.accessibilityLabelledBy;
   if (labelElementId) {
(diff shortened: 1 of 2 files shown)


mdjastrzebski:
Closing as replaced by #1909 . Thank you for raising awareness of this issue.

Issue #1917, opened 2026-06-15
node doesn't exit because of live timers when waitFor/findBy encounters an exception


ggilles:
Sometimes, we had our CI jobs stuck seemingly forever after running tests. Jest reported that tests were done, but node didn't exit.

I managed to make this a simple reproduction. It involves a combination of waitFor and findBy*.

Here is a simple test which triggers the bug every time:
```ts
test('should exit nicely when a findByTestId encounters a throw', async () => {
    await render(<Text>Element tree that cannot be formatted</Text>);

    // this can happen in a fault component, but in fact any exception happening while rendering the error seems to trigger the problem
    jest.spyOn(screen, 'toJSON').mockImplementation(() => {
        throw new RangeError('Invalid string length');
    });

    expect(
        await screen.findByTestId('missing-element', undefined, { timeout: 10, interval: 5 }),
    ).toBeDefined();
}, 100);
```

This test will fail: this is expected. What is not expected is that after the test fails, node is stuck and never exits.

Another variant is this one:
```ts
test('should not keep Node alive when Jest aborts a findBy query whose timeout decoration throws', async () => {
    await render(<Text>Element tree that cannot be formatted</Text>);

    jest.spyOn(screen, 'toJSON').mockImplementation(() => {
        throw new RangeError('Invalid string length');
    });

    expect(await screen.findByTestId('missing-element')).toBeDefined();
}, 100);
```

Again, the test should fail, but given it has a timeout of 100ms, node should exit immediately, but it seems that the findByTestId timeout keeps node alive and node eventually exits after several seconds.

## Expected behavior

The test should fail and node should exit immediately after printing `Teardown done`.

## Steps to Reproduce

I created a repo containing reproduction of the issue: 

https://github.com/ggilles/react-native-testing-library-timers

## Versions

The latest version is impacted (14.0.0), as demonstrated by the repo above. This problem also exists on version 13.3.3.

mdjastrzebski:
Fix released in v14.0.1 🎉

PR #1918, opened 2026-06-15
fix cleanup of waitFor timers when an error occurs during onTimeout printout

ggilles:
This is an attempt at fixing the timer cleanup bug described in https://github.com/callstack/react-native-testing-library/issues/1917.

### Summary

There are 2 fixes in there:

* added a try/catch around processing onTimeout so that if an error happens while doing so (while dumping the screen for example), timers are still cleaned up
* added cleanup of running timers when the top level timer expires.

### Test plan

I included tests for both cases explained above. Very new to this codebase, so I'm not sure I'm doing it correctly. Forgive me is isn't suitable 🙏 .

### DIFF
--- a/src/__tests__/wait-for.test.tsx
+++ b/src/__tests__/wait-for.test.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { Pressable, Text, TouchableOpacity, View } from 'react-native';
 
 import { configure, fireEvent, render, screen, waitFor } from '..';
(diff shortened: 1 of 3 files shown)


mdjastrzebski:
@ggilles thank you for finding and fixing these issues 🙌 I've reviewed your PR and added polished some rough edges.

ggilles:
@mdjastrzebski Thanks for taking it over 🙏 .

PR #1932, opened 2026-09-04
fix: detect bare `Switch` host component name

christophpurrer:
### Summary

The iOS native component spec codegens under the name `Switch`, but `HOST_SWITCH_NAMES` only listed the legacy `RCTSwitch` name, so `isHostSwitch` returned `false` for elements with that host name.

Add `Switch` alongside `RCTSwitch`, mirroring how `HOST_TEXT_NAMES` already accepts both `Text` and `RCTText`.

See: https://github.com/react/react-native/pull/58323

### Test plan

Added detects raw Switch component to src/__tests__/host-component-names.test.tsx. It fails on main and passes with this change. Full suite, typecheck, lint and format:check all pass.


src/__tests__/host-component-names.test.tsx
```diff
@@ -38,6 +38,13 @@ test('detects host Switch component', async () => {
   expect(isHostSwitch(screen.root)).toBe(true);
 });
 
+// The iOS native component spec codegens under the name `Switch`, so the bare
+// name has to be recognised alongside the legacy `RCTSwitch` one.
+test('detects raw Switch component', async () => {
+  await render(React.createElement('Switch', { testID: 'switch' }));
```
mdjastrzebski, line 44: What does <Switch> render? `RCTSwitch`? In real code how do you access that `Switch` with out RCT prefix?

mdjastrzebski:
Does it affect any stable RN version?

We have a nightly job running for every RN version, incl. latest, next and canary, that checks the following test:
```
test('detects host Switch component', async () => {
  await render(<Switch />);
  expect(isHostSwitch(screen.root)).toBe(true);
});
```

mdjastrzebski:
@christophpurrer is this issue still valid? Pls check my questions above

PR #1934, opened 2026-09-05
feat: warn when firing events on disabled elements

### LINKED ISSUE #1718: UX: feedback when event could not be triggered due to disabled state, etc

trinadhkoya:
## Summary

Closes #1718.

Firing an event on a disabled element (e.g. a `Pressable` with `disabled={true}`) currently triggers no handler **silently**, which is confusing when debugging why a test "does nothing". This adds a warning in that case, opt-out via `configure({ disabledEventWarning: false })`.

```
▲ Tried to fire the "press" event on a disabled element, so no handler was called.
  If this is intentional, you can disable this warning via `configure({ disabledEventWarning: false })`.
```

## Implementation

- Detection reuses the existing `computeAriaDisabled` helper (the one behind `toBeDisabled`) and emits through the existing `logger`.
- The warning is **gated on no handler being found**, so:
  - events that bubble to an **enabled parent** do **not** warn (a handler did run);
  - `pointerEvents="none"` does **not** warn (it isn't a disabled state).
- `TextInput` editability (`editable={false}`) is intentionally **excluded** — it's a distinct concept from disabled state, and warning on `changeText`/`focus` there produced false positives.
- New config option `disabledEventWarning: boolean`, **on by default**.

## Scope

This PR covers **`fireEvent`**. `userEvent` uses a separate dispatch path (`user-event/press`) with its own disabled handling, so I've left it out here — happy to follow up with a matching change for `userEvent` if you're good with this approach.

## Open question

Per the discussion on #1718: I've defaulted the warning to **on** (opt-out), matching the leaning in the thread. Happy to switch it to **opt-in** for the first release if you'd prefer.

## Testing

- Added tests in `fire-event.test.tsx`: warns on disabled element, does **not** warn when bubbling to an enabled parent, does **not** warn for `pointerEvents="none"`, and is silenced by `configure({ disabledEventWarning: false })`.
- Updated the `configure()` default-config assertion and documented the option in `config.mdx`.
- `yarn typecheck`, `yarn lint`, `yarn format:check`, and the full `yarn test` suite (817 tests) all pass.


trinadhkoya:
Heads-up from my side: after opening this I realized #1726 already resolves #1717/#1718 with a fuller design — a unified `configure({ debug: true })` debugging mode covering both `fireEvent` and `userEvent`. That's clearly the better direction than the dedicated, on-by-default `disabledEventWarning` flag I proposed here, and I don't want to add a competing API.

I'd rather help land your approach. Happy to do whichever is most useful:

- help get #1726 over the line — rebase the current conflicts, pick up the remaining To-Do items (userEvent "no handler at all" warning, optional debug logging for matched handlers), and add tests;
- or adapt this PR to the `debug: true` API;
- or close this in favor of #1726.

Just let me know what you'd prefer. If you'd welcome the help on #1726, I'm glad to put the work in.

PR #1935, opened 2026-09-09
ci: pin explicit permissions on website deploy job

CrickCrock28:
The deploy job uses `peaceiris/actions-gh-pages` with `GITHUB_TOKEN` to publish to `gh-pages`, but the workflow has no `permissions:` block, so it relies on the org default (currently `write`).

We are flipping the harvesthq org default workflow permissions from `write` to `read` (aligning with the BendingSpoons org posture — least-privilege GITHUB_TOKEN). This PR pins the deploy job to the explicit permissions it actually needs (`contents: write`) so the deploy keeps working unchanged when the org default changes.

Part of the org security-settings alignment tracked in the security team's audit (2026-09).

🤖 Generated with [OpenCode](https://opencode.ai) (Bsp-agent-beta)


CrickCrock28:
Opened by mistake — wrong target repo (fork default). Sorry for the noise.

